PR #4, opened 2020-08-07
Add CMake package confige, example and update README.md

k0ekk0ek:
This PR adds a proper CMake package config file and a helloworld example. Apart from that it also updates the README.md to be a little more in line with that of Eclipse Cyclone DDS and point to the correct repositories etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,92 +1,180 @@
-# CXX API for CycloneDDS
+# C++ binding for Eclipse Cyclone DDS
 
-This project provides a CXX API for [CycloneDDS](https://github.com/eclipse-cyclonedds/cyclonedds/)
+An implementation of the [ISO/IEC C++ PSM][1], or simply put, a C++ binding
+for [Eclipse Cyclone DDS][2]. Cyclone DDS is developed completely in the open
+as an Eclipse IoT project (see [eclipse-cyclone-dds][3]) with a growing list
+of [adopters][4] (if you're one of them, please add your [logo][5]). It is a
+tier-1 middleware for the Robot Operating System [ROS 2][6].
 
-[![Build Status](https://dev.azure.com/thijssassen/CXX-API/_apis/build/status/ThijsSassen.cdds-cxx?branchName=master)](https://dev.azure.com/thijssassen/CXX-API/_build/latest?definitionId=4&branchName=master)
-
-## Requirements for the CXX API
-
-In order to build the Idl compiler you need a Linux, Mac or Windows 10 machine with the following
-installed on your host:
-
-  * Git
-  * [Conan](https://conan.io/), this is needed to install build dependencies
-  * [CMake](https://cmake.org/download/), version 3.7 or later.  (Version 3.6 should work but you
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -1,92 +1,180 @@
...
+Windows 10 machine (or, with some caveats, a \*BSD, OpenIndiana one) with the
+following installed on your host:
 
-## Building
+ * C and C++ compilers (most commonly GCC on Linux, Visual Studio on Windows,
+   Xcode on macOS);
+ * [Git](https://git-scm.com/) version control system;
+ * [CMake](https://cmake.org/download/), version 3.7 or later;
+ * [Eclipse Cyclone DDS](https://github.com/eclipse-cyclonedds/cyclonedds/)
+ * [CXX Idl compiler](https://github.com/ADLINK-IST/idlpp-cxx/)
```
eboasson, line 30: I think it would be advisable to add a note that this is a temporary situation and that there will soon be a new IDL compiler in Cyclone that will also support C++

README.md
```diff
@@ -1,92 +1,180 @@
-# CXX API for CycloneDDS
+# C++ binding for Eclipse Cyclone DDS
 
-This project provides a CXX API for [CycloneDDS](https://github.com/eclipse-cyclonedds/cyclonedds/)
+An implementation of the [ISO/IEC C++ PSM][1], or simply put, a C++ binding
+for [Eclipse Cyclone DDS][2]. Cyclone DDS is developed completely in the open
+as an Eclipse IoT project (see [eclipse-cyclone-dds][3]) with a growing list
+of [adopters][4]i (if you're one of them, please ad your [logo][5]). It is a
```
eboasson, line 6:
```suggestion
of [adopters][4]i (if you're one of them, please add your [logo][5]). It is a
```

examples/helloworld/publisher.cpp
```diff
@@ -0,0 +1,88 @@
...
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#include <stdio.h>
+#include <stdlib.h>
+
+#ifdef _WINDOWS
```
eboasson, line 15: These days there is std::thread::sleep_for and with std::chrono providing nice literals so that one can write `std::thread::sleep_for(20ms)` (see https://en.cppreference.com/w/cpp/thread/sleep_for). I think that'd be a lot nicer than this old hack.

k0ekk0ek: The literals are only available since `C++14`. I opted for `std::chrono::milliseconds(xx)` instead, which is available since `C++11`.

k0ekk0ek:
Updated. Mixed the fixes into the commits right away since it isn't all that much code.

ithier:
Can this be merged ? I'm keen to get my hands on it :smile:

PR #10, opened 2020-09-02
Idlcxx martijn

reicheratwork:
Created idlcxx library containing streamer generator functions.
The function idl_streamers_generate takes an idl_tree (the result of an idl file being parsed by the cyclonedds idl processor), and generates a number of functions.
For each C++ representation of each IDL object (as defined by the OMG C++11 language mapping) the following functions are generated:
- write_struct: takes a C++ IDL object and writes it in CDR representation to the supplied memory address
- write_size: takes a C++ IDL object and returns the number of bytes it would take up, if written in CDR representation
- read_struct: takes a C++ IDL object and a memory address and attempts to read the data at the address into the IDL object, assuming it is CDR formatted

Created idl_ostream to help with string buffering/writing/etc.
Added unittests for both idl_ostream and streamer generator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,6 +148,7 @@ windows1809_vs2017: &windows1809_vs2017
     # simply add the extracted folder to the path.
     - choco install innoextract
     - choco install maven --ignore-dependencies
+    - choco install winflexbison3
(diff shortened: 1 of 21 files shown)


k0ekk0ek:
Thanks @reicheratwork!

PR #12, opened 2020-09-05
Tie together generation of serialization and type support

k0ekk0ek:
This PR adds an initial backend for the IDL compiler and takes care of some other minor things. It's dependent on another [PR](https://github.com/eclipse-cyclonedds/cyclonedds/pull/591), so we have to wait until that one is merged before tests will pass. An initial test with `HelloWorldData.idl` leads to the expected result, but the backend is missing certain features still, so generation will likely fail for more complex files


k0ekk0ek:
That takes care of running the tests. The solution was pretty straightforward but it took me some time to figure out the problem. See `src/idlcxx/tests/CMakeLists.txt` on how to iterate over tests and add to the environment. This needs to happen for `cyclonedds` at some point to, but I'm adding that to the list. There's some tests failing, but I'm going to leave that to the person who wrote the test. This should provide a decent enough base for people to continue their work at least. @eboasson, are you ok with the changes and the decision to leave the failing test for the next PR (code is already in there...)?

eboasson:
@k0ekk0ek yes, go ahead. It is an unbearable thought that Windows’ habit of inflicting of cruel and unusual punishment on humanity would put an end to all progress ...

PR #13, opened 2020-09-09
Added support for enumerated discriminators in Unions, an optional discriminator parameter in the union's setter functions and support for inheritance between data-structures.

e-hndrks:
Signed-off-by: erik.hendriks@adlinktech.com

### DIFF
--- a/src/idlcxx/src/backendCpp11Type.c
+++ b/src/idlcxx/src/backendCpp11Type.c
@@ -15,9 +15,10 @@
 #include <assert.h>
 #include <stdlib.h>
 #include <limits.h>
+
 #include "idlcxx/backendCpp11Utils.h"
 #include "idlcxx/backendCpp11Type.h"
-
+#include "idl/string.h"
 
 typedef struct cpp11_member_state_s
 {
@@ -31,7 +32,8 @@ typedef struct cpp11_struct_context_s
 {
   cpp11_member_state *members;
   uint32_t member_count;
-  const char *name;
+  char *name;
+  char *base_type;
 } cpp11_struct_context;
 
 static idl_retcode_t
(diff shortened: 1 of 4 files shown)


src/idlcxx/src/backendCpp11Type.c
```diff
@@ -386,16 +410,16 @@ get_potential_nr_discr_values(const idl_union_t *union_node)
...
-        nr_discr_values = 0xffff;
+        nr_discr_values = USHRT_MAX;
         break;
       case IDL_INT32:
-        nr_discr_values = 0xffffffffU;
+        nr_discr_values = UINT_MAX;
         break;
       case IDL_INT64:
-        nr_discr_values = 0xffffffffffffffffULL;
+        nr_discr_values = ULONG_MAX;
```
eboasson, line 422: It is much better to use UINTx_MAX where x is 8, 16, 32 or 64. The 64-bit case is almost certainly not as intended on a 32-bit machine.

e-hndrks:
@k0ekk0ek : I split up the data representation testcase into separately testable units. I disabled the union testcases due to the scoping issue we run into when using a union with an enumerated discriminant.

k0ekk0ek:
The `CUnit_streamer_generator_array_instance_namespace_present` test fails rather consistently. Don't know if it's only the case for x86, but it seems to be the top one of the Windows builds every time. Since it isn't caused by your changes I'm fine with it for now though. It's a shame you implemented two paths for unions rather than just fixing expression evaluation and there's a number of other things that perhaps should've been done in libidl rather than this backend(?) Other backends for instance may benefit from the union case selection work? We can move things around once changes become less frequent and I'm sure we/I will get a better understanding what should go where once I'll get started on the C backend.

PR #15, opened 2020-09-15
Key generation, typedefs and bug-fixes implemented. First Hello world example.

reicheratwork:
Fixed bug in ostream with varargs list being re-used incorrectly.
Fixed bug in ostream with buffers being transferred incorrectly.
Added key calculation functions to the streamer generator.
Added typedef streaming function generation.
Added unittests for key calculation and typedefs.
Moved the helloworld serdata and sertopic classes to templates to make it easier to create topics using the generated classes.

PR #16, opened 2020-09-15
Adapt backed for scoping changes in libidl

k0ekk0ek:
This PR basically changes every `node->identifier` dereference into `idl_identifier(node)` because every declaration now uses a `name` (location + identifier) for the benefit of better error reporting. Apart from that it also avoids reuse of the same `va_list` in one of the output functions.

### DIFF
--- a/src/idlcxx/src/backendCpp11Type.c
+++ b/src/idlcxx/src/backendCpp11Type.c
@@ -87,7 +87,7 @@ get_cpp11_declarator_data(idl_backend_ctx ctx, const idl_node_t *node)
 
   member_data->member_type_node = ((const idl_member_t *) node->parent)->type_spec;
   member_data->declarator_node = declarator;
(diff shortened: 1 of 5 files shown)


k0ekk0ek:
And now unions work too. I love fixing code by removing most of it :smile:

k0ekk0ek:
Btw, @e-hndrks, I fixed one union case. The others work too, but minimum value selected is negative. It's "correct" according to the current logic, but you may want to look into that(?) Of course, I could also be entirely wrong and the value in the testcase should be negative, but that just seems a little weird.

k0ekk0ek:
Travis isn't green because of the `streamer_generator` tests that fail consistently (and maven...). I looked into it, it's not a straight-forward one-line fix and by the looks of it, @mreicher has a fix for this in #15, so I'm going to merge this, make his changes compatible with the current state and go from there.

PR #19, opened 2020-09-18
Bug fixes and include filtering

reicheratwork:
Fixed the following bugs:
- typedef function calls
- 32 windows not generating code for arrays of instances correctly
Added filtering so no code is generated for included files.

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
@@ -104,17 +104,6 @@ protected:
     AnyDataWriterDelegate(const dds::pub::qos::DataWriterQos& qos,
                           const dds::topic::TopicDescription& td);
 
-
-    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction _copyIn)
-    {
-        this->copyIn = _copyIn;
-    }
-
-    inline org::eclipse::cyclonedds::topic::copyInFunction getCopyIn()
-    {
-        return this->copyIn;
-    }
-
     inline void setSampleSize(size_t _sampleSize)
     {
         this->sampleSize = _sampleSize;
@@ -123,16 +112,6 @@ protected:
     inline size_t getSampleSize()
     {
         return this->sampleSize;
-    }
-
-    inline void setCopyOut(org::eclipse::cyclonedds::topic::copyOutFunction _copyOut)
-    {
-        this->copyOut = _copyOut;
-    }
-
-    inline org::eclipse::cyclonedds::topic::copyOutFunction getCopyOut()
-    {
-        return this->copyOut;
     }
 
     void
--- a/src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
@@ -105,9 +105,9 @@ protected:
                           const dds::topic::TopicDescription& td);
 
 
-    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction copyIn)
+    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction _copyIn)
     {
-        this->copyIn = copyIn;
+        this->copyIn = _copyIn;
     }
 
     inline org::eclipse::cyclonedds::topic::copyInFunction getCopyIn()
@@ -115,9 +115,9 @@ protected:
         return this->copyIn;
     }
 
-    inline void setSampleSize(size_t sampleSize)
+    inline void setSampleSize(size_t _sampleSize)
     {
-        this->sampleSize = sampleSize;
+        this->sampleSize = _sampleSize;
     }
 
     inline size_t getSampleSize()
@@ -125,9 +125,9 @@ protected:
         return this->sampleSize;
     }
 
-    inline void setCopyOut(org::eclipse::cyclonedds::topic::copyOutFunction copyOut)
+    inline void setCopyOut(org::eclipse::cyclonedds::topic::copyOutFunction _copyOut)
     {
-        this->copyOut = copyOut;
+        this->copyOut = _copyOut;
     }
 
     inline org::eclipse::cyclonedds::topic::copyOutFunction getCopyOut()
--- a/src/ddscxx/include/org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp
@@ -168,16 +168,6 @@ public:
     void add_query(org::eclipse::cyclonedds::sub::QueryDelegate& query);
     void remove_query(org::eclipse::cyclonedds::sub::QueryDelegate& query);
 
-    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction _copyIn)
-    {
-        this->copyIn = _copyIn;
-    }
-
-    inline org::eclipse::cyclonedds::topic::copyInFunction getCopyIn() const
-    {
-        return this->copyIn;
-    }
-
     inline void setSampleSize(size_t _sampleSize)
     {
         this->sampleSize = _sampleSize;
@@ -186,16 +176,6 @@ public:
     inline size_t getSampleSize() const
     {
         return this->sampleSize;
-    }
-
-    void setCopyOut(org::eclipse::cyclonedds::topic::copyOutFunction _copyOut)
-    {
-        this->copyOut = _copyOut;
-    }
-
-    inline org::eclipse::cyclonedds::topic::copyOutFunction getCopyOut() const
-    {
-        return this->copyOut;
     }
 
     void setSample(void* sample);
--- a/src/ddscxx/include/org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp
@@ -108,6 +108,10 @@ public:
       void**& c_sample_pointers,
       dds_sample_info_t*& c_sample_infos,
       int num_read) {
+      (void)samples;
+      (void)c_sample_pointers;
+      (void)c_sample_infos;
+      (void)num_read;
       assert(1);
     }
 public:
@@ -164,9 +168,9 @@ public:
     void add_query(org::eclipse::cyclonedds::sub::QueryDelegate& query);
     void remove_query(org::eclipse::cyclonedds::sub::QueryDelegate& query);
 
-    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction copyIn)
+    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction _copyIn)
     {
-        this->copyIn = copyIn;
+        this->copyIn = _copyIn;
     }
 
     inline org::eclipse::cyclonedds::topic::copyInFunction getCopyIn() const
@@ -174,9 +178,9 @@ public:
         return this->copyIn;
     }
 
-    inline void setSampleSize(size_t sampleSize)
+    inline void setSampleSize(size_t _sampleSize)
     {
-        this->sampleSize = sampleSize;
+        this->sampleSize = _sampleSize;
     }
 
     inline size_t getSampleSize() const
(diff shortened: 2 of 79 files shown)


k0ekk0ek:
Thanks @reicheratwork! I'll be happy to merge once the newly introduced warnings are taken care of.

reicheratwork:
Will fix them and push the commit as soon as possible,

k0ekk0ek:
Thanks for fixing the bug @reicheratwork, but it would be better if you integrate the fix into the commit that introduced it. I have problems with accepting bug-fix commits in a PR that introduced it too. If you could do that last thing, I'm happy to merge.

src/ddscxx/include/org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp
```diff
@@ -164,29 +168,29 @@ class OMG_DDS_API AnyDataReaderDelegate : public org::eclipse::cyclonedds::core:
     void add_query(org::eclipse::cyclonedds::sub::QueryDelegate& query);
     void remove_query(org::eclipse::cyclonedds::sub::QueryDelegate& query);
 
-    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction copyIn)
+    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction _copyIn)
     {
-        this->copyIn = copyIn;
+        this->copyIn = _copyIn;
     }
```
k0ekk0ek, line 174: `setCopyIn` can probably be removed entirely in the `idlcxx` branch? Same for `setCopyOut`? @e-hndrks, we probably don't need them anymore right?

src/ddscxx/src/org/eclipse/cyclonedds/core/MiscUtils.cpp
```diff
@@ -18,6 +18,10 @@
 
 #include "dds/dds.h"
 
+#ifdef _WIN32
+#pragma warning(disable : 4996)
+#endif
+
```
k0ekk0ek, line 24: I'm wondering what is deprecated that we're still using? If it's something that's marked deprecated in Cyclone DDS, we probably want to use the alternative instead.

src/ddscxx/src/org/eclipse/cyclonedds/topic/AnyTopicDelegate.cpp
```diff
@@ -192,9 +192,9 @@ AnyTopicDelegate::discover_topics(
     ISOCPP_THROW_EXCEPTION(ISOCPP_UNSUPPORTED_ERROR, "Discovery of AnyTopics not implemented yet");
 }
 
-void AnyTopicDelegate::set_copy_out(org::eclipse::cyclonedds::topic::copyOutFunction copyOut)
+void AnyTopicDelegate::set_copy_out(org::eclipse::cyclonedds::topic::copyOutFunction _copyOut)
 {
-    this->copyOut = copyOut;
+    this->copyOut = _copyOut;
 }
```
k0ekk0ek, line 198: We can probably remove `set_copy_out`?

src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
```diff
@@ -105,29 +105,29 @@ class OMG_DDS_API AnyDataWriterDelegate : public org::eclipse::cyclonedds::core:
...
+    inline void setCopyIn(org::eclipse::cyclonedds::topic::copyInFunction _copyIn)
     {
-        this->copyIn = copyIn;
+        this->copyIn = _copyIn;
     }
 
     inline org::eclipse::cyclonedds::topic::copyInFunction getCopyIn()
     {
         return this->copyIn;
     }
```
k0ekk0ek, line 116: These can probably be removed? Don't know if @e-hndrks is removing them in his upcoming work, but these aren't used anymore(?)

src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
```diff
@@ -105,29 +105,29 @@ class OMG_DDS_API AnyDataWriterDelegate : public org::eclipse::cyclonedds::core:
...
     {
         return this->sampleSize;
     }
 
-    inline void setCopyOut(org::eclipse::cyclonedds::topic::copyOutFunction copyOut)
+    inline void setCopyOut(org::eclipse::cyclonedds::topic::copyOutFunction _copyOut)
     {
-        this->copyOut = copyOut;
+        this->copyOut = _copyOut;
     }
```
k0ekk0ek, line 131: Same.

src/idlcxx/tests/streamer_generator.c
```diff
@@ -2292,9 +2414,9 @@ CU_Test(streamer_generator, bounded_sequence)
   test_bounded_sequence();
 }
 
-CU_Test(streamer_generator, typedef_resolution)
+CU_Test(streamer_generator, typedef_resolution, .disabled =true)  //not working at the moment, therefore skipping
```
k0ekk0ek, line 2417: This one actually might work right now(?)

k0ekk0ek (changes requested):
Thanks @reicheratwork, apart from a few minor nitpicks this looks great! It'd certainly be good to actually see if people introduce new warnings :slightly_smiling_face:

src/idlcxx/tests/streamer_generator.c
k0ekk0ek: If you take the latest `idlcxx` branch from Cyclone DDS that is... :sweat_smile:

src/ddscxx/src/org/eclipse/cyclonedds/core/MiscUtils.cpp
reicheratwork: The main issue is that the errors that are being suppressed is that we are using functions like fopen, which for windows is superceded by functions like fopen_s, but since these functions are not present in linux, i think we should still use fopen and the like and suppress these errors in windows compiles.

src/idlcxx/tests/streamer_generator.c
reicheratwork: I will check 👍

src/ddscxx/src/org/eclipse/cyclonedds/core/MiscUtils.cpp
k0ekk0ek: In this case it maybe because of `strncpy`? We have a `ddsrt_strlcpy` available for that. I don't mind use of `fopen`, but for functions where we have better alternatives in e.g. `ddsrt`, I think it's better to replace the instance. Also, for these `#pragma`'s it may be better to use the `DDSRT_MSVC_ON` macro's.

src/idlcxx/tests/streamer_generator.c
reicheratwork: Yep, passes now!
Thanks!

src/ddscxx/src/org/eclipse/cyclonedds/core/MiscUtils.cpp
reicheratwork: I will look into which of the "unsafe" functions can be replaced by their `ddsrt` versions.
I will also check whether the `DDSRT_MSVC_ON` macro's can replace the `pragma` statements

src/idlcxx/src/backendCpp11Utils.c
```diff
@@ -200,8 +201,8 @@ get_cpp11_fully_scoped_name(const idl_node_t *node)
   scoped_enumerator[0] = '\0';
   while(nr_scopes)
   {
-    strncat(scoped_enumerator, "::", scoped_enumerator_len);
-    strncat(scoped_enumerator, scope_names[--nr_scopes], scoped_enumerator_len);
+    ddsrt_strlcat(scoped_enumerator, "::", scoped_enumerator_len);
+    ddsrt_strlcat(scoped_enumerator, scope_names[--nr_scopes], scoped_enumerator_len);
```
k0ekk0ek, line 205: Can't use ddsrt in idlcxx

src/idlcxx/src/generator.c
```diff
@@ -86,6 +87,8 @@ generate_types(
   idl_ostream_t *stm;
   size_t len;
 
+  DDSRT_WARNING_MSVC_OFF(4996)
```
k0ekk0ek, line 90: Same

src/idlcxx/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ target_include_directories(
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>)
 
-target_link_libraries(idlcxx PUBLIC CycloneDDS::idl)
+target_link_libraries(idlcxx PUBLIC CycloneDDS::idl CycloneDDS::ddsc)
```
k0ekk0ek, line 29: Please don't link `ddsc` into `idlcxx`.

k0ekk0ek (changes requested):

src/idlcxx/src/backendCpp11Utils.c
eboasson: @k0ekk0ek it is perhaps wise to reconsider that position, especially in the light of https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/19#discussion_r492791663 which at least suggests that this is a mistake so easily made that it would be better for it not to be considered a mistake.

The only complication I see is a (likely, future) circular dependency caused by linking idlc with libddsc, and having two definitions of the same symbols around if ddsrt is (statically) linked into idlc and idlc and libddsc are both shared libraries linked into the same process. Making ddsrt a shared library in its own right would solve both, but is a bit of barrier to LTO.

Another possible solution (one that I would be fine with) is linking idlc twice: once for bootstrapping libddsc wherein ddsrt is statically linked, and once against libddsc for distribution.

k0ekk0ek: I really prefer not to use `ddsrt` in `idlc` or `libidl` (this is about `idlcxx`, which is a separate library of course). But anyway, `ddsrt` must run on platforms that really don't offer much. The same is not true for platforms where the idl compiler will run. That doesn't mean we need to copy the code, (I/we)'ll think of something for that, but not having `ddsc` in `idl` ensures it's as small as possible. I'm hoping that at some point it'll pay off. For instance, we could have a webapp people can use to generate code. Which probably means Python (or similar) calling into C, at that point, I'm sure people prefer not to have `ddsc` loaded too(?) I also haven't let go of the idea of moving it into a separate project entirely. People may find a decent BSD-licensed IDL compiler very useful. For example, I would've really appreciated it if one would already have been available :sweat_smile:

src/ddscxx/include/org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp
reicheratwork: Removed this function, since it is not used anywhere.

src/ddscxx/src/org/eclipse/cyclonedds/topic/AnyTopicDelegate.cpp
reicheratwork: Removed this function, since it is not used anywhere.

src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
reicheratwork: Removed this function, since it is not used anywhere.

src/ddscxx/include/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.hpp
reicheratwork: Removed this function, since it is not used anywhere.

src/idlcxx/src/backendCpp11Utils.c
reicheratwork: At the moment, I will remove the changes that require the `ddsrt` library.
This will mean that the file-wide warning suppressions for those files will have to come back as well, since single-line warning suppression pragmas cause warnings for linux compilers.

src/idlcxx/src/generator.c
reicheratwork: Will remove these.

src/idlcxx/CMakeLists.txt
reicheratwork: Will remove this.

src/idlcxx/src/backendCpp11Utils.c
k0ekk0ek: Another use-case for a separate library (just popped into my mind) might be that we eventually use it to generate code for e.g. builtin topics. I'm not saying we will at some point, but it's not unthinkable either(?)

k0ekk0ek:
LGTM. Thanks!

PR #22, opened 2020-09-24
Fix warnings, switch back to Xcode 11.5 and treat warnings as errors

k0ekk0ek:
This PR basically takes some of the warning fixes (albeit slightly modified) from @reicheratwork done on #19 and apply them to `master`. There's two slight differences.
1. Use a preprocessor `#pragma` to suppress the `inherits via dominance` warnings which ensures new warnings of the same class don't slip in unnoticed. Also, this way it's ensured applications that pull-in the headers don't receive the same warning, although I haven't tried if that'd be the case.
2. Instead of exporting all the `*Delegate` classes in their entirety, I opted to prefix only the `public` functions with `OMG_DDS_API`. Exporting the entire class is "supposedly" bad practice.

Other than that, this PR enables `-DWERROR=On`, so that newly introduced warnings are treated as errors.


k0ekk0ek:
I didn't try to compile the example yet, so best not merge this immediately. Just in case.

reicheratwork:
There are a few lagging fixes which fit into this pull requrest from my #23 PR, maybe merge this into the idlcxx branch and add those to this PR?

k0ekk0ek:
Example builds and runs fine, so these changes should be good to go. @e-hndrks, maybe you can have a quick look at the warning fixes regarding `OMG_DDS_APP` (Fix non dll-interface class warnings) and the suppression of `inherits via dominance` warnings? @eboasson, you may have an opinion regarding (Fix warning about strncpy being unsafe)?

k0ekk0ek:
@e-hndrks, did you have a chance to look at the fixes regarding `OMG_DDS_API` yet?

k0ekk0ek:
Had another look at the "needs to have dll-interface" warnings. I opted to disable the warnings instead. Also removed quite a lot of `OSPL_` remnants. Where a macro was used to decide between `::std` and some of the other choices, I opted to require `C++11` and go with `::std`. @e-hndrks can you do a quick review? I'll make sure to integrate these into `idlcxx` as well.

reicheratwork (approved):

src/ddscxx/include/dds/core/detail/macros.hpp
```diff
@@ -119,64 +47,34 @@
...
 
-#if defined(OSPL_DDS_CXX11)
-#  define OSPL_ENUM enum class
-#  define OSPL_ENUM_LABEL(_escope,_etype,_elabel) _escope::_etype::_elabel
-#else
-#  define OSPL_ENUM enum
-#  define OSPL_ENUM_LABEL(_escope,_etype,_elabel) _escope::_elabel
-#endif
+#define OSPL_ENUM enum class
+#define OSPL_ENUM_LABEL(_escope,_etype,_elabel) _escope::_etype::_elabel
```
e-hndrks, line 77: I think both OSPL_ENUM and OSPL_ENUM label can be removed altogether. They were meant to abstract the generated data model from whether the compiler supported C++11 semantics or not. Since we have now stated that C++11 support is mandatory, all data generated models no longer use this abstraction and directly specify enum classes.

k0ekk0ek: Yes, they can once the new IDL compiler is being used. Left them in for now because of the code that's generated by the older one.

k0ekk0ek:
Removed some commented-out variables. If the tests pass, I think it's time to merge this into master?

k0ekk0ek:
I'm thinking a month and a half is enough time for people to comment(?) :slightly_smiling_face:  Merging this in.

PR #23, opened 2020-09-24
Additional fixes for compiler warnings

reicheratwork:
Split into ddscxx and idlcxx parts


k0ekk0ek:
Thanks @reicheratwork! Integrated the `ddscxx` ones in the first commit on #22. I think I'm ok with the changes, I can't tell if the changes in `serdata_to_ser_ref` make sense, but maybe it makes sense if @e-hndrks reviews the `idlcxx` part?

reicheratwork:
The serdata_to_ser_ref part is more about removing unused variables for reducing compiler warning, than any changing of functionality or fixing bugs.

PR #24, opened 2020-09-25
Generate appropriate include statements and prevent duplicate macro invocations.

e-hndrks:


src/idlcxx/src/backendCpp11Trait.c
```diff
@@ -87,14 +87,32 @@ generate_traits(idl_backend_ctx ctx, const idl_node_t *node)
 static idl_retcode_t
 generate_macro_call(idl_backend_ctx ctx, const idl_node_t *node)
 {
-  assert(node->mask & IDL_STRUCT);
-
-  char *struct_name = get_cpp11_fully_scoped_name(node);
+  if (node->mask & IDL_STRUCT)
```
reicheratwork, line 90: Should this be (node->mask & IDL_STRUCT) == IDL_STRUCT, since this may cause issues if IDL_STRUCT consists of more than 1 byte set.
There is also a utility function in libidl called `bool idl_is_struct(idl_node_t *node)` testing exactly this thing.

src/idlcxx/src/backend.c
```diff
@@ -270,12 +270,9 @@ idl_walk_tree(
       matches_target_file = true;
     }
 
-    if (matches_target_file)
+    if (matches_target_file && (target_node->mask & mask))
```
reicheratwork, line 273: Is it okay if only one of the bytes of mask matches, or should all of them match?

reicheratwork:
This is, in my opinion, good enough to be merged back, depending on the answers to my comments.

src/idlcxx/src/backend.c
e-hndrks: @reicheratwork The mask in here is meant to be a mask, not an exact match. So I could request to walk over the parse tree and match for any of the following types: (IDL_MODULE | IDL_STRUCT | IDL_UNION). These bits are fully mutually exclusive, but I don't want to make three separate walks to visit the three separate types.

reicheratwork: Check, this was more of a sanity check from my end, whether this was intentional or whether you might have missed something.
In this case, I would say this comment can be considered to be answered to satisfaction.

src/idlcxx/src/backendCpp11Trait.c
e-hndrks: @reicheratwork Given the fact that each type only represents one bit, the additional check would technically not be needed. And given the fact that the system was designed to have one bit per category, I don't think this will change any time soon. However, if we do think this should be avoided, we should do so not only in the backendCpp11Trait.c file, but also in the backendCpp11Type.c file.

k0ekk0ek: That would be preferred, so I agree with @reicheratwork here, but I'm happy to accept it for now. Probably best to use the functions in the future though, they do some extra checks, so it's easier to spot errors/bugs.

k0ekk0ek (approved):

PR #25, opened 2020-09-30
Several fixes

reicheratwork:
Fixing compiler warnings in compiled code and code generated from idl files:
  - added missing copy assignment operators
  - fixed doxygen code creating compiler warnings
Generated code quality of life fixes:
  - no longer opening "empty" namespaces
  - max key size immediately returns the "infinite size" number when one member has infinite size,
     instead of calculating all previous and following members' maximum size as well
Number of bugfixes in serdata implementation, mainly related to calculating the hashes at the correct places

### DIFF
--- a/src/ddscxx/include/dds/sub/LoanedSamples.hpp
+++ b/src/ddscxx/include/dds/sub/LoanedSamples.hpp
@@ -121,6 +121,14 @@ public:
      */
     LoanedSamples(const LoanedSamples& other);
 
+    /**
+     * Copy-assigns a LoanedSamples instance to another.
+     *
+     * No actual data samples are copied.<br>
+     * Just references and reference counts are updated.
+     */
+    LoanedSamples& operator=(const LoanedSamples& other) = default;
+
 
 public:
     /**
--- a/src/idlcxx/src/streamer_generator.c
+++ b/src/idlcxx/src/streamer_generator.c
@@ -36,6 +36,9 @@ format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->key_size_stream, _str, #
 #define format_key_max_size_stream(indent,ctx,_str,...) \
 format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->key_max_size_stream, _str, ##__VA_ARGS__);
 
+#define format_key_max_size_intermediate_stream(indent,ctx,_str,...) \
+format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->key_max_size_intermediate_stream, _str, ##__VA_ARGS__);
+
 #define format_key_stream(indent,ctx,_str,...) \
 format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->key_stream, _str, ##__VA_ARGS__);
 
@@ -190,12 +193,14 @@ struct context
   idl_ostream_t* read_stream;
   idl_ostream_t* key_size_stream;
   idl_ostream_t* key_max_size_stream;
+  idl_ostream_t* key_max_size_intermediate_stream;
   idl_ostream_t* key_stream;
   size_t depth;
   functioncontents_t streamer_funcs;
   functioncontents_t key_funcs;
   context_t* parent;
   bool in_union;
+  bool key_max_size_unlimited;
   const char* parsed_file;
 };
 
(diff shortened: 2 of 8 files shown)


src/ddscxx/include/dds/sub/LoanedSamples.hpp
```diff
@@ -121,6 +121,14 @@ class dds::sub::LoanedSamples
...
     LoanedSamples(const LoanedSamples& other);
 
+    /**
+     * Copy-assigns a LoanedSamples instance to another.
+     *
+     * No actual data samples are copied.<br>
+     * Just references and reference counts are updated.
+     */
+    LoanedSamples& operator=(const LoanedSamples& other) = default;
+
```
e-hndrks, line 131: This is an official OMG spec file. I would be careful in modifying anything in here unless it is really critical, in which case we need to file an issue with the OMG. This does not seem to be one of those critical issues.

e-hndrks (changes requested):

src/idlcxx/tests/streamer_generator.c
```diff
@@ -1356,10 +1356,8 @@ void generate_array_instance_funcs(idl_ostream_t* ostr, bool ns)
 "  }\n\n"\
 "  size_t typedef_key_max_size_td_6(const td_6 &obj, size_t position)\n"\
 "  {\n"\
-"    if (position != UINT_MAX)   position += (4 - (position&0x3))&0x3;  //alignment\n"\
-"    if (position != UINT_MAX)   position += 4;  //bytes for sequence entries\n"\
-"    position = UINT_MAX;\n"\
-"    return position;\n"\
+"    (void)position;\n"\
+"    return UINT_MAX;\n"\
```
e-hndrks, line 1580: Why is position explicitly ignored, while obj, which is also not touched is not explicitly ignored?
I realize that this is a testcase, but I guess that the string in question reflects the code that is actually generated?

src/idlcxx/tests/streamer_generator.c
```diff
@@ -1738,10 +1736,8 @@ void generate_array_instance_funcs(idl_ostream_t* ostr, bool ns)
 "}\n\n"\
 "size_t typedef_key_max_size_td_1(const td_1 &obj, size_t position)\n"\
 "{\n"\
-"  if (position != UINT_MAX)   position += (4 - (position&0x3))&0x3;  //alignment\n"\
-"  if (position != UINT_MAX)   position += 4;  //bytes for sequence entries\n"\
-"  position = UINT_MAX;\n"\
-"  return position;\n"\
+"  (void)position;\n"\
+"  return UINT_MAX;\n"\
```
e-hndrks, line 2009: See also the previous comment

src/ddscxx/include/dds/sub/LoanedSamples.hpp
e-hndrks: After consultation with @reicheratwork I found out that this was needed in order to avoid compilation warnings due to the rule of three and rule of five. I think we will need to indeed file this an a spec issue with the OMG.

src/idlcxx/tests/streamer_generator.c
reicheratwork: You are correct, this is caused by a shortcoming in my code that does not see the difference between the max size function of an object instance and a typedef definition.
I will look into a fix for this

src/idlcxx/tests/streamer_generator.c
reicheratwork: Same.

reicheratwork: Same.

src/idlcxx/src/streamer_generator.c
```diff
@@ -753,73 +765,82 @@ idl_retcode_t add_null(context_t* ctx, int nbytes, bool stream, bool is_key)
...
 
-const char* determine_cast(idl_mask_t mask)
+char* determine_cast(idl_node_t* typespec)
 {
-  if ((mask & IDL_ENUM) == IDL_ENUM)
-    return uint32_cast;
+  if (idl_is_enum(typespec))
+  {
+    char* ns = idl_strdup(""), *returnval = NULL;
+    resolve_namespace(typespec, &ns);
```
e-hndrks, line 796: As a consequence of examining the implementation of resolve_namespace() I discovered something that is not so much an issue in this commit, but in the implementation of the resolve_namespace() function itself. It always uses the IDL name for the namespaces, and forgets to apply the IDL-C++ identifier transformation (i.e. prefix _cxx_ when the identifier clashes with a C++ keyword).

reicheratwork: You are correct, this was an oversight on my part.
I will fix this.

e-hndrks: Fix looks good to me.

e-hndrks (approved):
No further comments on this pull request.

PR #27, opened 2020-10-02
Fix issue where C++ topics would always be considered keyless.

e-hndrks:
A new Trait function is now generated that determines whether a type
has keys or not. The outcome of this new function is now used to set
the is_keyed propery correctly.

Signed-off-by: Erik Hendriks <erik.hendriks@adlinktech.com>


reicheratwork (approved):
Looks like it does what is says on the tin.
Approved.

k0ekk0ek:
LGTM. If the tests pass after this [PR](https://github.com/eclipse-cyclonedds/cyclonedds/pull/611) is merged into `cyclonedds`, I'll merge this one too.

PR #30, opened 2020-10-08
Several fixes

reicheratwork:
- file filtering is now no longer done on namespaces/modules, since these can be reopened in multiple files
- bug fix in flush_ostream function
- rewrote sequence implementation


k0ekk0ek (approved):
Thanks @reicheratwork. Overall the changes make sense, so happy to take this. I do have some things I'd like to alter before merging it to master, but let's worry about that later.

PR #31, opened 2020-10-14
Fixes for generated code and datawriter/reader

reicheratwork:
In generated code, the vector is checked for emptyness before memcopying the container
Removed the use of c99 to c++ copy functions in datawriter/reader since these are no longer necessary


k0ekk0ek:
I think I'm mostly fine with the changes. Removing the copy in/out functions is the right thing to do since they're not there anymore. Have you verified instances can be registered etc?

k0ekk0ek:
We verified basic read/write works with the above changes, so it's a step in the right direction. No warnings are introduced, so while `idlcxx` is definitely not ready to master, but I'd be happy to take these changes into the `idlcxx` branch.

PR #38, opened 2020-10-21
Changing ECA signature

RamziKaroui:
Signed-off-by: Ramzi.Karoui@adlinktech.com <Ramzi.Karoui@adlinktech.com>


examples/ATC/.DS_Store
trittsv, line 1: @RamziKaroui the macos .DS_Store files should not be checked into git, may you can remove this file?

examples/ATC/publisher.cpp
```diff
@@ -0,0 +1,149 @@
...
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ *
+ */
+
+#include <iostream>
+#include "dds/dds.hpp"
+#include "dds/core/ddscore.hpp"
+#include "ATCDataModel_DCPS.hpp"
```
trittsv, line 15: @RamziKaroui may you can update the source code too for the current master. There is no _DCPS anymore. 
And no need to include dds/core/ddscore, its already included in dds/dds.hpp

PR #40, opened 2020-10-28
Fix several bugs in algorihm that tries to establish lowest unused discriminator value of a union.

e-hndrks:
These bugs were causing the idl compiler to crash.

### DIFF
--- a/src/idlcxx/src/backendCpp11Type.c
+++ b/src/idlcxx/src/backendCpp11Type.c
@@ -333,8 +333,8 @@ static idl_retcode_t
 count_labels(idl_backend_ctx ctx, const idl_node_t *node)
 {
   uint32_t *nr_labels = (uint32_t *) idl_get_custom_context(ctx);
-  (void)node;
-  ++(*nr_labels);
+  const idl_case_label_t *label = (const idl_case_label_t *) node;
+  if (label->const_expr) ++(*nr_labels);
   return IDL_RETCODE_OK;
 }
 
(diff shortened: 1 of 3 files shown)


src/idlcxx/src/backendCpp11Type.c
```diff
@@ -114,9 +114,22 @@ struct_generate_attributes(idl_backend_ctx ctx)
...
+
+    idl_file_out_printf(ctx, "%s %s_",
+      struct_ctx->members[i].type_name,
+      struct_ctx->members[i].name);
+    if (defValue)
+    {
+      if (idl_declarator_is_array((const idl_declarator_t *) declarator_node)) {
+        idl_file_out_printf_no_indent(ctx, " = { }");
+      } else {
+        idl_file_out_printf_no_indent(ctx, " = %s", defValue);
```
reicheratwork, line 129: After building this on my own system, I ran into the following: typedefs of primitives do not have default constructed values
This gives a difference of behaviour if you have a default constructor with side-by-side primitives and typedefs of the same primitives containing different data (the primitives being zeroed, and the typedefs containing garbage).

reicheratwork (changes requested):
Also add default construction values for typedefs resolving to primitive types with the same values as the type they resolve to.

src/idlcxx/src/backendCpp11Type.c
e-hndrks: @reicheratwork You are right: I will modify the code to also perform proper initialization to typedefs of (arrays of) primitives.

reicheratwork (approved):
Looks to do exactly what it says on the tin, the only thing i think that needs to be done is remove backendCpp11Type.c line 147, since the variable def_value_present is no longer used, and gives a warning when compiling using CLang.

e-hndrks:
@reicheratwork: You are right. Removed the unused variable and re-pushed the result.

k0ekk0ek:
LGTM. Thanks @e-hndrks!

PR #41, opened 2020-11-05
Several fixes

reicheratwork:
- Fix bounded strings
- Fix boolean sequences
- Fix implicit key stream for member struct
- Fix treatment of unions as members

### DIFF
--- a/src/idlcxx/src/streamer_generator.c
+++ b/src/idlcxx/src/streamer_generator.c
@@ -227,8 +227,8 @@ static uint64_t array_entries(idl_declarator_t* decl);
 static idl_retcode_t add_default_case(context_t* ctx);
 static idl_retcode_t process_node(context_t* ctx, idl_node_t* node);
 static idl_retcode_t process_instance(context_t* ctx, idl_declarator_t* decl, idl_type_spec_t* spec, bool is_key);
-static idl_retcode_t process_struct_decl(context_t* ctx, idl_declarator_t* decl, idl_struct_t* spec, bool is_key);
-static idl_retcode_t process_struct_impl(context_t* ctx, const char* accessor, bool is_key, bool key_is_all_members);
+static idl_retcode_t process_constructed_type_decl(context_t* ctx, idl_declarator_t* decl, idl_type_spec_t* spec, bool is_key);
+static idl_retcode_t process_constructed_type_impl(context_t* ctx, const char* accessor, bool is_key, bool key_is_all_members);
 static idl_retcode_t process_base_decl(context_t* ctx, idl_declarator_t* decl, idl_type_spec_t* spec, bool is_key);
 static idl_retcode_t process_base_impl(context_t* ctx, const char *accessor, uint64_t entries, idl_type_spec_t* spec, bool is_key);
 static idl_retcode_t process_string_decl(context_t* ctx, idl_declarator_t* decl, idl_string_t* spec, bool is_key);
@@ -559,6 +559,11 @@ bool has_keys(idl_type_spec_t* spec)
       mem = (idl_member_t*)(mem->node.next);
     }
   }
+  //is union
+  else if (idl_is_union(spec))
+  {
+    if (idl_is_masked(((idl_union_t*)spec)->switch_type_spec, IDL_KEY)) return true;
+  }
   //is typedef
   else if (idl_is_typedef(spec))
   {
@@ -610,9 +615,9 @@ idl_retcode_t process_instance(context_t* ctx, idl_declarator_t* decl, idl_type_
   {
     return process_base_decl(ctx, decl, spec, is_key);
   }
-  else if (idl_is_struct(spec))
+  else if (idl_is_struct(spec) || idl_is_union(spec))
   {
-    return process_struct_decl(ctx, decl, (idl_struct_t*)spec, is_key);
+    return process_constructed_type_decl(ctx, decl, spec, is_key);
   }
   else if (idl_is_string(spec))
   {
@@ -668,7 +673,7 @@ uint64_t array_entries(idl_declarator_t* decl)
   return entries;
 }
 
-idl_retcode_t process_struct_decl(context_t* ctx, idl_declarator_t* decl, idl_struct_t* spec, bool is_key)
+idl_retcode_t process_constructed_type_decl(context_t* ctx, idl_declarator_t* decl, idl_type_spec_t* spec, bool is_key)
 {
   assert(ctx);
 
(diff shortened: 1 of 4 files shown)


e-hndrks:
@reicheratwork I have got a couple of review comments:
- I see that for boolean vectors you now resize the vector prior to copying the elements. That was not needed for other primitives, since you used the assign() operation which takes care of resizing. But how about non-primitive elements (e.g. vector of string, vector of struct, etc)? I see that there you still iterate over the elements, but you don't seem to invoke the resize() operation either. Was that forgotten or do non primitive sequences get resized some other way?
- I see that you added a bool function called has_keys(), which returns false in case the type is not a struct or a typedef to a struct. Did you realize that a union is allowed to specify its discriminator to be a key as well? See section 7.2.2.4.4.4.8 of the XTypes spec.

The rest looks OK to me.

reicheratwork:
Thanks for the review!

On your first point:
> * I see that for boolean vectors you now resize the vector prior to copying the elements. That was not needed for other primitives, since you used the assign() operation which takes care of resizing. But how about non-primitive elements (e.g. vector of string, vector of struct, etc)? I see that there you still iterate over the elements, but you don't seem to invoke the resize() operation either. Was that forgotten or do non primitive sequences get resized some other way?
That is being done here: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/33c150d81a6b27fd2e9d860ae1414a38f6bcb3c5/src/idlcxx/src/streamer_generator.c#L1754 so i think this is not an issue

reicheratwork:
On the second point: you are correct, at the moment the entire contents of the union (discriminator and contents) are treated as the key, irrespective of the annotations on the union discriminator, this will be corrected.

src/idlcxx/src/streamer_generator.c
```diff
@@ -532,6 +547,27 @@ void load_locals(context_t* ctx, bool* storage)
...
+bool has_keys(idl_type_spec_t* spec)
+{
+  //is struct
+  if (idl_is_struct(spec))
+  {
+    idl_member_t* mem = ((idl_struct_t*)spec)->members;
+    while (mem)
+    {
+      if (idl_is_masked(mem, IDL_KEY)) return true;
+      mem = (idl_member_t*)(mem->node.next);
```
k0ekk0ek, line 559: You could use `idl_next` here

src/idlcxx/src/streamer_generator.c
```diff
@@ -532,6 +547,27 @@ void load_locals(context_t* ctx, bool* storage)
...
 
+bool has_keys(idl_type_spec_t* spec)
+{
+  //is struct
+  if (idl_is_struct(spec))
+  {
+    idl_member_t* mem = ((idl_struct_t*)spec)->members;
+    while (mem)
+    {
+      if (idl_is_masked(mem, IDL_KEY)) return true;
```
k0ekk0ek, line 558: The type_spec for a member can be a struct itself, shouldn't this be recursive? Actually, I'd probably unalias the type_spec first and then see if it's a struct (or a union?).

src/idlcxx/src/streamer_generator.c
```diff
@@ -721,9 +783,7 @@ idl_retcode_t check_alignment(context_t* ctx, int bytewidth, bool is_key)
       format_write_stream(0, ctx, false, "size_t ");
       ctx->streamer_funcs.alignmentpresent = true;
     }
-    format_write_stream(0, ctx, false, alignmentbytes, ctx->depth);
-    format_write_stream(0, ctx, false, " = ");
-
+    format_write_stream(0, ctx, false, alignmentbytes " = ", ctx->depth);
```
k0ekk0ek, line 791: From the context I could make out `alignmentbytes` was a macro. I won't require you to fix it right now, but perhaps it's good to know that macro's are always uppercase within the Eclipse Cyclone DDS project and is common practice in a lot of other code bases.

src/idlcxx/src/streamer_generator.c
```diff
@@ -46,6 +46,14 @@ format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->key_write_stream, _str,
...
 format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->key_read_stream, _str, ##__VA_ARGS__);
 
+#define format_max_size_stream(indent,ctx,key,_str,...) \
+format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->max_size_stream, _str, ##__VA_ARGS__); \
+if (key) { format_key_max_size_stream(indent, ctx, _str, ##__VA_ARGS__); }
+
+#define format_max_size_intermediate_stream(indent,ctx,key,_str,...) \
+format_ostream_indented(indent ? ctx->depth*2 : 0, ctx->max_size_intermediate_stream, _str, ##__VA_ARGS__); \
+if (key) { format_key_max_size_intermediate_stream(indent, ctx, _str, ##__VA_ARGS__); }
+
```
k0ekk0ek, line 56: Not a fan of using macro's like this (also not sure if the paste operator is used properly). You don't have to change it right now, but with constructions like this, I usually find there's a more elegant solution. If you must use macro's though, probably best to not use parameters twice (I could pass `some_char_pointer++`) and escape them to avoid the most obvious errors.

k0ekk0ek (commented):
Thanks @reicheratwork! Looks mostly good to me. There's some "style-comments" that I'm fine with for now and one actual question regarding the `has_keys` function.

src/idlcxx/src/streamer_generator.c
reicheratwork: idl_next looks like a good replacement for these kinds of expressions, I will switch to idl_next

src/idlcxx/src/streamer_generator.c
reicheratwork: That is not done here, because it is not necessary.
This function checks if any of the members are set to be key values.
This is used to generate the correct key stream for structs and unions.
Only if none of the members are set as keys, should the complete struct be added to the key stream, if it itself is set to be a key value.
If any of the member struct's members are set as keys, then only the members that are set to key are added to the key stream.

struct a {
  long b_1;
  long b_2;
};

struct c {
  @key a m_a;
};

struct d {
  @key long e_1;
  long e_2;
};

struct f {
  @key d m_d;
};

In the case of c, the key stream will contain both b_1 and b_2, while in the case of f, the key stream will only contain e_1.

src/idlcxx/src/streamer_generator.c
reicheratwork: You are correct.
I should keep this in mind when go through an iteration of cleanup and tidying.

src/idlcxx/src/streamer_generator.c
reicheratwork: Maybe a better approach is to not use macro's, but to define them as functions?
Though those functions would still require the re-use of variadic function parameters.

k0ekk0ek (approved):

PR #42, opened 2020-11-06
Fix issue where idl compiler crashes on a union with no explicit nor implicit default case

e-hndrks:


reicheratwork:
Seems to do what it is supposed to, so this can be merged, depending on the Jenkins result.

k0ekk0ek:
LGTM @e-hndrks, thanks! @reicheratwork, I think I'll stop waiting for the Jenkins results to come in and settle for Travis :stuck_out_tongue_winking_eye:

PR #46, opened 2020-11-17
Fixes for compiler warnings

reicheratwork:
Fixes for old-style casts in ddscxx and code generated through idlcxx


src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
```diff
@@ -479,9 +479,16 @@ dds::sub::detail::DataReader<T>::common_constructor(
             dds::sub::DataReaderListener<T>* listener,
             const dds::core::status::StatusMask& mask)
 {
+#if defined(_MSC_VER)
+#pragma warning( push )
+#pragma warning( disable : 4127 )
+#endif
```
k0ekk0ek, line 485: I've added a `DDSCXX_MSVC_WARNING_OFF` for this purpose, better use that instead.

src/idlcxx/src/backendCpp11Utils.c
```diff
@@ -324,7 +324,7 @@ get_cpp11_templ_type_const_value(const idl_constval_t *constval)
     return NULL;
   assert(constval->value.str);
   len = strlen(constval->value.str);
-  if (!(str = malloc(len + 2 /* quotes */ + 1)))
+  if ((str = malloc(len + 2 /* quotes */ + 1)) == 0)
```
k0ekk0ek, line 327: Which warning does this fix? Better use `NULL` instead of `0`, it's C code rather than C++.

src/idlcxx/tests/CMakeLists.txt
```diff
@@ -11,14 +11,50 @@
...
+        string(REGEX REPLACE "0x0d,0x0a" "0x0a" filedata ${filedata})
+        # Add terminating NULL to have the array be interchangeable with a string
+        string(CONCAT filedata ${filedata} "0x0")
+        # Append data to output file
+        file(APPEND ${out_dir_}/${output} "const unsigned char ${filename}[] = {${filedata}};\nconst unsigned ${filename}_size = sizeof(${filename});\n")
+    endforeach()
+endfunction()
+
+create_resources(generator_test_cases generator_test_cases.h)
+
```
k0ekk0ek, line 42: Would be better to add this as a separate target so it's done at compile time not configure time. I think I'm also missing that the files are marked as configure files? Meaning that they target is recompiled if the contents of the files is changed.

src/idlcxx/tests/CMakeLists.txt
```diff
@@ -11,14 +11,50 @@
...
 set(idlcxx_test_sources
     "cpp11Backend.c"
     "idl_ostream.c"
     "streamer_generator.c")
 
+#suppressing variadic macro arguments warning for unittests due to CUNIT using them in their test functions
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+  set_source_files_properties(${idlcxx_test_sources} PROPERTIES COMPILE_OPTIONS "-Wno-gnu-zero-variadic-macro-arguments")
+endif()
+
```
k0ekk0ek, line 52: Better do this automatically in the function that's used to process the test sources.

k0ekk0ek (changes requested):

src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
reicheratwork: I will use that then, since I do not like this approach anyway.

src/idlcxx/src/backendCpp11Utils.c
reicheratwork: These changes were made to fix MSVC warning 4706 (assignment within conditional expression: https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4706?view=msvc-160).
By having the explicit comparison operator, this warning does not occur, while the assignment still works as expected.

Using NULL in stead of 0 is a good catch though.

src/idlcxx/tests/CMakeLists.txt
reicheratwork: I will change the processed test files to configure files.

src/idlcxx/tests/CMakeLists.txt
reicheratwork: Yup.

src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
reicheratwork: Seems like this macro is waiting in another PR which is yet to be merged.

k0ekk0ek: Should be in `idlcxx` now. Merge the PR last week.

reicheratwork: Used the DDSCXX macro in stead of the loose pragma statements.

k0ekk0ek (changes requested):

src/idlcxx/src/generator.c
```diff
@@ -94,13 +94,13 @@ generate_types(
   if (idl_asprintf(&file, "%s%s%s.hpp", dir, sep, basename) == -1) {
     ret = IDL_RETCODE_NO_MEMORY;
     goto err;
-  } else if (!(inc = figure_guard(file))) {
+  } else if ((inc = figure_guard(file)) == 0) {
```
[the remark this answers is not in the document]
reicheratwork: I will change this, and i will also look for other places where a check for a NULL pointer is done through ==0

k0ekk0ek (approved):

PR #48, opened 2020-11-25
Extend ContentFilteredTopic and allow static configuration from C++

k0ekk0ek:
The PR builds on the [extended topic filter interface in Eclipse Cyclone DDS](https://github.com/eclipse-cyclonedds/cyclonedds/pull/606) and allows users to set alternative filter functions and filter on:
* sample
* sample and a user specified argument; or
* sampleinfo and a user specified argument; or
* sample, sampleinfo and a user specified argument

Next to that, this PR enables users to [set config programatically](https://github.com/eclipse-cyclonedds/cyclonedds/pull/604) from C++ by adding a DomainParticipant constructor that takes a `ddsi_config`.

@reicheratwork, if you would be so kind and review the changes, that'd be much appreciated.

### DIFF
--- a/src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
+++ b/src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
@@ -28,8 +28,131 @@
 #include <dds/topic/Filter.hpp>
 #include <org/eclipse/cyclonedds/topic/TopicDescriptionDelegate.hpp>
 #include <org/eclipse/cyclonedds/core/ScopedLock.hpp>
+#include <org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp>
 
 #ifdef OMG_DDS_CONTENT_SUBSCRIPTION_SUPPORT
+
+// Required for C++11
+#if __cplusplus == 201103L
+namespace std
+{
+template<bool cond, class T = void>
+using enable_if_t = typename std::enable_if<cond, T>::type;
+
+template<class T = void>
+using decay_t = typename std::decay<T>::type;
+}  // namespace std
+#endif
+
+// meta-programming helpers
+namespace detail
+{
+template<typename T>
+struct remove_first_from_tuple;
+
+template<typename T, typename ... Ts>
+struct remove_first_from_tuple<std::tuple<T, Ts...>>
+{
+  using type = std::tuple<Ts...>;
+};
+
+template<typename T>
+using remove_first_from_tuple_t = typename remove_first_from_tuple<T>::type;
+
+template<class F>
+struct callable_traits;
+
+template<class R, class ... Args>
+struct callable_traits<R(Args...)>
+{
+  using return_type = R;
+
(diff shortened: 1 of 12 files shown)


src/ddscxx/src/org/eclipse/cyclonedds/domain/DomainParticipantDelegate.cpp
```diff
@@ -117,6 +117,66 @@ org::eclipse::cyclonedds::domain::DomainParticipantDelegate::DomainParticipantDe
...
+)
+  : domain_id_(id), qos_(qos)
+{
+    org::eclipse::cyclonedds::domain::DomainWrap::map_ref_iter it;
+    dds_qos_t * ddsc_qos;
+    dds_domainid_t did;
+    dds_return_t ret;
+
+    if ((id != org::eclipse::cyclonedds::domain::default_id()) && (id > 230)) {
+      ISOCPP_THROW_EXCEPTION(ISOCPP_INVALID_ARGUMENT_ERROR, "Invalid domain_id: %d", (int)id);
```
reicheratwork, line 135: Old (C-style) cast used here and at some following lines.
It would be better to use static_cast<int> to be more in line with the rest of the code.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -476,21 +468,13 @@ void sertopic_realloc_samples(
...
-
   ptrs[0] = new T();
 }
 
 template <typename T>
 void sertopic_free_samples(
-  const struct ddsi_sertopic* d, void** ptrs, size_t count,
-  dds_free_op_t op)
+  const struct ddsi_sertopic*, void** ptrs, size_t, dds_free_op_t op)
 {
```
reicheratwork, line 477: the assert(count == 1) several lines down should also be removed, since it is no longer one of the function parameters

src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
```diff
@@ -28,43 +28,224 @@
...
 namespace topic {
 namespace detail {
 
 class FunctorHolderBase
 {
 public:
-    FunctorHolderBase() { };
-
-    virtual ~FunctorHolderBase() { };
+    FunctorHolderBase() = default;
```
reicheratwork, line 164: I think all these explicitly defined default copy/move constructors/assignment operators are superfluous, since when you define none of them the compiler will generate them implicitly for you.

reicheratwork:
Looks to do exactly what it is supposed to.
Except for the superfluous assert, all other comments are more style-related than anything else.
So when these changes are implemented, this PR can be merged.

k0ekk0ek:
Travis is acting up. Personal build results available [here](https://travis-ci.com/github/k0ekk0ek/cyclonedds-cxx/builds/204648539). Comments have been addressed, so I'm merging.

PR #52, opened 2020-12-04
Fix bug of re-alignment not being done after non-base members

reicheratwork:
The re-alignment of writing data was not being done after
writing members containing sequences of lengths not known
at compile time


reicheratwork:
@k0ekk0ek could you review this please?

k0ekk0ek:
Thanks @reicheratwork! I can reason what the fix is supposed to do, but a bit more background information would've been appreciated.

PR #53, opened 2020-12-15
Fixes for unions and bounded sequences

reicheratwork:
- Fixed case indices not being generated for idl unions
- Fixed reading items beyond the bound for bounded sequences
- Fixed issue with take_cdr causing segmentation violations with take_cdr on empty keyless topics

### DIFF
--- a/src/idlcxx/src/streamer_generator.c
+++ b/src/idlcxx/src/streamer_generator.c
@@ -1504,20 +1504,51 @@ idl_retcode_t process_case_label(context_t* ctx, idl_case_label_t* label)
     char* buffer = NULL;
     idl_constval_t* cv = (idl_constval_t*)ce;
 
-    if (idl_is_masked(ce, IDL_INTEGER_TYPE))
+    switch (ce->mask % (IDL_BASE_TYPE * 2))
     {
-      if (idl_asprintf(&buffer, "%lu", cv->value.ullng) == -1)
+    case IDL_INT8:
+      if (idl_asprintf(&buffer, "%" PRId8, cv->value.int8) == -1)
         return IDL_RETCODE_NO_MEMORY;
-    }
-    else if (idl_is_masked(ce, IDL_BOOL))
-    {
-      if ((idl_strdup(cv->value.bln ? "true" : "false")) == 0)
+      break;
+    case IDL_OCTET:
+    case IDL_UINT8:
+      if (idl_asprintf(&buffer, "%" PRIu8, cv->value.uint8) == -1)
         return IDL_RETCODE_NO_MEMORY;
-    }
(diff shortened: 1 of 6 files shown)


src/idlcxx/src/streamer_generator.c
```diff
@@ -1504,20 +1505,52 @@ idl_retcode_t process_case_label(context_t* ctx, idl_case_label_t* label)
     char* buffer = NULL;
     idl_constval_t* cv = (idl_constval_t*)ce;
 
-    if (idl_is_masked(ce, IDL_INTEGER_TYPE))
+    switch (ce->mask % (IDL_BASE_TYPE * 2))
```
k0ekk0ek, line 1508: Please refrain from doing do this in the future. The mask is not a supported "interface". When changes occur these things will break. I'll allow this in for now as the next iteration will have `idl_type()`, but it would have made sense for the backend to add function for this itself.

src/idlcxx/src/streamer_generator.c
```diff
@@ -1504,20 +1504,52 @@ idl_retcode_t process_case_label(context_t* ctx, idl_case_label_t* label)
...
+        return IDL_RETCODE_NO_MEMORY;
+      break;
+    case IDL_BOOL:
+      if ((buffer = idl_strdup(cv->value.bln ? "true" : "false")) == NULL)
+        return IDL_RETCODE_NO_MEMORY;
+      break;
+    case IDL_CHAR:
+      if (idl_asprintf(&buffer, "\'%c\'", cv->value.chr) == -1)
         return IDL_RETCODE_NO_MEMORY;
+      break;
```
k0ekk0ek: I have a strong feeling a more elegant solution would've been possible, but it'll work. Maybe I should add a function to `libidl` to print a constant value to a buffer or something. I guess we'll see when I get started on the C backend.

src/idlcxx/src/streamer_generator.c
```diff
@@ -1504,20 +1504,52 @@ idl_retcode_t process_case_label(context_t* ctx, idl_case_label_t* label)
...
+    default:
+      fprintf(stderr, "unsupported union switch case\n");
+      return IDL_RETCODE_SEMANTIC_ERROR;
```
k0ekk0ek: This isn't a semantic error. It never should happen or `libidl` would be at fault, so this seems like the wrong place to check for these kinds of errors. An assert would be better.

k0ekk0ek (changes requested):

src/idlcxx/src/streamer_generator.c
reicheratwork: I will replace it with an assert.

k0ekk0ek:
LGTM. Thanks @reicheratwork.

PR #55, opened 2021-01-14
Fix for lifetime of entities not matching the attached callback funct…

reicheratwork:
…ions

Listeners on DataReader classes causing the
org::eclipse::cyclonedds::core::EntityDelegate::prevent_callbacks()
function to assert(false) when all references to it have expired
but there are still callbacks referencing it.
This is done by adding these entities to a deffered destructor class,
which handles the end-of-life of these entities outside the thread
where they have callbacks working on them.

### DIFF
--- a/src/ddscxx/CMakeLists.txt
+++ b/src/ddscxx/CMakeLists.txt
@@ -34,6 +34,7 @@ set(sources
     src/org/eclipse/cyclonedds/core/InstanceHandleDelegate.cpp
     src/org/eclipse/cyclonedds/core/EntitySet.cpp
     src/org/eclipse/cyclonedds/core/MiscUtils.cpp
+    src/org/eclipse/cyclonedds/core/DeferredDestruction.cpp
     src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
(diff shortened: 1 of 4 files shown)


sumanth-nirmal:
@k0ekk0ek Do you think this can be merged soon?

k0ekk0ek:
This slipped my mind completely and to be honest, I'll have to do a bit of reading up. @reicheratwork, is this PR up-to-date etc?

k0ekk0ek (changes requested):
I have some basic comments (there's probably more), but I'm more interested in the mechanism used. Short answer is: I'm not totally convinced this is the right fix. IMHO the fact that `assert` is called is not a bug, it's an indication that something else is wrong right? Isn't there a simpler way to avoid the situation than adding another thread? If it's the right fix, wouldn't be possible to use a thread managed by DDSI (garbage collector seems fit for this purpose)? If it's the right fix then if this can occur with the readers, then logic dictates the same can happen for almost all other entities right?

@reicheratwork, @sumanth-nirmal, sorry to further delay this, but can either of you elaborate? Maybe describe the situation a bit?

reicheratwork:
This PR is now absorbed into PR #381

PR #57, opened 2021-01-19
Implement a zero-copy mechanism for loaned read/take calls.

e-hndrks:
Signed-off-by: Erik Hendriks <erik.hendriks@adlinktech.com>

### DIFF
--- a/src/ddscxx/include/dds/sub/TDataReader.hpp
+++ b/src/ddscxx/include/dds/sub/TDataReader.hpp
@@ -378,7 +378,7 @@ public:
          *
          * @param  sfit     Forward-inserting container iterator
          * @param  max_samples Maximum samples to read and copy into the given container
-         * @return          The number of samples in the LoanedSamples container
+         * @return          The number of samples in the given container
          * @throws dds::core::Error
          *                  An internal error has occurred.
          * @throws dds::core::NullReferenceError
@@ -402,7 +402,7 @@ public:
          * Selector with possible filters set.
          *
          * @param  sbit     Back-inserting container iterator
-         * @return          The number of samples in the LoanedSamples container
+         * @return          The number of samples in the given container
          * @throws dds::core::Error
          *                  An internal error has occurred.
          * @throws dds::core::NullReferenceError
@@ -425,7 +425,7 @@ public:
          * Selector with possible filters set.
          *
          * @param  sbit     Back-inserting container iterator
-         * @return          The number of samples in the LoanedSamples container
+         * @return          The number of samples in the given container
          * @throws dds::core::Error
          *                  An internal error has occurred.
          * @throws dds::core::NullReferenceError
@@ -1291,7 +1291,7 @@ public:
      *
      * @param  sfit     Forward-inserting container iterator
      * @param  max_samples Maximum samples to take and copy into the given container
-     * @return          The number of samples in the LoanedSamples container
+     * @return          The number of samples in the given container
      * @throws dds::core::Error
      *                  An internal error has occurred.
      * @throws dds::core::NullReferenceError
@@ -1357,7 +1357,7 @@ public:
      * Look @ref anchor_dds_sub_datareader_invalidsamples "here" for more information.
      *
      * @param  sbit     Back-inserting container iterator
-     * @return          The number of samples in the LoanedSamples container
+     * @return          The number of samples in the given container
      * @throws dds::core::Error
      *                  An internal error has occurred.
      * @throws dds::core::NullReferenceError
@@ -1419,7 +1419,7 @@ public:
      * Look @ref anchor_dds_sub_datareader_invalidsamples "here" for more information.
      *
      * @param  sbit     Back-inserting container iterator
-     * @return          The number of samples in the LoanedSamples container
+     * @return          The number of samples in the given container
      * @throws dds::core::Error
      *                  An internal error has occurred.
      * @throws dds::core::NullReferenceError
@@ -1628,8 +1628,8 @@ public:
      * can be used to indicate that the created entity should not respond to any of its available
      * statuses. The DDS will therefore attempt to propagate these statuses to its factory.
      * The special constant dds::core::status::StatusMask::all() can be used to select all applicable
-     * statuses specified in the “Data Distribution Service for Real-time Systems Version
-     * 1.2” specification which are applicable to the PublisherListener.
+     * statuses specified in the "Data Distribution Service for Real-time Systems Version
+     * 1.2" specification which are applicable to the PublisherListener.
      *
      * @anchor anchor_dds_sub_datareader_commpropagation
      * <i>Status Propagation</i><br>
@@ -1645,7 +1645,7 @@ public:
      * status is not activated in its mask, the application is not notified of the change.
      *
      * The statuses DATA_ON_READERS_STATUS and DATA_AVAILABLE_STATUS are
-     * “Read Communication Statuses” and are an exception to all other plain
+     * "Read Communication Statuses" and are an exception to all other plain
      * communication statuses: they have no corresponding status structure that can be
      * obtained with a get_<status_name>_status operation and they are mutually
      * exclusive. When new information becomes available to a DataReader, the Data
--- a/src/ddscxx/include/dds/sub/detail/SampleRef.hpp
+++ b/src/ddscxx/include/dds/sub/detail/SampleRef.hpp
@@ -71,6 +71,7 @@ public:
 
     SampleRef& copy(const SampleRef& other)
     {
+        static_cast<void>(ddsi_serdata_ref(other.data_));
         this->data_ = other.data_;
         this->info_ = other.info_;
 
--- a/src/ddscxx/include/dds/sub/detail/SampleRef.hpp
+++ b/src/ddscxx/include/dds/sub/detail/SampleRef.hpp
@@ -1,5 +1,5 @@
-#ifndef OMG_DDS_PUB_DETAIL_SAMPLE_HPP_
-#define OMG_DDS_PUB_DETAIL_SAMPLE_HPP_
+#ifndef OMG_DDS_PUB_DETAIL_SAMPLEREF_HPP_
+#define OMG_DDS_PUB_DETAIL_SAMPLEREF_HPP_
 
 /* Copyright 2010, Object Management Group, Inc.
  * Copyright 2010, PrismTech, Corp.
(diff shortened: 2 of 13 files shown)


e-hndrks:
@reicheratwork Could you review this pull request please?

src/ddscxx/include/dds/sub/TDataReader.hpp
```diff
@@ -1628,8 +1628,8 @@ class dds::sub::DataReader : public dds::sub::TAnyDataReader< DELEGATE<T> >
      * can be used to indicate that the created entity should not respond to any of its available
      * statuses. The DDS will therefore attempt to propagate these statuses to its factory.
      * The special constant dds::core::status::StatusMask::all() can be used to select all applicable
-     * statuses specified in the “Data Distribution Service for Real-time Systems Version
-     * 1.2” specification which are applicable to the PublisherListener.
+     * statuses specified in the "Data Distribution Service for Real-time Systems Version
+     * 1.2�" specification which are applicable to the PublisherListener.
```
reicheratwork, line 1632: maybe a typo here?

src/ddscxx/include/dds/sub/detail/SampleRef.hpp
```diff
@@ -0,0 +1,118 @@
...
+{
+namespace detail
+{
+template <typename T>
+class SampleRef;
+}}}
+
+#include <dds/sub/SampleInfo.hpp>
+#include <org/eclipse/cyclonedds/topic/datatopic.hpp>
+//#include <dds/sub/detail/TSampleImpl.hpp>
```
reicheratwork, line 35: Might this be useful in the future, if not, remove?

src/ddscxx/include/dds/sub/detail/Sample.hpp
```diff
@@ -111,9 +111,9 @@ class Sample
         return &this->data_;
     }
 
-    dds::sub::SampleInfo *info_ptr()
+    dds::sub::SampleInfo &info_ptr()
```
reicheratwork, line 114: Maybe rename this function to info_ref, since that is what it is doing now?

src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
```diff
@@ -64,9 +64,6 @@ class LoanedSamplesHolder : public SamplesHolder
 
         uint32_t cpp_sample_size = this->samples_.delegate()->length();
         void **c_sample_pointers = new void * [cpp_sample_size];
-        for (uint32_t i = 0; i < cpp_sample_size; ++i) {
```
reicheratwork, line 67: Is this correctly deleted?
Since now you just get an empty array of pointer back from this function, which have not even been initialized.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -71,8 +72,26 @@ class ddscxx_serdata : public ddsi_serdata {
...
+
+  T* getT()
+  {
+    if (!m_t)
+    {
+      m_t = new T();
+      switch (kind)
+      {
+      case SDK_KEY:
+        m_t->key_read(calc_offset(data(), 4), 0);
```
reicheratwork, line 84: Shouldn't there be a check for if the data field has been initialized?
This is the case when there is data being read, but also when it is being written?

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -263,7 +282,7 @@ ddsi_serdata_t *serdata_from_sample(
       assert(0);
     }
     d->key_md5_hashed() = msg->key(d->key());
-    d->getT() = *msg;
```
reicheratwork, line 266: Somehow I get the impression that an unnecessary copy step is being done here, the sample pointer is loaned, right?

reicheratwork (changes requested):

reicheratwork (changes requested):
The conditions under which the getT() function for the ddscxx_serdata class is called should be examined especially, maybe do some lazy initialization,only when the blob pointer is set?

src/ddscxx/include/dds/sub/TDataReader.hpp
e-hndrks: You are right: small typo. The previous version was using double quotes that weren't ASCII, which looks nice on webbrowsers,but not in plain code. When replacing them with plain ASCII double quotes, some weird character must have stayed behind. I will correct  this.

src/ddscxx/include/dds/sub/detail/SampleRef.hpp
e-hndrks: The file SampleRef.hpp was initially copied from Sample.hpp, and then modified to hold a Ref to the sample instead of a copy of the sample. The original Sample.hpp spec file also has this commented out include statement, but I agree that it would be better to remove it from this copy, and let the spec take care removing the original. Will correct this.

src/ddscxx/include/dds/sub/detail/Sample.hpp
e-hndrks: You are right, it used to return a pointer but was changed to return a ref instead, but the name wasn't changed with it. However, I noticed that the info() function itself was already overloaded with a const and non-const getter, so there is not even a need for another function that returns the non-const pointer. I will remove info_ptr() and replace all its invocations with calls to the non-const getter of info() instead. I will do this both for Sample.hpp and for SampleRef.hpp.

src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
e-hndrks: That is correct. The SampleInfo is now copied in another loop, that is also used to copy the samples themselves. (See set_sample_contents). It seemed to make sense to copy the Sample and its SampleInfo in the same loop instead of making it two separate loops.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
e-hndrks: I don't understand this remark in this particular context. The only change made here is that getT() now returns a pointer to the sample, instead of a reference to the sample. Hence the '.' should be replaced by a '->' in the line you now quote. It could be that you mean this a a more general remark, but then again: how do you know if a sample is initialized since even a sample filled with 0's (the default) still counts as a valid sample.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
e-hndrks: The line you quote here is deleted, but the line that replaces it indeed makes a full copy of a sample. What is important is the context of this piece of code: the serdata_to_sample function. Normally this function is not called by the loaned C++ read/take, which uses read/take_cdr under the hood and just returns the serdata istself. However, in case of a read/take that uses pre-allocated resources I will have to make a copy of the sample into the preallocated resources. It is only in these cases that the serdata_to_sample will be invoked, and in those cases it is conceptually correct.

src/ddscxx/include/dds/sub/detail/SampleRef.hpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    SampleRef& operator=(const SampleRef& other)
+    {
+        return copy(other);
+    }
+
+    SampleRef& copy(const SampleRef& other)
+    {
+        this->data_ = other.data_;
```
eboasson, line 74: Shouldn't this do `ddsi_serdata_ref(other.data_)`?

eboasson (commented):
Thanks @e-hndrks, this feature will be a nice improvement. Still, I do have a few comments/questions because I am not quite convinced it covers all bases. I expect various realistic use cases won't run into problems, but I'd appreciate it if you'd give it another thought. (And please do keep in mind that I may have overlooked some analogous cases.)

src/ddscxx/include/dds/sub/detail/SampleRef.hpp
```diff
@@ -0,0 +1,117 @@
...
+    {
+        return info_;
+    }
+
+    dds::sub::SampleInfo& info()
+    {
+        return info_;
+    }
+
+    bool operator ==(const SampleRef& other) const
```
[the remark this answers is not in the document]
e-hndrks: Good catch. This is actually a by-product from copy-pasting most of the SampleRef from Sample, where Sample also didn't implement the == operator. For the latter there is actually an internal ticket (OSPL-2430) describing the reason for not implementing it: namely the fact that the old IDL compiler didn't generate == operators for the generated sample datatypes yet, and since the == operator of Sample will need to delegate to the == operator of both the SampleInfo and the sample it couldn't be done. With the advent of the new IDL compiler this is no longer an issue, and should be solved as well. For SampleRef I will do that right here, for Sample I guess we should do that as a separate issue.

src/ddscxx/include/dds/sub/detail/SampleRef.hpp
e-hndrks: You are correct. I will correct it.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -71,8 +72,26 @@ class ddscxx_serdata : public ddsi_serdata {
...
   const bool& key_md5_hashed() const { return m_key_md5_hashed; }
   void populate_hash();
-  T& getT() { return m_t; }
-  const T& getT() const { return m_t; }
+
+  T* getT()
+  {
+    if (!m_t)
+    {
+      m_t = new T();
```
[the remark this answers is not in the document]
e-hndrks: Good point. I amagined this would all be exectuted while holding some kind of lock. But if that is indeed not the case, then an atomic compare and swap like you suggest seems to be the way to go. I will look more into it and modify the code accordingly.

src/ddscxx/include/dds/sub/detail/SampleRef.hpp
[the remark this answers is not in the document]
e-hndrks: It turns out the current IDL compiler still doesn't implement the == operator on the data samples. Therefore I propose to fix this issue in a separate pull request that not only covers the generation of == operators on the generated data samples, but also the correct implementation of the == operators for dds::sub::Sample and for dds::sub::SampleRef.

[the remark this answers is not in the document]
eboasson: Proposal accepted 🙂

reicheratwork (approved):

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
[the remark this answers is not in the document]
e-hndrks: Resolved the issue in line with @eboasson's proposal.

PR #62, opened 2021-04-13
Include idlcxx Generate.cmake file in PackageConfig.cmake.in.

jpace121:
I've been excitedly following the development of cyclonedds-cxx, and have been particularly interested in the development of the non-Java based idl compiler/generators since I have a somewhat irrational bias against Java. I compiled and installed cyclonedds and cyclonedds-cxx on the idlcxx branch to try it out. 

When building a HelloWorld-esqe test program, I ran into a problem where cmake would fail, complaining about not being able to find `idlcxx_generate`.

The change in this PR fixed the problem for me locally.

Thanks for your great work.

### DIFF
--- a/PackageConfig.cmake.in
+++ b/PackageConfig.cmake.in
@@ -15,4 +15,8 @@ find_package(CycloneDDS REQUIRED)
 
 include("${CMAKE_CURRENT_LIST_DIR}/@PROJECT_NAME@Targets.cmake")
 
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Hi @jpace121, thanks! Some changes are planned because changes in included files aren't noticed in this version, like [PR 702 in cyclonedds](https://github.com/eclipse-cyclonedds/cyclonedds/pull/702), but it'll definitely do for now.

PR #63, opened 2021-04-22
Initial implementation of templated streaming and update of generator

k0ekk0ek:
This PR contains the initial implementation of templated streaming grabbed (and consolidated) from [@reicheratwork's templated_streaming branch](https://github.com/reicheratwork/cyclonedds-cxx/tree/templated_streaming) and a reimplementation of the C++ generator that works with the current master branch of *cyclonedds* (after [#771](https://github.com/eclipse-cyclonedds/cyclonedds/pull/771) is merged) by my hand.

The move towards templated streaming comes from the desire to support Extensible and Dynamic Topic Types for DDS more easily and cleans up a lot of code. That being said, it's not yet done! There's a couple comments I have myself regarding non-use of namespaces for certain types and streaming functions. Others (perhaps @e-hndrks and @sumanth-nirmal?) might have some remarks/comments/etc too. More testing is definitely required.

As for the generator, it has been simplified, stripped and consolidated. It should support more than the previous implementation and solve a couple of bugs, but no extensive testing has been done so far.

This is the first version that combines templated streaming and supports the most recent version of the IDL compiler. While certainly not perfect, I felt it was time to create a branch so that others see the C++ binding is alive and kicking and can start contributing if desired.

@reicheratwork, if you would be so kind as to review and, if you agree, approve.

@eboasson, I decided to create a branch from the *idlcxx* branch and issue a PR to make the process a little more democratic. If you can spare the time (and feel like it :slightly_smiling_face:), please give it a quick scan.

On a sidenote: I've removed the tests but plan to re-add them soon.


src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
```diff
@@ -34,14 +34,7 @@
...
 #if __cplusplus == 201103L
-namespace std
-{
-template<bool cond, class T = void>
-using enable_if_t = typename std::enable_if<cond, T>::type;
-
-template<class T = void>
-using decay_t = typename std::decay<T>::type;
-}  // namespace std
+#include <org/eclipse/cyclonedds/core/Missing.hpp>
```
reicheratwork, line 37: This is a manner to not have to redefine these missing definitions for older c++ compilers for each file which requires things like std::enable_if, though the name and location of the definition file Missing.hpp is something that may not be the optimal choice.
Should that be changed?

src/ddscxx/include/org/eclipse/cyclonedds/topic/cdr_dummys.hpp
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+#ifndef CDR_DUMMYS_HPP_
+#define CDR_DUMMYS_HPP_
+
+#include <string>
+#include <vector>
+#include <array>
+
```
reicheratwork, line 18: Maybe the choice of the prefix idl_ is not the ideal one, could be that something like cdr_ is better?

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -23,16 +23,40 @@
 #include "dds/ddsi/q_radmin.h"
 #include "dds/ddsi/ddsi_serdata.h"
 #include "dds/ddsi/ddsi_sertopic.h"
+#include "basic_cdr_ser.hpp"
```
reicheratwork, line 26: When the number of supported stream types will be expanded, each will need to have their header included separately, maybe we should bundle these in its own definition header?

reicheratwork (approved):
Looks like a very slimmed down version of the previous implementations, very good!
The comments made are more about choices regarding location/naming of files and class names, so style, not implementation details.
The only place I think might be lacking is the coverage of the testcases.
While the explicit testcase file approach may not be to everyone's preference, it explicitly tests and shows where the test fails.

src/ddscxx/include/org/eclipse/cyclonedds/topic/cdr_dummys.hpp
```diff
@@ -0,0 +1,45 @@
...
+template<size_t N>
+class idl_string : public std::string {
+public:
+  idl_string() : std::string() {}
```
reicheratwork, line 22: Maybe an even more elegant way is to inherit constructors here?

`using std::string::string;` in stead of the explicit declarations?

sumanth-nirmal:
@k0ekk0ek @reicheratwork  I quickly tested this branch with `https://github.com/k0ekk0ek/cyclonedds/tree/templated-streaming` (since this is required to compile) and it fails for the following:
1. arrays
2. string constant
3. boolean constant
4. bounded string unbounded sequence
5. key annotation on uint8
6. Nested IDL (including an IDL file in another)

Simple IDL file with all the above failing cases
```idl
module example {
  module idl {
    struct info
    {
      string node_name;
      octet guid[16];
      float float32_values_[3];
      double float64_values_[3];
      const string FOO = "foo";
      sequence<string<10>> bounded_string_unbounded_seq_values;
      @key uint8 sensor_id;
      const boolean BOOL_CONST_ = TRUE;
    };
  };
};
```
The `idlc` aborts for all the above cases and some times with `free(): invalid pointer`
```sh
ade$ cyclone_ws/cyclonedds/build/bin/idlc -l cyclone_ws/cyclonedds-cxx/build/lib/libidlcxx.so Example_.idl
```
I only tested by generating the code and didn't yet compile the generated code. 
BTW, The option to allow case-sensitive entities in IDL work :smile:

k0ekk0ek:
Thanks @sumanth-nirmal! In the array allocation code there was a bug. The `genalloc` was a quick-fix to get something working until the more elegant solution (for which a PR is open on the cyclonedds repo) has been merged is in. With the latest fix a slightly modified version of the IDL above works (`const` cannot be defined in `struct`).

```
module example {
  module idl {
    const string FOO = "foo";
    const boolean BOOL_CONST_ = TRUE;

    struct info
    {
      string node_name;
      octet guid[16];
      float float32_values_[3];
      double float64_values_[3];
      sequence<string<10> > bounded_string_unbounded_seq_values;
      @key uint8 sensor_id;
    };
  };
};
```

If you issue `idlc -l libidlcxx.so -h`, you'll notice the backend provides support for things like `-fsequence-template=...`. e.g. for sequences you'll be able to provide a string with placeholders, template for lack of a better term, which allows you to specify something like `-fsequence-template=std::vector<{TYPE}, additional code>`.

sumanth-nirmal:
@k0ekk0ek With the fix now, all the above cases work except for an unbounded sequence of bounded strings, something like 

```idl
module test_msgs {
  module msg {
    module dds_ {
      struct custom_types
      {
        sequence<string<10>, 10> bounded_string_bounded_seq_values;
        sequence<string<10>> bounded_string_unbounded_seq_values;
      };
    };    // module dds_
  };    // module msg
};  // module test_msgs
```

k0ekk0ek:
Probably due to the `>>` in the unbounded sequence. The compiler confuses it with a shift operator. At some point I'm hoping to "fix" that, but [even C++ only supports it since C++11](https://en.wikipedia.org/wiki/C%2B%2B11#Right_angle_bracket).

src/idlcxx/src/traits.c
```diff
@@ -0,0 +1,144 @@
...
+        "  {\n"
+        "    return %3$s;\n"
+        "  }\n\n"
+        "  static const char *getTypeName()\n"
+        "  {\n"
+        "    return \"%2$s\";\n" /* skip preceeding "::" according to convention */
+        "  }\n\n"
+        "  static ddsi_sertopic *getSerTopic(const std::string& topic_name)\n"
+        "  {\n"
+        "    auto *st = new ddscxx_sertopic<%1$s>(topic_name.c_str());\n"
```
sumanth-nirmal, line 66: I don't see the definitions of `ddsi_sertopic`, `ddscxx_sertopic` getting included in the generated code, and they are not either part of `TopicTraits.hpp` or `DataRepresentation.hpp`, how are these included?

sumanth-nirmal:
> Probably due to the `>>` in the unbounded sequence. The compiler confuses it with a shift operator. 

I think so too. Having a sequence with bounded string as `sequence<string<10> > foo` (not the space between `>` `>`) generates `idl_sequence<idl_string<10>, 0>`. I think this can be fixed in a follow-up issue.

src/ddscxx/include/org/eclipse/cyclonedds/topic/cdr_dummys.hpp
k0ekk0ek: Indeed, they should not be prefixed with `idl_`. Because 1) These types have nothing to do with IDL, 2) They are not in a reserved namespace and 3) Using a prefix as a namespace is the C way of doings because namespaces aren't available there. I think `org::eclipse::cyclonedds::<something>` or `dds::<something>` should be used (probably the former). There's `dds/core/types.hpp`, so perhaps `org/eclipse/cyclonedds/core/types.hpp` is the right location? They're mainly "aliases" to make serialization more straightforward right? I'd probably opt for `org::eclipse::cyclonedds::core::string` etc. I'm a little worried what happens if a user provides a type that does not match the requirements though. Will the serializer/compiler handle that properly?

src/ddscxx/include/org/eclipse/cyclonedds/topic/cdr_dummys.hpp
k0ekk0ek: I'm not the worlds foremost C++ expert, but if the language allows you to do that, then an alias seems like the right choice.

k0ekk0ek:
The basics are in place, so I'm merging this into the templated-streaming branch to build upon.

PR #68, opened 2021-04-30
Fix SampleRef for nullptr checks

sumanth-nirmal:
This MR should fix segfaults when accessing LoanedSamples. The changes are straight forward and as follows:

- Check if the data ptr is null, before accessing it
- Check for self-assignment in the operator
- Make copy() method private and change the behavior to not return

Signed-off-by: Sumanth Nirmal <sumanth.724@gmail.com>

### DIFF
--- a/src/ddscxx/include/dds/sub/detail/SampleRef.hpp
+++ b/src/ddscxx/include/dds/sub/detail/SampleRef.hpp
@@ -61,26 +61,28 @@ public:
 
     virtual ~SampleRef()
     {
-        ddsi_serdata_unref(data_);
+        if (data_ != nullptr)
+        {
+            ddsi_serdata_unref(data_);
+        }
     }
 
     SampleRef& operator=(const SampleRef& other)
     {
-        return copy(other);
-    }
-
-    SampleRef& copy(const SampleRef& other)
-    {
-        static_cast<void>(ddsi_serdata_ref(other.data_));
-        this->data_ = other.data_;
-        this->info_ = other.info_;
-
-        return *this;
+      if (this != &other)
+      {
+          copy(other);
+      }
+      return *this;
     }
 
 public:
     const T& data() const
     {
+        if (data_ == nullptr)
+        {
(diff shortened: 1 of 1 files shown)


eboasson (approved):
If `data_` is a null pointer, then it will definitely crash. I'm assuming these changes fix it and they do look good to me. What isn't clear to me is *why* that would ever be the case, though I suspect it has to do with default-constructed `SampleRef`s (sometimes?) in the vector of `SampleRef`s inside `LoanedSamples`.

I'd like someone with more detailed knowledge to have a look. @e-hndrks, your name comes to mind, would you be willing to have a look? Thanks!

sumanth-nirmal:
@e-hndrks Can you take a quick look at this?

e-hndrks:
Hi @sumanth-nirmal, I have proposed an alternative fix in PR #88. Since the issue is caused by allocating the loan based on the number of available samples instead of on the number of samples that pass the filter, it seems like a null pointer check is combatting the symptoms rather than the root cause. My suggested fix first executes the filtered read and then instantiates the loan based on the number of elements that matched.

k0ekk0ek:
This is fixed by @e-hndrks in #88. Some of this PR has been integrated in #79 (with credits retained). Closing this one.

dkroenke:
@k0ekk0ek I think there is still an issue because with that patch the iceoryx_dds gateway does not crash anymore. The changes from here are not really integrated from #79.
For now i can live with a patch that applies the changes from here but maybe we can reopen that here.

k0ekk0ek:
Yeah, there's a bug in that part still. @sumanth-nirmal, noticed it and made #122 (#118 will provide some more detail). Eventually, we'll have to make the resize non-required, but checking if the data is a `nullptr` will work. The aforementioned PR has been integrated just now, so with the latest master you should be fine(?)

dkroenke:
@k0ekk0ek I can confirm that the issue is solved after merging https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/122 master.
Thank you!

PR #71, opened 2021-05-04
Fix idlcxx_generate, make assert in alignment conditional and finish up unions

k0ekk0ek:
* `idlcxx_generate` didn't add a dependency on the right target causing parallel Windows builds, like our CI builds, to fail.
* Aparently the cursor is always NULL in alignment of the CDR stream initially, or at least, it depends on the `add_zeroes` parameter. This PR corrects that.
* Union support depended on something not yet available in the IDL tree structure, namely the number of unused labels. [PR 784](https://github.com/eclipse-cyclonedds/cyclonedds/pull/784) in the *cyclonedds* repository fixes that. (until that one is merged, this PR will fail to build).

### DIFF
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -152,48 +152,93 @@ process_case(
 {
   struct streams *streams = user_data;
   const idl_case_t* _case = (const idl_case_t*)node;
-  char *type = NULL;
+  char *type, *value = "";
   const char *name;
+  bool simple, single;
 
+  const char *readfmt;
   static const char writefmt[] =
-    "  write(instance.%s());\n"
-    "  break;\n";
-  static const char readfmt[] =
-    "{\n"
-    "  %1$s obj;\n"
-    "  read(str, obj);\n"
-    "  %2$s(obj, _disc_temp);\n"
-    "}\n"
-    "  break;\n";
+    "      write(str, instance.%s());\n"
+    "      break;\n";
   static const char movefmt[] =
-    "  move(str, instance.%s());\n"
-    "  break;\n";
+    "      move(str, instance.%s());\n"
+    "      break;\n";
   static const char maxfmt[] =
-    "{\n"
-    "  size_t pos = str.position();\n"
-    "  size_t alignment = str.alignment();\n"
-    "  max(str, instance.%s());\n"
-    "  union_max = std::max(str.position(),union_max);\n"
-    "  str.position(pos);\n"
-    "  str.alignment(alignment);\n"
-    "}\n";
+    "  {\n"
+    "    size_t pos = str.position();\n"
+    "    size_t alignment = str.alignment();\n"
+    "    max(str, instance.%s());\n"
+    "    if (union_max < str.position()) {\n"
+    "      union_max = str.position();\n"
+    "      alignment_max = str.alignment();\n"
+    "    }\n"
+    "    str.position(pos);\n"
+    "    str.alignment(alignment);\n"
+    "  }\n";
 
   (void)pstate;
   (void)path;
 
-  /* process case labels first */
-  if (!revisit)
+  single = (idl_degree(_case->labels) == 1);
+  simple = idl_is_base_type(_case->type_spec);
+
+  if (single)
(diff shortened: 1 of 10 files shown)


k0ekk0ek:
Did some basic manual testing with unions in the Hello World example and publishing at least seems to work. I have to carry out a couple more tests, but this is mostly done.

src/idlcxx/src/streamers.c
```diff
@@ -152,48 +152,90 @@ process_case(
 {
   struct streams *streams = user_data;
   const idl_case_t* _case = (const idl_case_t*)node;
-  char *type = NULL;
+  char *type, *value = "";
   const char *name;
+  bool simple, single;
 
   static const char writefmt[] =
```
reicheratwork, line 159: maybe replace writefmt and movefmt with:
    "      %s(str, instance.%s());\n"
    "      break;\n";

and then in the print add "write" for the write stream and "move" for the move stream?

src/idlcxx/src/streamers.c
```diff
@@ -353,65 +395,51 @@ process_switch_type_spec(
...
+    "  read(str, d);\n";
   static const char movefmt[] =
-    "  move(str, %1$s);\n"
-    "  switch(%1$s)\n{\n";
+    "  auto d = instance._d();\n"
+    "  move(str, d);\n";
   static const char maxfmt[] =
-    "  max(str, %1$s);\n"
+    "  max(str, instance._d());\n"
     "  size_t union_max = str.position();\n";
```
reicheratwork, line 408: also create a holder for the "largest case" alignment
replace with "size_t union_max = str.position(), alignment_max = str.alignment();\n";

src/idlcxx/src/streamers.c
```diff
@@ -152,48 +152,90 @@ process_case(
...
-    "  max(str, instance.%s());\n"
-    "  union_max = std::max(str.position(),union_max);\n"
-    "  str.position(pos);\n"
-    "  str.alignment(alignment);\n"
-    "}\n";
+    "  {\n"
+    "    size_t pos = str.position();\n"
+    "    size_t alignment = str.alignment();\n"
+    "    max(str, instance.%s());\n"
+    "    union_max = std::max(str.position(),union_max);\n"
```
reicheratwork, line 171: we also need to store the alignment in this case, so replace with:

"if (union_max > str.position()) {\n"
"  union_max = str.position();\n"
"  alignment_max = str.alignment();\n"
"}\n"

reicheratwork (changes requested):
One thing that i thought was missing after looking at the generated code, is that for the generated function "max" does not give the correct end result for unions:

after all the possible branches have been gone over the stream position and alignment need to be set to the value that they had with the case which gave the largest stream position

k0ekk0ek:
Build fails due to cyclonedds [#791](https://github.com/eclipse-cyclonedds/cyclonedds/pull/791) hasn't been merged yet.

src/idlcxx/src/streamers.c
reicheratwork: My apologies, the larger than sign should have been a smaller than sign, as union_max needs to be "pushed forward" if a larger value is encountered, not the other way around.

src/idlcxx/src/streamers.c
```diff
@@ -152,48 +152,93 @@ process_case(
...
-    "  max(str, instance.%s());\n"
-    "  union_max = std::max(str.position(),union_max);\n"
-    "  str.position(pos);\n"
-    "  str.alignment(alignment);\n"
-    "}\n";
+    "  {\n"
+    "    size_t pos = str.position();\n"
+    "    size_t alignment = str.alignment();\n"
+    "    max(str, instance.%s());\n"
+    "    if (union_max > str.position()) {\n"
```
reicheratwork, line 171: Should be `union_max > str.position()` since we want to select the largest, not the smallest.

reicheratwork (changes requested):
I approve these changes, if the the comparison operator in streamers.c:171 is changed.

k0ekk0ek:
Thanks @reicheratwork! Your comments have been processed :+1:

Issue #70, opened 2021-04-30
Using LoanedSamples invokes deserialization multiple times 


sumanth-nirmal:
### Current Behavior

When using the `LoanedSamples` as described below with the following version of Cyclone DDS
- cyclonedds => [idcxx branch](https://github.com/eclipse-cyclonedds/cyclonedds/tree/idlcxx), commit - [a5db83fe94e9f4afe896e335e69bc5bea741e4d8](https://github.com/eclipse-cyclonedds/cyclonedds/commit/a5db83fe94e9f4afe896e335e69bc5bea741e4d8)
- cyclonedds-cxx => [idlcxx branch](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commits/idlcxx), commit - [e8e09967876fecca97f2eba440d5448c83c8912b](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/e8e09967876fecca97f2eba440d5448c83c8912b)

```cpp
dds::sub::LoanedSamples<DDSType> dds_messages;
dds_messages = m_datareader.select().state(dds::sub::status::DataState::any()).max_samples(1).take();

if (dds_messages.length() > 0U) {
    const auto & it = dds_messages.begin();
    const auto & sample = it->data();
    const auto & sample_info = it->info();
...
# use sample and info
}
```
The same sample is deserialized twice. \
As `ddscxx_serdata` stores the deserialized sample, calling  `getT()` will anyways get the deserialized sample. However, the current code tries to deserialize the sample twice, once [here](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/e8e09967876fecca97f2eba440d5448c83c8912b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L88) in [`getT()`](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/e8e09967876fecca97f2eba440d5448c83c8912b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L178) which is called from `serdata_from_ser()` and second time [here](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/e8e09967876fecca97f2eba440d5448c83c8912b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L178) in `serdata_from_ser()`. \
I think this can be optimized to invoke `read_struct()` only once.

### Expected Behaviour

[...]

PR #74, opened 2021-05-07
Fix for cyclonedds-cxx/idlcxx issue #70

reicheratwork:
Multiple deserialization is prevented by:

- the getT() calls no longer are preceded by deserialization of
the data, since that already happens in the getT call
- added a separate setT function which creates the correct sample
in-place without having to deserialize the data as with getT, allowing
skipping of one deserialization step

@sumanth-nirmal @e-hndrks : could you check whether this works as expected and that I did not forget anything?

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
@@ -61,6 +61,7 @@ class ddscxx_serdata : public ddsi_serdata {
 
 public:
   bool hash_populated = false;
+  bool sample_filled = false;
   static const struct ddsi_serdata_ops ddscxx_serdata_ops;
   ddscxx_serdata(const ddsi_sertopic* topic, ddsi_serdata_kind kind);
   ~ddscxx_serdata() { delete m_t.load(std::memory_order_acquire); }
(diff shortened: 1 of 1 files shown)


sumanth-nirmal (changes requested):

reicheratwork:
Improved the implementation by splitting out a separate setter function `setT` which does not do unnecessary deserializations.

@sumanth-nirmal could you check this?

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -74,6 +74,23 @@ class ddscxx_serdata : public ddsi_serdata {
...
+      t = new T(*toset);
+      T* exp = nullptr;
+      if (!m_t.compare_exchange_strong(exp, t, std::memory_order_seq_cst)) {
+        delete t;
+        t = exp;
+      }
+    } else {
+      *t = *toset;
+    }
+    return t;
```
sumanth-nirmal, line 91: We don't need to return anything, correct?

reicheratwork:
@sumanth-nirmal The function could just as well return void, but as this function changes the contents of the entity, I think it is a good habit to return the thing that has changed (in this case the pointer of the contained sample).
Also the symmetry with the getT() function also returning the pointer to the sample is another reason why I thought that returning the sample pointer is a good idea.

sumanth-nirmal:
@reicheratwork I would agree. The changes look good from me. @e-hndrks Can you probably also glance over?

sumanth-nirmal (approved):

k0ekk0ek:
Included in #79 as well.

PR #76, opened 2021-05-11
improvements to templated streaming

reicheratwork:
Improvements for the following:
- array declarators not being parsed correctly
- template streamers are more compatible with user-supplied container classes
- processing of typedefs
- safety of reading bounded sequences
- more descriptive exceptions thrown
- improved generated comparison operators for structs
- added generation of comparison operators for unions

### DIFF
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -80,6 +80,35 @@ struct streams {
   size_t keys;
 };
 
+void setup_streams(struct streams* str, struct generator* gen)
+{
+  assert(str);
+  memset(str, 0, sizeof(struct streams));
+  str->generator = gen;
+}
+
+void cleanup_streams(struct streams* str)
+{
+  if (str->write.data)
+    free(str->write.data);
+  if (str->read.data)
+    free(str->read.data);
+  if (str->move.data)
+    free(str->move.data);
+  if (str->max.data)
+    free(str->max.data);
+  if (str->key_write.data)
+    free(str->key_write.data);
+  if (str->key_read.data)
+    free(str->key_read.data);
+  if (str->key_move.data)
+    free(str->key_move.data);
+  if (str->key_max.data)
+    free(str->key_max.data);
+  if (str->key.data)
+    free(str->key.data);
+}
+
 static idl_retcode_t
 process_instance(
   const idl_pstate_t *pstate,
@@ -390,6 +419,32 @@ process_struct(
 }
 
 static idl_retcode_t
+process_module(
+  const idl_pstate_t* pstate,
+  const bool revisit,
+  const idl_path_t* path,
+  const void* node,
+  void* user_data)
+{
+  struct streams* streams = user_data;
+
+  (void)pstate;
+  (void)path;
+
+  if (revisit) {
+    if (putf(&streams->key_max,"}\n\n") < 0)
+      return IDL_RETCODE_NO_MEMORY;
+  } else {
+    const char* name = get_cpp11_name(node);
+    if (putf(&streams->write, "namespace %1$s\n{\n", name) < 0)
+      return IDL_RETCODE_NO_MEMORY;
+    return IDL_VISIT_REVISIT;
+  }
+
+  return IDL_RETCODE_OK;
+}
+
+static idl_retcode_t
 process_switch_type_spec(
   const idl_pstate_t *pstate,
   bool revisit,
@@ -500,6 +555,60 @@ process_case_label(
   return IDL_RETCODE_OK;
 }
 
+static idl_retcode_t
+process_typedef(
+  const idl_pstate_t* pstate,
+  const bool revisit,
+  const idl_path_t* path,
+  const void* node,
+  void* user_data)
+{
+  (void)revisit;
+  (void)path;
+
+  const char* fmt =
+    "template<typename T>\n"
+    "void %2$s_%1$s(T& str, %1$s& instance)\n{\n";
+  const char* constfmt =
+    "template<typename T>\n"
+    "void %2$s_%1$s(T& str, const %1$s& instance)\n{\n";
+  const char* closefmt = "}\n\n";
+
+  struct streams* streams = user_data;
+  idl_typedef_t* td = (idl_typedef_t*)node;
+  const idl_declarator_t* declarator;
+
+  IDL_FOREACH(declarator, td->declarators) {
+    const char* name = get_cpp11_name(declarator);
+
+    if (putf(&streams->write, constfmt, name, "write")
+     || putf(&streams->read, fmt, name, "read")
+     || putf(&streams->move, constfmt, name, "move")
+     || putf(&streams->max, constfmt, name, "max")
+     || putf(&streams->key_write, constfmt, name, "key_write")
+     || putf(&streams->key_read, fmt, name, "key_read")
+     || putf(&streams->key_move, constfmt, name, "key_move")
+     || putf(&streams->key_max, constfmt, name, "key_max"))
+      return IDL_RETCODE_NO_MEMORY;
+
+    idl_retcode_t ret = process_instance(pstate, streams, declarator, td->type_spec, true);
+    if (ret != IDL_RETCODE_OK)
+      return ret;
+
+    if (putf(&streams->write, closefmt)
+     || putf(&streams->read, closefmt)
+     || putf(&streams->move, closefmt)
+     || putf(&streams->max, closefmt)
+     || putf(&streams->key_write, closefmt)
+     || putf(&streams->key_read, closefmt)
+     || putf(&streams->key_move, closefmt)
+     || putf(&streams->key_max, closefmt))
+      return IDL_RETCODE_NO_MEMORY;
+  }
+
+  return IDL_RETCODE_OK;
+}
+
 static idl_retcode_t flush(struct generator *gen, struct streams *streams)
 {
   if (streams->write.data && fputs(streams->write.data, gen->header.handle) < 0)
@@ -543,10 +652,26 @@ generate_streamers(const idl_pstate_t* pstate, struct generator *gen)
   idl_visitor_t visitor;
   const char *sources[] = { NULL, NULL };
 
-  memset(&streams, 0, sizeof(streams));
+  setup_streams(&streams, gen);
+
   memset(&visitor, 0, sizeof(visitor));
 
-  streams.generator = gen;
+  visitor.visit = IDL_TYPEDEF | IDL_MODULE;
+  visitor.accept[IDL_ACCEPT_TYPEDEF] = &process_typedef;
+  visitor.accept[IDL_ACCEPT_MODULE] = &process_module;
+  assert(pstate->sources);
+  sources[0] = pstate->sources->path->name;
+  visitor.sources = sources;
+
+  if ((ret = idl_visit(pstate, pstate->root, &visitor, &streams)) == IDL_RETCODE_OK)
+    ret = flush(gen, &streams);
+
+  cleanup_streams(&streams);
+
+  memset(&visitor, 0, sizeof(visitor));
+
+  setup_streams(&streams, gen);
+
   visitor.visit = IDL_STRUCT | IDL_UNION | IDL_MEMBER | IDL_CASE | IDL_CASE_LABEL | IDL_SWITCH_TYPE_SPEC | IDL_INHERIT_SPEC;
   visitor.accept[IDL_ACCEPT_STRUCT] = &process_struct;
   visitor.accept[IDL_ACCEPT_UNION] = &process_union;
@@ -555,31 +680,13 @@ generate_streamers(const idl_pstate_t* pstate, struct generator *gen)
   visitor.accept[IDL_ACCEPT_CASE_LABEL] = &process_case_label;
   visitor.accept[IDL_ACCEPT_SWITCH_TYPE_SPEC] = &process_switch_type_spec;
   visitor.accept[IDL_ACCEPT_INHERIT_SPEC] = &process_inherit_spec;
-  assert(pstate->sources);
   sources[0] = pstate->sources->path->name;
   visitor.sources = sources;
 
   if ((ret = idl_visit(pstate, pstate->root, &visitor, &streams)) == IDL_RETCODE_OK)
     ret = flush(gen, &streams);
 
-  if (streams.write.data)
-    free(streams.write.data);
-  if (streams.read.data)
-    free(streams.read.data);
-  if (streams.move.data)
-    free(streams.move.data);
-  if (streams.max.data)
-    free(streams.max.data);
-  if (streams.key_write.data)
-    free(streams.key_write.data);
-  if (streams.key_read.data)
-    free(streams.key_read.data);
-  if (streams.key_move.data)
-    free(streams.key_move.data);
-  if (streams.key_max.data)
-    free(streams.key_max.data);
-  if (streams.key.data)
-    free(streams.key.data);
+  cleanup_streams(&streams);
 
   return ret;
 }
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -451,20 +451,20 @@ process_case(
                        "      read(str, obj);\n"
                        "      instance.%2$s(obj);\n"
                        "    }\n"
-                       "      break;";
+                       "      break;\n";
   else
     readfmt = simple ? "    {\n"
                        "      %1$s obj = %3$s;\n"
                        "      read(str, obj);\n"
                        "      instance.%2$s(obj, d);\n"
                        "    }\n"
-                       "      break;"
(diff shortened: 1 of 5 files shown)


k0ekk0ek:
Thanks @reicheratwork! Changes look mostly fine. I'm currently in the process of getting API tests to run. Unfortunately not all of them pass yet. There's a memory leak and such. If you don't mind, I want to get that work in first, and then rebase this PR on that. If you're curious, the work is being done here: https://github.com/k0ekk0ek/cyclonedds-cxx/tree/port-ddscxx-tests.

src/idlcxx/src/streamers.c
```diff
@@ -106,15 +114,172 @@ process_instance(
...
+    return IDL_RETCODE_NO_MEMORY;
+
+  idl_retcode_t ret = IDL_RETCODE_OK;
+  if (idl_is_sequence(seq->type_spec))
+    ret = unroll_sequence (pstate, streams, (idl_sequence_t*)seq->type_spec, depth + 1, new_accessor, is_key);
+  else
+    ret = check_endpoint (pstate, streams, seq->type_spec, new_accessor, is_key);
+
+  if (ret != IDL_RETCODE_OK)
+    goto done;
```
k0ekk0ek, line 256: Make this an early return please.

src/idlcxx/src/streamers.c
```diff
@@ -106,15 +114,172 @@ process_instance(
...
+    if (putf(&streams->key_read, rfmt, depth, accessor)
+      || putf(&streams->key_write, wfmt, depth, accessor, maximum, "write")
+      || putf(&streams->key_move, wfmt, depth, accessor, maximum, "move")
+      || putf(&streams->key_max, mfmt, depth, maximum))
+    return IDL_RETCODE_NO_MEMORY;
+  }
+
+  char* new_accessor = NULL;
+  if (idl_asprintf(&new_accessor, "%s[i_%u]", accessor, depth) < 0)
+    return IDL_RETCODE_NO_MEMORY;
```
k0ekk0ek, line 229: Better use `IDL_PRINTA` to allow for early out.

src/idlcxx/src/streamers.c
```diff
@@ -106,15 +114,172 @@ process_instance(
...
+
+  if (is_key)
+  {
+    if (putf(&streams->key_read, rfmt, depth, accessor)
+      || putf(&streams->key_write, wfmt, depth, accessor, maximum, "write")
+      || putf(&streams->key_move, wfmt, depth, accessor, maximum, "move")
+      || putf(&streams->key_max, mfmt, depth, maximum))
+    return IDL_RETCODE_NO_MEMORY;
+  }
```
k0ekk0ek: Can you combine this into one if statement? `if (is_key && (...))`. Also, please use the style that's used throughout the generator.

src/idlcxx/src/streamers.c
```diff
@@ -106,15 +114,172 @@ process_instance(
...
+        || putf(&streams->key_read, afmt, entries)
+        || putf(&streams->key_move, afmt, entries)
+        || putf(&streams->key_max, afmt, entries))
+        return IDL_RETCODE_NO_MEMORY;
+    }
+
+    //update accessor: accessor = accessor + "[_a]"
+    char *tmp = NULL;
+    if (idl_asprintf(&tmp, "%s[_a]", accessor) < 0)
+      goto err;
```
k0ekk0ek: Please use IDL_PRINTA, it was specifically introduced to avoid having to do `alloc` + `copy` + `free` and having to include that into the logic throughout the generator.

src/idlcxx/src/streamers.c
```diff
@@ -106,15 +114,172 @@ process_instance(
...
+      || putf(&streams->max, afmt, entries))
+      return IDL_RETCODE_NO_MEMORY;
+    if (is_key)
+    {
+      if (putf(&streams->key_write, afmt, entries)
+        || putf(&streams->key_read, afmt, entries)
+        || putf(&streams->key_move, afmt, entries)
+        || putf(&streams->key_max, afmt, entries))
+        return IDL_RETCODE_NO_MEMORY;
+    }
```
k0ekk0ek: I'd simply create two functions, `process_instance` and `process_array_instance`, to keep things simple.

src/idlcxx/src/streamers.c
```diff
@@ -664,6 +672,73 @@ process_case_label(
...
+  struct streams* streams = user_data;
+  idl_typedef_t* td = (idl_typedef_t*)node;
+  const idl_declarator_t* declarator;
+  idl_retcode_t ret = IDL_RETCODE_OK;
+
+  IDL_FOREACH(declarator, td->declarators) {
+    if (IDL_PRINTA(&name, get_cpp11_fully_scoped_name, declarator, streams->generator) < 0)
+    {
+      ret = IDL_RETCODE_NO_MEMORY;
+      break;
```
k0ekk0ek: Just return directly here, much easier. Same goes for the rest of the code. Thanks.

src/idlcxx/src/streamers.c
```diff
@@ -141,23 +143,30 @@ unroll_sequence(const idl_pstate_t* pstate,
...
     "  for (uint32_t i_%1$u = 0; i_%1$u < se_%1$u; i_%1$u++) {\n";
-  const char* rfmt = "  uint32_t se_%1$u;\n"\
+  const char* rfmt = maximum ? "  uint32_t se_%1$u;\n"\
     "  read(str, se_%1$u);\n"\
-    "  %2$s.resize(se_%1$u);\n"\
+    "  for (uint32_t i_%1$u = 0; i_%1$u < se_%1$u; i_%1$u++) {\n"\
+    "  %2$s temp_%1$u;\n" : "  uint32_t se_%1$u;\n"\
+    "  read(str, se_%1$u);\n"\
+    "  %3$s.resize(se_%1$u);\n"\
     "  for (uint32_t i_%1$u = 0; i_%1$u < se_%1$u; i_%1$u++) {\n";
```
k0ekk0ek: This is mostly unreadable. Please use:
```
condition
  ? "string-constant"
  : "alternate-string-constant"
```

src/idlcxx/src/streamers.c
```diff
@@ -169,27 +178,45 @@ unroll_sequence(const idl_pstate_t* pstate,
     return IDL_RETCODE_NO_MEMORY;
 
   idl_retcode_t ret = IDL_RETCODE_OK;
+  char* new_read_accessor = NULL;
+  if (maximum)
+  {
+    if (idl_asprintf(&new_read_accessor, "temp_%u", depth) < 0)
+    {
+      ret = IDL_RETCODE_NO_MEMORY;
+      goto done;
```
k0ekk0ek, line 238: Early return please, same for the rest of the code.

src/idlcxx/src/streamers.c
```diff
@@ -169,27 +178,45 @@ unroll_sequence(const idl_pstate_t* pstate,
     return IDL_RETCODE_NO_MEMORY;
 
   idl_retcode_t ret = IDL_RETCODE_OK;
+  char* new_read_accessor = NULL;
+  if (maximum)
+  {
+    if (idl_asprintf(&new_read_accessor, "temp_%u", depth) < 0)
```
k0ekk0ek, line 235: `IDL_PRINTA`.

src/idlcxx/src/streamers.c
```diff
@@ -169,27 +178,45 @@ unroll_sequence(const idl_pstate_t* pstate,
     return IDL_RETCODE_NO_MEMORY;
 
   idl_retcode_t ret = IDL_RETCODE_OK;
+  char* new_read_accessor = NULL;
+  if (maximum)
+  {
```
k0ekk0ek, line 234: Use style consistent with the rest of the code in the generator please.

src/idlcxx/src/types.c
```diff
@@ -212,36 +212,42 @@ emit_struct(
...
+        "  %1$s() = default;\n\n";
   if (idl_fprintf(gen->header.handle, fmt, name) < 0)
     return IDL_RETCODE_NO_MEMORY;
 
-  /* constructor parameters */
-  visitor.accept[IDL_ACCEPT_DECLARATOR] = &emit_parameter;
-  if ((ret = idl_visit(pstate, _struct->members, &visitor, user_data)))
-    return ret;
+  if (_struct->members)
+  {
```
k0ekk0ek, line 226: Style.

src/idlcxx/src/streamers.c
```diff
@@ -816,44 +880,41 @@ generate_streamers(const idl_pstate_t* pstate, struct generator *gen)
...
-  visitor.visit = IDL_STRUCT | IDL_UNION | IDL_MEMBER | IDL_CASE | IDL_CASE_LABEL | IDL_SWITCH_TYPE_SPEC | IDL_INHERIT_SPEC | IDL_TYPEDEF;
+  visitor.visit = IDL_TYPEDEF | IDL_MODULE;
+  visitor.accept[IDL_ACCEPT_TYPEDEF] = &process_typedef;
+  visitor.accept[IDL_ACCEPT_MODULE] = &process_module;
+  assert(pstate->sources);
+  sources[0] = pstate->sources->path->name;
+  visitor.sources = sources;
+
+  if ((ret = idl_visit(pstate, pstate->root, &visitor, &streams)) == IDL_RETCODE_OK)
+    ret = flush(gen, &streams);
```
k0ekk0ek, line 882: Make these early out please.

k0ekk0ek (changes requested):

k0ekk0ek:
Thanks @reicheratwork. I'm mostly fine with the logic, but there are some comments. Also, I generally dislike PRs that contain commits with fixes that introduce mistakes made in the same PR. Please consolidate those. I'll have another read once the initial comments have been processed.

src/idlcxx/src/streamers.c
```diff
@@ -500,6 +555,62 @@ process_case_label(
...
+     || putf(&streams->read, fmt, name, "read")
+     || putf(&streams->move, constfmt, name, "move")
+     || putf(&streams->max, constfmt, name, "max")
+     || putf(&streams->key_write, constfmt, name, "key_write")
+     || putf(&streams->key_read, fmt, name, "key_read")
+     || putf(&streams->key_move, constfmt, name, "key_move")
+     || putf(&streams->key_max, constfmt, name, "key_max"))
+      return IDL_RETCODE_NO_MEMORY;
+
+    idl_retcode_t ret = process_instance(pstate, streams, declarator, td->type_spec, true, true);
```
k0ekk0ek, line 811: One too many argument to `process_instance`?

src/idlcxx/src/streamers.c
```diff
@@ -543,10 +654,26 @@ generate_streamers(const idl_pstate_t* pstate, struct generator *gen)
...
+  setup_streams(&streams, gen);
+
+  memset(&visitor, 0, sizeof(visitor));
+
+  visitor.visit = IDL_TYPEDEF | IDL_MODULE;
+  visitor.accept[IDL_ACCEPT_TYPEDEF] = &process_typedef;
+  visitor.accept[IDL_ACCEPT_MODULE] = &process_module;
+  assert(pstate->sources);
+  sources[0] = pstate->sources->path->name;
+  visitor.sources = sources;
```
k0ekk0ek, line 880: Visiting twice seems odd?

src/idlcxx/src/streamers.c
```diff
@@ -665,8 +879,10 @@ generate_streamers(const idl_pstate_t* pstate, struct generator *gen)
   sources[0] = pstate->sources->path->name;
   visitor.sources = sources;
 
-  if ((ret = idl_visit(pstate, pstate->root, &visitor, &streams)) == IDL_RETCODE_OK)
-    ret = flush(gen, &streams);
+  idl_retcode_t ret = IDL_RETCODE_OK;
+  if (((ret = idl_visit(pstate, pstate->root, &visitor, &streams)) != IDL_RETCODE_OK) ||
+      ((ret = flush(gen, &streams)) != IDL_RETCODE_OK))
```
k0ekk0ek, line 884: You may have to cleanup the streams here. It may have failed half-way through, probably best to add an error label or something which frees and returns `ret`.

k0ekk0ek (changes requested):
I think we're almost there. Couple of small remarks. If you fix those, I'll be happy to merge these changes. Thanks @reicheratwork!

src/idlcxx/src/streamers.c
reicheratwork: You are right, seems that two commits have become intermixed.

src/idlcxx/src/streamers.c
```diff
@@ -500,6 +555,62 @@ process_case_label(
...
+  const char* closefmt = "}\n\n";
+  const char* name = NULL;
+
+  struct streams* streams = user_data;
+  idl_typedef_t* td = (idl_typedef_t*)node;
+  const idl_declarator_t* declarator;
+
+  IDL_FOREACH(declarator, td->declarators) {
+    if ((name = get_cpp11_name(declarator)) == NULL)
+      return IDL_RETCODE_NO_MEMORY;
```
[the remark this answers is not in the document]
reicheratwork: Force of habit, I guess.
I get a pointer, I first check whether I can use it :)

src/idlcxx/src/streamers.c
reicheratwork: This was done so that the streaming functions for the typedefs would be in their own namespace, but those for the structs and unions would not be.
I chose this because the typedefs need this to be able to distinguish between different typedefs which map onto the same C++ class, which could cause the compiler to not be able to determine which of the streaming functions to use because they have the same signature.
For the structs and unions, the streaming function signatures are because the parameter which is part of the function signature is unique.

The main reason was to not have to make very long namespaces for calling the streaming functions of non-typedef instances.

src/idlcxx/src/streamers.c
reicheratwork: Check.

I will go over the paths and see when and where to clean up.

src/idlcxx/src/streamers.c
k0ekk0ek: I think I'm fine with it for this PR, but I'd prefer the functions for constructed types be placed in their own namespace as well. Same goes for the functions we use to serialize strings, sequences, etc. They're currently in the global namespace, which is bad practice when using C++ if you ask me. We'll tackle it in a follow-up PR.

src/idlcxx/src/streamers.c
k0ekk0ek: Think you forgot to resolve this one?

k0ekk0ek:
Regarding bounded sequences, like we discussed, we should reject the entire sample and not fill it as far as allowed, but let's worry about that in a follow-up PR.

One minor thing you forgot to fix, other than that I'll happily accept the changes.

src/idlcxx/src/streamers.c
reicheratwork: Changed the manner in which arrays are processed, they are now unrolled by the function unroll_array

src/idlcxx/src/streamers.c
reicheratwork: will do

src/idlcxx/src/streamers.c
reicheratwork: will change to from idl_asprintf to IDL_PRINTA wherever possible

src/idlcxx/src/streamers.c
reicheratwork: will do

src/idlcxx/src/streamers.c
reicheratwork: will do

src/idlcxx/src/streamers.c
reicheratwork: will do

src/idlcxx/src/streamers.c
reicheratwork: will do

src/idlcxx/src/streamers.c
reicheratwork: yep

src/idlcxx/src/streamers.c
reicheratwork: will do

src/idlcxx/src/streamers.c
reicheratwork: will do

src/idlcxx/src/streamers.c
reicheratwork: yes

src/idlcxx/src/streamers.c
reicheratwork: I have added all the points that still need improvement and work to issue #83

k0ekk0ek (approved):

k0ekk0ek:
Follow items are listed on #83. Some tests are known to fail, but fixing those was not part of this PR.

PR #54, opened 2020-12-17
Update to use ddsi_sertype

eboasson:
With conditional compilation for continued support of `ddsi_sertopic`.

I suppose there are tricks in C++ to make this nicer and avoid any pollution of the global namespace. Suggestions are welcome!

PR #79, opened 2021-05-14
Add tests, fix some bugs and consolidate some PRs

k0ekk0ek:
This PR adds the tests originally found in the [cdds-cxx](https://github.com/ThijsSassen/cdds-cxx/) repository and offered as a PR in #9. I've cleaned them up and used CMake's `gtest_add_tests` instead of the custom Module that was used before. Running the tests unveiled a number of issues. Some of which I fixed. Leaving a couple of issues wrt listeners and some reader tests that segfault if content filters are used. @e-hndrks is looking into the latter issue, I'll look into the listener myself, but I didn't want to postpone issuing a PR as the current status is improved (the test only shows the bug, not introduce it). The PR includes #54 retrofitted to the templated-streaming branch, as well as the changes for #70 and some of the changes in #68. I've left the original author intact for those commits. I've also added generating comparison operators for generated structs (unions will follow shortly), because the functionality was used in the tests. The idea is that this PR will form the base for #76 and any future work with the tests being enabled etc.


k0ekk0ek:
@reicheratwork, @eboasson, care to review?

reicheratwork:
It is a lot all at once, but as far as I can see, does what it claims to do.

This PR has my approval.

src/ddscxx/tests/FindDataWriter.cpp
```diff
@@ -0,0 +1,270 @@
...
+    this->CreateWriters();
+
+    cnt = dds::pub::find<
+                dds::pub::DataWriter<Space::Type1>,
+                std::back_insert_iterator<std::vector<dds::pub::DataWriter<Space::Type1> > > >(
+            this->publisher,
+            TOPIC1_NAME,
+            std::back_inserter<std::vector<dds::pub::DataWriter<Space::Type1> > >(found));
+
+#if 1
```
eboasson, line 131: Intriguing that this is different for the publisher and the subscriber

eboasson (approved):
I don't presume to know what:
```
-    if (this != static_cast<TCondition*>(&rhs)) {
+    const TCondition<DELEGATE> &t = rhs;
+    if (this != &t) {
```
is for, the original looks rather nasty to me, but then so does the replacement ...

The changes in the sources of the C++ binding look fine to me and the test code mostly, too. There is an odd discrepancy between the subscriber and the publisher with respect to the `find` operation, but that suggests a bug that can well be fixed in a different PR.

That leaves the test failures in the CI. The timeouts on Windows look like deadlocks that I suspect might be related to #55. I am ok with merging it in with these failures, especially given that (1) the changes to the language binding don't seem like they might have introduced them, and (2) now at least there is visibility. A similar reasoning works for the test failures checking callback counts.

k0ekk0ek:
Thanks @reicheratwork, @eboasson

PR #88, opened 2021-06-01
Fix issue where filtered read/take may result in a crash when returning loan

e-hndrks:
Fix issue where filtered read/take may result in uninitialized SampleRefs for loans that are dimensioned to hold all available samples, not just the filtered ones. In cases where the loan is bigger than the number of matching samples, the destructor of the loan would crash on the uninitialized SampleRef objects.


k0ekk0ek:
LGTM. Thanks @e-hndrks!

PR #89, opened 2021-06-01
DataReader with selector Forward iterator fix

### LINKED ISSUE #87: DataReader<T>::take/read passing the wrong number of max samples
In TDataReaderImpl.hpp:
The unittests DataReaderSelector.read_FWIterator_instance and  DataReaderSelector.take_FWIterator_instance show that the functions DataReader<T>::take and DataReader<T>::read pass the wrong number of max_samples to AnyDataReaderDelegate::take/read functions.
The max_samples parameter which is the maximum number of instances following the passed iterator SamplesFWIterator, is ignored. This is replaced by the max_samples_ member of selector.

The correct behaviour should be the smallest of max_samples and selector.max_samples_.

reicheratwork:
- Fixes #87
- Pass correct maximum number of samples to init_samples_buffers,
  it will not exceed the size of the container the iterator belongs to


k0ekk0ek:
LGTM. Thanks @reicheratwork!

PR #90, opened 2021-06-03
Drop CUnit and update to GoogleTest 1.10.0

k0ekk0ek:
For some reason one specific build fails after I merged #88. I have a feeling it's an outdated cache or something because I couldn't reproduce it on an Ubuntu 20.04 machine locally, but we should drop CUnit and update GoogleTest anyway. The problem does not occur on my Azure Pipeline builds so this should fix the problem.

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -66,12 +66,12 @@ steps:
     name: setup_windows
   - task: Cache@2
     inputs:
-      key: pip | 1 | $(Agent.OS)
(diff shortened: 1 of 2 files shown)


k0ekk0ek:
@reicheratwork, care to take a look?

reicheratwork:
Sure

reicheratwork (approved):

reicheratwork:
@k0ekk0ek: does pretty much what it says on the tin, approved.

PR #92, opened 2021-06-03
Save Coverity build log for debugging

k0ekk0ek:
This PR retains the Coverity build log for debugging. The C++ build does not suffer from the same problems as *cyclonedds*, but I couldn't figure out what exactly is wrong with it in a test repository. Perhaps it's the fact that it's mostly header files that do not result in object files but at this point I'm not really sure.


reicheratwork (approved):

reicheratwork:
Looks good to me.

PR #93, opened 2021-06-07
Switch Coverity Scan build to Clang

k0ekk0ek:
For some reason using GNU GCC is still problematic for C++ code. The same problem does not exist when using Clang.


k0ekk0ek:
I've verified this works correctly [here](https://dev.azure.com/k0ekk0ek/continuous-integration/_build/results?buildId=591&view=results).

eboasson (approved):

PR #94, opened 2021-06-07
Drop dependencies on idlpp-cxx and Java based IDL compiler, serialize to CDR directly

k0ekk0ek:
This PR drops the need for the Java based IDL compiler and with that [idlpp-cxx](https://github.com/ADLINK-IST/idlpp-cxx/) by adding a new serializer (credits go to @reicheratwork) that is generated by a backend for the Bison based IDL compiler currently used in Eclipse Cyclone DDS. Apart from that it also adds a great bunch of unit tests, continuous integration based on Azure Pipelines and improvements to the C++ binding itself like zero-copy mechanism for loaned read/take calls and the addition of C++ equivalents of the write_cdr/read_cdr found in *cyclonedds* (credits go to @e-hndrks).

There are still some defects to fix, tests to stabilize etc, but it should already be more stable and more performant that what is currently on the master branch. This should be really close to what will eventually be the 0.1 release (or whatever version number will be attached to the first release) of cyclonedds-cxx.


eboasson (approved):
About time, I'd say! 😀

I am not going to go over all the details in the PR: firstly, it was all reviewed before it got added to `templated-streaming`, secondly, the state of master is such that a regression can't be detected anyway.

The two failing builds on the CI fail in testcases that deal with matching, I remember having commented on such a failure in the past (something to do with running tests in parallel but not isolating them well enough). I am fine with fixing those while everything is already in master, but it would be even nicer if they could get fixed prior to merging. I'll leave that decision up to you.

k0ekk0ek:
Defects detected by Coverity Scan have all been fixed. So it's about time this got merged back to the master branch.

PR #95, opened 2021-06-07
Drop CUnit related CMake modules

k0ekk0ek:
Title says it all.


k0ekk0ek:
@reicheratwork, can you have a look at this PR? Thanks!

reicheratwork:
This removes all CUnit stuff as far as I can see, and the only unittest failures are not associated with the modified code.

Looks good to me.

PR #96, opened 2021-06-07
Fixes and cleanup

reicheratwork:
- Fixed missing includes with TopicTraits
- Fixed generation of member setter functions
- Suppressed a large amount of coverity warnings, which are not critical

### DIFF
--- a/examples/helloworld/subscriber.cpp
+++ b/examples/helloworld/subscriber.cpp
@@ -84,9 +84,11 @@ int main() {
                 std::this_thread::sleep_for(std::chrono::milliseconds(20));
             }
         }
-    }
-    catch (const dds::core::Exception& e) {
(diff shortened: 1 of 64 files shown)


reicheratwork:
@e-hndrks I think this is also something that you have a lot of expertise in, so could you please review this?

k0ekk0ek:
I'll let @e-hndrks decide on the C++ changes.

k0ekk0ek (approved):
Approved under the condition that @e-hndrks approves the newly introduced macros :+1:

e-hndrks (approved):
I think you did a great job cleaning up unnecessary code yet staying as close to the original behavior of the spec as possible. That said, the original spec seems to be inconsistent when it comes to using virtual destructors (which aren't needed when all the sub-classes add no state of their own) and so we could clean up even further. I would advice to do that in a separate iteration, which we can then also file as an issue on the OMG specification.

PR #100, opened 2021-06-11
Generated isSelfContained trait and streaming fixes

reicheratwork:
- move streaming functions to namespace org::eclipse::cyclonedds::core::cdr
- generated streaming functions for maximum size are now return immediately if they are of unbounded size
- streaming functions for typedefs avoid name clashes by adding the typedef namespace to the function name
- add the topic trait isSelfContained indicating whether the topic does not contain any pointers
- group the generated traits by namespace

### DIFF
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -511,7 +511,7 @@ print_header(FILE *fh, const char *in, const char *out)
 
 static idl_retcode_t print_guard_if(FILE* fh, const char *guard)
 {
-  static const char fmt[] =
+  static const char *fmt =
     "#ifndef %1$s\n"
     "#define %1$s\n\n";
   if (idl_fprintf(fh, fmt, guard) < 0)
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -278,16 +278,15 @@ static void copy(
   memmove(dest+off, src, cnt);
 }
 
-int get_cpp11_fully_scoped_name(
-  char *str, size_t size, const void *node, void *user_data)
+static int get_cpp11_fully_scoped_name_seps(
+  char *str, size_t size, const void *node, const char *sep)
 {
-  const char *name, *sep = "::";
+  const char *name;
   size_t cnt, off, len = 0;
   static const idl_mask_t mask =
     IDL_MODULE | IDL_STRUCT | IDL_UNION | IDL_ENUM |
     IDL_ENUMERATOR | IDL_DECLARATOR;
 
-  (void)user_data;
   assert(str && size);
 
   for (const idl_node_t *n = node; n; n = n->parent) {
@@ -316,6 +315,20 @@ int get_cpp11_fully_scoped_name(
   str[ (len < size ? len : size - 1) ] = '\0';
 
   return (int)len;
+}
+
+int get_cpp11_fully_scoped_name(
+  char *str, size_t size, const void *node, void *user_data)
+{
+  (void)user_data;
+  return get_cpp11_fully_scoped_name_seps(str, size, node, "::");
+}
+
+int get_cpp11_name_typedef(
+  char *str, size_t size, const void *node, void *user_data)
+{
+  (void)user_data;
+  return get_cpp11_fully_scoped_name_seps(str, size, node, "_");
 }
 
 int get_cpp11_type(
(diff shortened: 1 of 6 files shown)


src/idlcxx/src/streamers.c
```diff
@@ -992,6 +960,47 @@ process_case_label(
...
 }
 
+static idl_retcode_t
+process_typedef_decl(
+  const idl_pstate_t* pstate,
+  struct streams* streams,
+  const idl_type_spec_t* type_spec,
+  const idl_declarator_t* declarator)
+{
+  static const instance_location_t loc = { .parent = "instance", .type = TYPEDEF | NORMAL_INSTANCE | KEY_INSTANCE };
```
k0ekk0ek, line 1031: `process_instance` takes an `instance_location_t` rather than a `const instance_location_t`

src/idlcxx/src/streamers.c
```diff
@@ -124,6 +124,8 @@ struct streams {
   idl_buffer_t key_move;
   idl_buffer_t key_max;
   size_t keys;
+  bool key_max_sz_unlimited;
+  bool max_sz_unlimited;
```
k0ekk0ek, line 128: Am I right that currently all types are printed to the buffer in one go (I believe that was one of your last PRs)? If so, I think this doesn't fit very well with that as whether a type is "unlimited" is per-type and so this introduces complexity.

src/idlcxx/src/generator.c
```diff
@@ -278,16 +278,15 @@ static void copy(
   memmove(dest+off, src, cnt);
 }
 
-int get_cpp11_fully_scoped_name(
+int get_cpp11_fully_scoped_name_seps(
   char *str, size_t size, const void *node, void *user_data)
 {
-  const char *name, *sep = "::";
+  const char *name, *sep = user_data;
```
k0ekk0ek, line 284: Probably best to change `void *user_data` to `const char *sep`. Don't think the function is used for `IDL_PRINTA` itself, so no point in keeping it generic. Also, probably best to change it to `static` because it's only used within this one file.

k0ekk0ek (changes requested):

k0ekk0ek:
Thanks @reicheratwork. Looks mostly ok, just a few minor nitpicks.

src/idlcxx/src/generator.c
reicheratwork: You are correct in this, it will be changed.

src/idlcxx/src/streamers.c
```diff
@@ -263,14 +274,14 @@ write_basic_type_streaming_functions(
...
   instance_location_t loc)
 {
-  const char* fmt = "  ::org::eclipse::cyclonedds::core::cdr::%2$s(streamer, %1$s);\n";
+  const char* fmt = "  %2$s(streamer, %1$s);\n";
   const char* read_fmt = fmt;
   if ((idl_mask(type_spec) & IDL_BOOL) == IDL_BOOL &&
       loc.type & SEQUENCE) {
     read_fmt =
       "  {\n"
       "    bool b;\n"
```
sumanth-nirmal, line 283: Since `b` is not initialized here, but it is assigned to another variable in the lines below. The compiler would warn with  "may be used uninitialized in this function" (though b gets initialized in the streamer function in the below line. Can we default initialize this bool value?

src/idlcxx/src/streamers.c
reicheratwork: I will correct this.

src/idlcxx/src/streamers.c
reicheratwork: I think you reached the wrong conclusion here.
If the unlimited size flag is set, the contents for the (key_)max_size function are discarded and just replaced with setting the stream cursor to the "unlimited" position (SIZE_MAX), in stead of filling it with the usual operations relating to any members.
Upon completing a struct, union, or typedef, and completing the tree, all streaming function contents are flushed to the file and the unlimited size flags are reset to false. This way each constructed type's functions will not be influenced by the previous ones' max_size_unlimited property.

src/idlcxx/src/streamers.c
reicheratwork: Sure, I will add initialization of the variable on creation.

k0ekk0ek (approved):

PR #105, opened 2021-06-24
Unittest fixes

reicheratwork:
Fixes for to get previously disabled unittests working, separated by type.

After the sub-commits are approved, will merge them into 1 or 2 commits and then create a real PR.

PR #106, opened 2021-06-24
Use unique topic name per test

k0ekk0ek:
A shameless rip from the tests for ddsc in cyclonedds to ensure tests that require it use a unique topic name to avoid intermittent failures. Using the command `ctest -T test -R EntityStatus -j 4 --repeat until-fail:1000` didn't show any failure when testing locally whereas it failed 2 out of 3 times before. It's meant to be an addition to the work @reicheratwork did on #105 to ensure builds for PRs are reliable.

### DIFF
--- a/src/ddscxx/tests/Listener.cpp
+++ b/src/ddscxx/tests/Listener.cpp
@@ -18,6 +18,7 @@
 #include "dds/ddsrt/sync.h"
 #include "dds/ddsrt/threads.h"
 
+#include "Util.hpp"
 #include "HelloWorldData.hpp"
 
 static uint32_t cb_called = 0;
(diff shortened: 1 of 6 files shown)


k0ekk0ek:
Check results say enough, but can you take quick look @reicheratwork?

src/ddscxx/tests/Listener.cpp
```diff
@@ -308,8 +309,7 @@ class Listener : public ::testing::Test
     }
 
     void SetUp() {
-        char buf[32];
-        ddsrt_pid_t pid = ddsrt_getpid();
+        char name[32];
```
reicheratwork, line 312: 32 characters is suffcient?

Not that I am really complaining, we have to give the string some size, just that this introduces a possibility of failure which can be triggered by giving a too large string.

reicheratwork (approved):
Does what it says on the box, so this gets my approval.

PR #112, opened 2021-06-29
Bounded string length fix

### LINKED ISSUE #109: Streamer aborts with max bound for bounded strings
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Cyclone version:**
- cyclonedds - master, [779c58](https://github.com/eclipse-cyclonedds/cyclonedds/commit/779c58206d6240a2cc405eece9a66bf70b97cd00)
- cyclonedds-cxx - master, 624da660d9d2c725555bd65bf5ff5c879dbee77b

When a bounded string with max bound is used the streamer aborts with an exception

**Detailed reproducer**

When using the following IDL file:

```hpp
module msg
{
struct log {
string<255> text_;
};
};
```
The streamer aborts when the bounded string is used with the maximum bound as below:
```cpp
TEST(tooch, bound)
{
  using DDSType = msg::log;
  dds::domain::DomainParticipant dp(0);
  dds::pub::Publisher pub(dp);
  dds::topic::Topic<DDSType> topic(dp, "test");
  dds::pub::DataWriter<DDSType> dw(pub, topic);
  DDSType msg{};
  msg.text_(
  "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx");
  dw.write(msg); // => Throws an exception
}
```

This throws an exception as below:

```sh
unknown file: Failure
C++ exception with description "Error Bad Parameter - write failed.
===============================================================================
Context     : void org::eclipse::cyclonedds::pub::AnyDataWriterDelegate::write
Node        : UnknownNode
" thrown in the test body.
```

The streamer should gracefully handle the bounded strings with the maximum bound

reicheratwork:
- This fixes #109
- The terminating null character in a string is no longer counted for
  the bounded length
- Add unittest to check the correctness of the fix


src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
```diff
@@ -253,8 +253,11 @@ void read_string(basic_cdr_stream& str, T& toread, size_t N)
 
   read(str, string_length);
 
+  //the string length in the CDR stream includes the terminating NULL
+  //therefore the checks on the string length decrease it by 1
+
   if (N &&
-      string_length > N &&
+      string_length - 1 > N &&
```
sumanth-nirmal, line 260: with an empty string, doesn't the `string_length - 1` lead to some large unintended value?

reicheratwork: It may seem that way, but this won't happen, because the length field read from the CDR stream includes the terminating NULL character (CORBA v3.3 section 9.3.2.7), unlike the value from std::string::length.

k0ekk0ek (approved):

PR #113, opened 2021-06-29
Add required headers to generated code

### LINKED ISSUE #108: Missing headers in the generated code
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Cyclone version:**
- cyclonedds - master, [779c58](https://github.com/eclipse-cyclonedds/cyclonedds/commit/779c58206d6240a2cc405eece9a66bf70b97cd00)
- cyclonedds-cxx - master, 624da660d9d2c725555bd65bf5ff5c879dbee77b

The generated code doesn't include the required headers in the following scenarios (however, things might work because these get included with some other transient includes)

1. [ ] The header `<cstdint>` is not included, which is required for the basic types like `uint8_t`, `uint16_t` etc.
2. [ ] The header `<array>` is not included when the types are array
    For the following IDL file
    ```hpp
    module test_msgs {
        module msg {
          module dds_ {
    
              struct Arrays_ {
                boolean bool_values_[3];
                octet byte_values_[3];
                octet char_values_[3];
              };
          };  // module dds_
        };  // module msg
    };  // module test_msgs
    ```
    The generated code doesn;t include the array header
    ```hpp
    #ifndef DDSCXX_ARRAY_HPP
    #define DDSCXX_ARRAY_HPP
    
    namespace test_msgs
    {
    namespace msg
    {
    namespace dds_
    {
    class Arrays_
    {
    private:
      std::array<bool, 3> bool_values__ = { };
      std::array<uint8_t, 3> byte_values__ = { };
      std::array<uint8_t, 3> char_values__ = { };
    ...
    ```
    
3. [ ] The string header is not included when the type is a string const
    For the IDL file below:
    ```hpp
    module msg
    {
        module PrimitivesConstants_Constants {
      const string<3> STRING_CONST_ = "foo";
    };

    struct PrimitivesConstants_ {
      octet structure_needs_at_least_one_member_;
    };

    };  // module msg
    ```
    The generated code doesn't include the header for the string
    ```hpp
    #ifndef DDSCXX_TOOCH_HPP
    #define DDSCXX_TOOCH_HPP
    
    namespace msg
    {
    namespace PrimitivesConstants_Constants
    {
    const std::string STRING_CONST_ = "foo";
    
    }
    
    class PrimitivesConstants_
    {
    private:
      uint8_t structure_needs_at_least_one_member__ = 0;
    ```
    1. [ ] This is also true when a custom container for the string is passed to the IDL compiler as follows `-f bounded-string-template="MyString<{BOUND}>" -f bounded-string-include="<my_string.hpp>"`
4. [ ] The generated code has `ddscxx_sertype`, but there is no include nor a forward decleration for this in the generated code
5. [ ] The generated code uses `write`/`read`.. from the [basic_cdr_streamer](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/624da660d9d2c725555bd65bf5ff5c879dbee77b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp), but doesn't include this `basic_cdr_ser.hpp` in the generated code.

The generated code should correctly include the required headers in all scenarios

k0ekk0ek:
* Include cstdint header if integer types or octets are used
* Include array header for anonymous arrays
* Include string header if string constants are used
* Include datatopic.hpp from TopicTraits.hpp

Fixes #108. @sumanth-nirmal, can you verify this fixes it for you?

Requires [PR 863 in cyclonedds](https://github.com/eclipse-cyclonedds/cyclonedds/pull/863) to be merged for the string constant issue.

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/TopicTraits.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/TopicTraits.hpp
@@ -20,7 +20,7 @@
 
 #include <vector>
 
-#include <org/eclipse/cyclonedds/topic/DataRepresentation.hpp>
+#include "org/eclipse/cyclonedds/topic/DataRepresentation.hpp"
 
 struct ddsi_sertype;
 
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -578,7 +578,7 @@ register_types(
   if (src && strcmp(loc->first.source->path->name, src) != 0)
     return IDL_VISIT_DONT_RECURSE;
 
-  if (idl_is_array(type_spec))
+  if (idl_is_array(node) || idl_is_array(type_spec))
     gen->uses_array = true;
 
   type_spec = idl_unalias(idl_type_spec(node), IDL_UNALIAS_IGNORE_ARRAY);
@@ -601,6 +601,10 @@ register_types(
       gen->uses_string = true;
   } else if (idl_is_union(type_spec)) {
     gen->uses_union = true;
+  } else if (idl_is_integer_type(type_spec)) {
+    gen->uses_integers = true;
+  } else if (idl_type(type_spec) == IDL_OCTET) {
+    gen->uses_integers = true;
   }
 
   /* FIXME: add support for @optional */
@@ -619,9 +623,10 @@ generate_includes(const idl_pstate_t *pstate, struct generator *generator)
 
   /* determine which "system" headers to include */
   memset(&visitor, 0, sizeof(visitor));
-  visitor.visit = IDL_DECLARATOR | IDL_SEQUENCE | IDL_UNION;
+  visitor.visit = IDL_DECLARATOR | IDL_SEQUENCE | IDL_UNION | IDL_CONST;
   visitor.accept[IDL_ACCEPT_DECLARATOR] = &register_types;
   visitor.accept[IDL_ACCEPT_SEQUENCE] = &register_types;
+  visitor.accept[IDL_ACCEPT_CONST] = &register_types;
   visitor.accept[IDL_ACCEPT_UNION] = &register_union;
   assert(pstate->sources);
   sources[0] = pstate->sources->path->name;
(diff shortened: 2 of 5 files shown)


eboasson (approved):
LGTM but (1) I am not too familiar with the intricacies of the IDL compiler and (2) the proof of the pudding is in the eating. So let's see whether @sumanth-nirmal confirms it works for him before merging 🙂

src/ddscxx/include/org/eclipse/cyclonedds/topic/TopicTraits.hpp
```diff
@@ -20,7 +20,8 @@
 
 #include <vector>
 
-#include <org/eclipse/cyclonedds/topic/DataRepresentation.hpp>
+#include "org/eclipse/cyclonedds/topic/DataRepresentation.hpp"
+#include "org/eclipse/cyclonedds/topic/datatopic.hpp"
```
sumanth-nirmal, line 24: This will now get the cyclone DDS headers (from `ddsi`, `ddsrt`) which I guess will eventually lead to `pedantic` errors in the generated code because of definitions like `struct sockaddr_storage addrs[];` in `ddsrt/sockets.h`

However, the fix for the inclusion of `<cstdint>`, `<array>`, `<string>` for string const work. I am not sure what's the best way to include the headers for `ddscxx_sertype` without pulling the headers from cyclone DDS core!

Also the header `basic_cdr_ser.hpp` for and `write`/`read`.. is not included.

k0ekk0ek: `basic_cdr_ser.hpp` is included indirectly (via TopicTraits.hpp from the top of my mind), so we should be good on that front.

k0ekk0ek: The builds don't show any warnings and I've built everything with `-Wpedantic` and GCC 11.1 myself. Before I go of and check everything (which I'll happily do), is the fact that warnings will occur a hunch or can you reproduce?

sumanth-nirmal: @k0ekk0ek It was my hunch, now I have tried compiling with this change and I don't see any pedantic errors (which I was expecting) so this should be good

sumanth-nirmal: At least in my environment, this works as expected (maybe some includes are coming through other transient include files)

Issue #119, opened 2021-07-01
Endianness not being taken into account when deserializing CDR data


reicheratwork:
When writing, the endianness byte (number 1) is set to `0x01` (e.g.:`datatopic.hpp::serdata_from_sample`) for little endian systems, but when reading, this is not taken into account.

reicheratwork:
This issue can be closed, as it has been fixed in commit 83240b3.

Issue #117, opened 2021-06-30
Deserialization of strings is incorrect


sumanth-nirmal:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Cyclone version:**
- cyclonedds - master, [779c58](https://github.com/eclipse-cyclonedds/cyclonedds/commit/779c58206d6240a2cc405eece9a66bf70b97cd00)
- cyclonedds-cxx - master, 624da660d9d2c725555bd65bf5ff5c879dbee77b

**Observed result or behavior:**

The deserialization doesn't happen correctly with the sequence/array of strings

**Detailed reproducer**

When using the following IDL file

```hpp
module my_msg {
  module dds {
    typedef string string__3[3];
    struct msg {
      string__3 string_values;
    };
  };
};
```
in the code as below:
```cpp
TEST(tooch, deserialize_string)
{
  using DDSType = my_msg::dds::msg;

  DDSType deserialized_msg{};

  // serialized buffer
  uint8_t serialized_buf[] =
  {0x01, 0x00, 0x01, 0x00, 0x0f, 0x00, 0x00, 0x00, 0x43, 0x6f, 0x6d, 0x70, 0x6c, 0x65, 0x78, 0x20,
    0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x31, 0x00, 0x00, 0x0f, 0x00, 0x00, 0x00, 0x43, 0x6f, 0x6d,
    0x70, 0x6c, 0x65, 0x78, 0x20, 0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x32, 0x00, 0x00, 0x0f, 0x00,
    0x00, 0x00, 0x43, 0x6f, 0x6d, 0x70, 0x6c, 0x65, 0x78, 0x20, 0x48, 0x65, 0x6c, 0x6c, 0x6f,
    0x33, 0x00, 0x00};

  //deserialize the buffer
  org::eclipse::cyclonedds::core::cdr::basic_cdr_stream cdr_streamer;
  cdr_streamer.set_buffer(serialized_buf + 4); // ignore the header
  org::eclipse::cyclonedds::core::cdr::read(cdr_streamer, deserialized_msg);
  EXPECT_FALSE(cdr_streamer.abort_status());
  EXPECT_STREQ(deserialized_msg.string_values()[0].c_str(), "Complex Hello1");
[...]

Issue #115, opened 2021-06-30
Serialization is incorrect with the updated streamer


sumanth-nirmal:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Cyclone version:**
- cyclonedds - master, [779c58](https://github.com/eclipse-cyclonedds/cyclonedds/commit/779c58206d6240a2cc405eece9a66bf70b97cd00)
- cyclonedds-cxx - master, 624da660d9d2c725555bd65bf5ff5c879dbee77b

**Observed result or behaviour:**

When an IDL type contains members with size > 1 byte, and when the value is 0 the serialization doesn't happen correctly

**Detailed reproducer**

When using the following IDL file

```hpp
module msg
{
struct log {
boolean bool_values_[3];
octet byte_values_[3];
octet char_values_[3];
float float32_values_[3];
};
};
```
in the code as below:
```cpp
TEST(tooch, serialize_deserialize)
{
  using DDSType = msg::log;

  // data
  DDSType msg{};
  msg.bool_values_(std::array<bool, 3>{false, true, false});
  msg.byte_values_(std::array<uint8_t, 3>{0, 255, 0});
  msg.char_values_(std::array<uint8_t, 3>{0, 255, 0});
  msg.float32_values_(std::array<float, 3>{0, 1.345, -12.34});

  org::eclipse::cyclonedds::core::cdr::basic_cdr_stream cdr_streamer;
  uint8_t buffer[1024];
  // fill the buffer to some non-zero to see the effect
  std::memset(buffer, 0xaa, sizeof(buffer));
  {
    //serialize
    cdr_streamer.set_buffer(buffer);
    org::eclipse::cyclonedds::core::cdr::write(cdr_streamer, msg);
    EXPECT_FALSE(cdr_streamer.abort_status());
  }
}
```
The serialized buffer looks as below:

```sh
00 01 00 00   ff 00 00 ff   00 00 00 00   00 aa aa aa   │ ················ │
 3f ac 28 f6   c1 45 70 a4 
```

The correct serialized buffer should look as below:

```sh
00 01 00 00   ff 00 00 ff   00 00 00 00   00 00 00 00   │ ················ │
 f6 28 ac 3f   a4 70 45 c1
```

I guess this is because of 2 reasons
[...]

PR #120, opened 2021-07-02
Serialization fixes

reicheratwork:
This fixes a number of issues in the CDR serialization, namely issues #115, #117 and #119:
- Alignment
- Byte Swapping on read side
and adds unittests to check for their correctness.


sumanth-nirmal:
I have tested this locally and does what it is supposed to do (the reproducers in #115 and #117 works as expected with this fix).

sumanth-nirmal (approved):

eboasson (approved):
I decided I really need to gain some familiarity with the C++ (de)ser and type definition generation, and decided to do so by exercising things a bit and seeing where they break. It won't come as a surprise to anyone that I came to the same conclusion. 🙂 Thanks @sumanth-nirmal and @reicheratwork!

PR #123, opened 2021-07-02
 Adds API to get status of DDScObjectDelegate

### LINKED ISSUE #97: API to get status of DDScObjectDelegate
## Brief feature description

While using the Cyclone DDS CXX API, I realized that having an API like `is_valid()` (vendor-specific API) for `DDScObjectDelegate` will be helpful to check if the object is valid (checking if it is closed) before accessing the object.

## Detailed information

With the current API, when I want to get a ddsc entity of an object then I need to do something like below:
```cpp
ds::core::cond::StatusCondition & condition{};
auto foo = condition.delegate()->get_ddsc_entity();
...
```

The `get_ddsc_entity()` will internally check if the object is closed and throws an exception if it is closed, which is good in most scenarios, but in some use-cases, it is good (or required) to check if the object is valid before accessing anything from this object, something like below
```cpp
ds::core::cond::StatusCondition & condition{};
if (condition.delegate()->is_valid() {
  auto foo = condition.delegate()->get_ddsc_entity();
  ...
}
```
---

From the code, it looks like it is easier to return this variable `closed` (of course guaranteeing the thread-safe behavior) to know the status of the delegate

https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/ae84dd0962d375d70143dfe346db198da32bb374/src/ddscxx/src/org/eclipse/cyclonedds/core/ObjectDelegate.cpp#L57

sumanth-nirmal:
Fixed #97

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/DDScObjectDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/DDScObjectDelegate.hpp
@@ -51,6 +51,7 @@ public:
     dds_entity_t get_ddsc_entity ();
     void set_ddsc_entity (dds_entity_t e);
     void add_to_entity_map (org::eclipse::cyclonedds::core::ObjectDelegate::weak_ref_type weak_ref);
+    bool is_valid () const;
 
 public:
     static ObjectDelegate::ref_type extract_strong_ref(dds_entity_t e);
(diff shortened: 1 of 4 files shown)


sumanth-nirmal:
@k0ekk0ek @reicheratwork @e-hndrks The changes in this MR are in the lines of what I was expecting for this issue #97 \
Please take a look to see if it make sense.

PR #125, opened 2021-07-06
Extensions to support shm

e-hndrks:


eboasson:
@e-hndrks methinks this unintentionally includes various commits and needs to be rebased.

examples/helloworld_shm/HelloWorldDataShm.idl
```diff
@@ -0,0 +1,19 @@
...
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+module HelloWorldDataShm
+{
+  struct Msg
+  {
+    long data;
```
reicheratwork, line 16: It is enough to just have this small of a message?

Maybe it is better to have a message with a huge payload to better illustrate the performance increase iceoryx can cause?

src/ddscxx/include/dds/pub/detail/DataWriter.hpp
```diff
@@ -69,6 +69,10 @@ class dds::pub::detail::DataWriter : public ::org::eclipse::cyclonedds::pub::Any
 
     void init(ObjectDelegate::weak_ref_type weak_ref);
 
+    T& loan_sample();
+
+    void return_loan(T& sample);
```
reicheratwork, line 74: This does not overlap with @e-hndrks work on the loan interface?

reicheratwork (changes requested):
Maybe have some larger messages in the example program to better illustrate the performance improvements?

e-hndrks:
Hi @e-hndrks , the reason why this is not a stand-alone Pull Request is that it builds on top of the changes that have already been made for the loan_sample API, which is therefore a prerequisite for this PR. However, the loan_sample API is currently part of a PR from @sumanth-nirmal, which hasn't been accpeted yet. For that reason, I started from his working branch.
Maybe that was not the smartest approach, but what would be a good alternative?

PR #129, opened 2021-07-06
Fix idlcxx two backend bugs

eboasson:
These two commits fix a generation issue in the type definitions and one in the generated serializer. Both deal with unions, both result in a uncompilable code.


k0ekk0ek (approved):

k0ekk0ek:
Test failure is probably a result of two tests being run simultaneously as the DataReaderSelector test doesn't set a custom topic per test (I did the ones that failed, but I really should have done all of them).

PR #130, opened 2021-07-07
Fixes for issues with typedef of base types in type support generator

k0ekk0ek:
This fixes the setters (and initializers) for members that use type specifiers that are typedefs of base types or enumerators and fixes issues where the number of cases for setters suddenly changes. Still have to write tests, so put in draft mode.


k0ekk0ek:
Carried out some manual tests and I think we're good, so this is ready for review. Also added a return value check to fix a Coverity issue.

eboasson (approved):
LGTM

PR #131, opened 2021-07-08
Always set the discrimant value to the one read

### LINKED ISSUE #128: Incorrect deserialization of unions with discriminants not mapping to a union case
This:
```
union a0005 switch (char) {
  case 'A':     a0006 a0009;
  case 'B':     double a0010;
};
```
generates a deserialization function:
```C++
template<typename T>
void read(T& streamer, ::a0005& instance)
{
  auto d = instance._d();
  read(streamer, d);
  switch(d)
  {
    case 'A':
  {
    ::a0006 obj;
  read__a0006(streamer, obj);
    instance.a0009(obj);
  }
  break;
    case 'B':
  {
    double obj = 0.0;
  read(streamer, obj);
    instance.a0010(obj);
  }
  break;
  }
  (void)streamer;
  (void)instance;
}
```
which means that it doesn't set the discriminant to the actual value read if it is neither 'A' nor 'B'. E.g., if the discriminant in the input is actually 'C', this will return a deserialized object with the discriminant set to 0 because of:
```C++
class a0005
{
private:
  char m__d;

  std::variant<::a0006, double> m__u;

  static const char _default_discriminator = 0x0;

  static char _is_discriminator(const char d)
  {
    switch (d) {
      case 'A':
        return 'A';
      case 'B':
        return 'B';
    }
    return _default_discriminator;
  }
```

eboasson:
This forces the discriminant value to be the actual one read from the
CDR, not the default value associated with the setter we used.  E.g.:

  union U switch (long) {
    case 1: case 2: long a;
    default: string b;
  };

setting a value with a(...) will initialize the discriminant to 1, even
when the intended discrimant value is 2; the same problem exists with
the default case, as well as when where there is no "default:" but the
discrimant value doesn't map to any case.

Note that this fix is suboptimal: for the vast majority of the unions,
this doesn't come into play and this merely sets the discrimant value
the same value twice.

This fixes #128 according to the (improved and even nastier) https://gist.github.com/eboasson/62632a6fcfdde7839b3230856d205a38

### DIFF
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -978,6 +978,21 @@ process_union(
     return IDL_RETCODE_NO_MEMORY;
 
   if (revisit) {
+    // Force the discriminant value to be the actual one read from the CDR, not the
+    // default value associated with the setter we used.  E.g.:
+    //   union U switch (long) {
+    //     case 1: case 2: long a;
+    //     default: string b;
+    //   };
+    // setting a value with a(...) will initialize the discriminant to 1, even when
+    // the intended discrimant value is 2; the same problem exists with the default
+    // case, as well as when where there is no "default:" but the discrimant value
+    // doesn't map to any case.
+    //
+    // FIXME: for the vast majority of the unions, this doesn't come into play this
+    // simply wastes time.
+    putf(&streams->read, "  instance._d(d);\n");
+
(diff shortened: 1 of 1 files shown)


k0ekk0ek (approved):

k0ekk0ek:
LGTM

Issue #133, opened 2021-07-12
Convenience Function missing for dds::core::policy::Liveliness::Automatic?


Kischy:
There is a function with a `const dds::core::Duration& lease_duration` parameter for Liveliness ManualByParticipant and ManualByTopic.

```cpp
template <typename D>
TLiveliness<D> TLiveliness<D>::ManualByParticipant(const dds::core::Duration& lease_duration)
{
    return TLiveliness(dds::core::policy::LivelinessKind::MANUAL_BY_PARTICIPANT, lease_duration);
}

template <typename D>
TLiveliness<D> TLiveliness<D>::ManualByTopic(const dds::core::Duration& lease_duration)
{
    return TLiveliness(dds::core::policy::LivelinessKind::MANUAL_BY_TOPIC, lease_duration);
}

```

For Liveliness Automatic I only found a parameterless function
```cpp
template <typename D>
TLiveliness<D> TLiveliness<D>::Automatic()
{
    return TLiveliness(dds::core::policy::LivelinessKind::AUTOMATIC, dds::core::Duration::infinite());
}
```

Should'nt the function for Liveliness Automatic also have a parameter for the lease_duration? The OMG standard says for LIVELINESS::AUTOMATIC:

> The infrastructure will automatically signal
> liveliness for the DataWriters at least as
> often as required by the lease_duration.

So the lease_duration is used with liveliness automatic.

eboasson:
> So the lease_duration is used with liveliness automatic

and indeed it is in the underlying implementation!

As far as I can tell, this detail was inherited from the C++ API implementation of OpenSplice. Interesting. Anyway, it looks to me like overloading the function by adding one that does take take `lease_duration` should be pretty straightforward. Having a variant of the function that defaults the lease duration to ∞ doesn't seem like it would be confusing or risky to me.

PR #134, opened 2021-07-12
Added Liveliness Automatic convenience function with lease_duration paramter.

Kischy:
Fixes Issue #133


eboasson (approved):
Looks good to me, thanks @Kischy !

Kischy:
@eboasson Hmm... why did the documentation fail here?

eboasson:
I'm not quite sure. As far as I can tell your PR is valid C++ and does what it aims to do, and therefore I suspect it is a limitation or bug of `sphinx` or some related part. They're even asking for a bug report!

Anyway, we do rely on the auto-generation of the documentation nowadays and so that needs to be looked into.

Kischy:
Ok, thanks for the explanation.

PR #135, opened 2021-07-16
Add builtin topics support

reicheratwork:
This will add builtin topics support for C++.

While the C++ message classes for the builtin topic were already created, there was no way to create a reader or a topic for them.

This PR adds:
- the creation of the C++ builtin topics
- the creation of the C++ readers for the builtin topics
- translation from the C to the C++ builtin topic message representation
- unittests for the builtin topics

@e-hndrks, @k0ekk0ek, @sumanth-nirmal:
Could you please check whether this adds the correct functionality and follows the correct locations and structuring of the features?

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicTraits.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicTraits.hpp
@@ -16,6 +16,142 @@
 #include <org/eclipse/cyclonedds/topic/TopicTraits.hpp>
 #include <org/eclipse/cyclonedds/topic/BuiltinTopicCopy.hpp>
 
-/* TODO: add builtin topic traits. */
+namespace org
+{
+namespace eclipse
+{
+namespace cyclonedds
+{
+namespace topic
+{
+
+template <>
+class TopicTraits<dds::topic::ParticipantBuiltinTopicData>
+{
+public:
+  static bool isKeyless()
+  {
+    return true;
+  }
+
+  static const char *getTypeName()
+  {
+    return "dds::topic::ParticipantBuiltinTopicData";
+  }
+
+  static ddsi_sertype *getSerType()
+  {
+    auto *st = new ddscxx_sertype<dds::topic::ParticipantBuiltinTopicData>();
+    return static_cast<ddsi_sertype*>(st);
+  }
+
+  static size_t getSampleSize()
+  {
+    return sizeof(dds::topic::ParticipantBuiltinTopicData);
+  }
+
+  static bool isSelfContained()
+  {
+    return false;
+  }
+};
+
+template <>
+class TopicTraits<dds::topic::TopicBuiltinTopicData>
+{
+public:
+  static bool isKeyless()
+  {
+    return true;
+  }
+
+  static const char *getTypeName()
+  {
+    return "dds::topic::TopicBuiltinTopicData";
+  }
+
+  static ddsi_sertype *getSerType()
+  {
+    auto *st = new ddscxx_sertype<dds::topic::TopicBuiltinTopicData>();
+    return static_cast<ddsi_sertype*>(st);
+  }
+
+  static size_t getSampleSize()
+  {
+    return sizeof(dds::topic::TopicBuiltinTopicData);
+  }
+
+  static bool isSelfContained()
+  {
+    return false;
+  }
+};
+
+template <>
+class TopicTraits<dds::topic::PublicationBuiltinTopicData>
+{
+public:
+  static bool isKeyless()
+  {
+    return true;
+  }
+
+  static const char *getTypeName()
+  {
+    return "dds::topic::PublicationBuiltinTopicData";
+  }
+
+  static ddsi_sertype *getSerType()
+  {
+    auto *st = new ddscxx_sertype<dds::topic::PublicationBuiltinTopicData>();
+    return static_cast<ddsi_sertype*>(st);
+  }
+
+  static size_t getSampleSize()
+  {
+    return sizeof(dds::topic::PublicationBuiltinTopicData);
+  }
+
+  static bool isSelfContained()
+  {
+    return false;
+  }
+};
+
+template <>
+class TopicTraits<dds::topic::SubscriptionBuiltinTopicData>
+{
+public:
+  static bool isKeyless()
+  {
+    return true;
+  }
+
+  static const char *getTypeName()
+  {
+    return "dds::topic::SubscriptionBuiltinTopicData";
+  }
+
+  static ddsi_sertype *getSerType()
+  {
+    auto *st = new ddscxx_sertype<dds::topic::SubscriptionBuiltinTopicData>();
+    return static_cast<ddsi_sertype*>(st);
+  }
+
+  static size_t getSampleSize()
+  {
+    return sizeof(dds::topic::SubscriptionBuiltinTopicData);
+  }
+
+  static bool isSelfContained()
+  {
+    return false;
+  }
+};
+
+}  /* topic */
+}  /* cyclonedds */
+}  /* eclipse */
+}  /* org */
 
 #endif /* CYCLONEDDS_TOPIC_BUILTIN_TOPIC_TRAITS_HPP */
--- a/src/ddscxx/src/dds/topic/BuiltinTopic.cpp
+++ b/src/ddscxx/src/dds/topic/BuiltinTopic.cpp
@@ -1,0 +1,94 @@
+/*
+ * Copyright(c) 2021 ADLINK Technology Limited and others
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+ * v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+
+/**
+ * @file
+ */
+
+#include <dds/topic/BuiltinTopic.hpp>
+
+namespace dds
+{
+
+namespace topic
+{
+
+class default_builtin_topic_qos: public qos::TopicQos {
+    public:
+        default_builtin_topic_qos(): qos::TopicQos() {
+            const char *partition = "__BUILT-IN PARTITION__";
+            dds_qos_t *qos = dds_create_qos ();
+            dds_qset_durability (qos, DDS_DURABILITY_TRANSIENT_LOCAL);
+            dds_qset_presentation (qos, DDS_PRESENTATION_TOPIC, false, false);
+            dds_qset_reliability (qos, DDS_RELIABILITY_RELIABLE, DDS_MSECS(100));
+            dds_qset_partition (qos, 1, &partition);
+            delegate().ddsc_qos(qos);
+            dds_delete_qos(qos);
+        }
+};
+
+template <>
+Topic<ParticipantBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<ParticipantBuiltinTopicData> >(new detail::Topic<ParticipantBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSParticipant",
+              default_builtin_topic_qos(),
+              DDS_BUILTIN_TOPIC_DCPSPARTICIPANT))
+{
+    this->delegate()->init(this->impl_);
+}
+
+template <>
+Topic<TopicBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<TopicBuiltinTopicData> >(new detail::Topic<TopicBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSTopic",
+              default_builtin_topic_qos(),
+              DDS_BUILTIN_TOPIC_DCPSTOPIC))
+{
+    this->delegate()->init(this->impl_);
+}
+
+template <>
+Topic<SubscriptionBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<SubscriptionBuiltinTopicData> >(new detail::Topic<SubscriptionBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSSubscription",
+              default_builtin_topic_qos(),
+              DDS_BUILTIN_TOPIC_DCPSSUBSCRIPTION))
+{
+    this->delegate()->init(this->impl_);
+}
+
+template <>
+Topic<PublicationBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<PublicationBuiltinTopicData> >(new detail::Topic<PublicationBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSPublication",
+              default_builtin_topic_qos(),
+              DDS_BUILTIN_TOPIC_DCPSPUBLICATION))
+{
+    this->delegate()->init(this->impl_);
+}
+
+}
+
+}
--- a/src/ddscxx/src/org/eclipse/cyclonedds/core/EntityDelegate.cpp
+++ b/src/ddscxx/src/org/eclipse/cyclonedds/core/EntityDelegate.cpp
@@ -62,6 +62,12 @@ org::eclipse::cyclonedds::core::EntityDelegate::~EntityDelegate()
 
 void org::eclipse::cyclonedds::core::EntityDelegate::enable()
 {
+  if (DDS_BUILTIN_TOPIC_DCPSPARTICIPANT == this->ddsc_entity
+   || DDS_BUILTIN_TOPIC_DCPSTOPIC == this->ddsc_entity
+   || DDS_BUILTIN_TOPIC_DCPSSUBSCRIPTION == this->ddsc_entity
+   || DDS_BUILTIN_TOPIC_DCPSPUBLICATION == this->ddsc_entity)
+    return;
+
   dds_return_t ret;
   enabled_ = true;
   ret = dds_set_listener (this->ddsc_entity, this->listener_callbacks);
--- a/src/ddscxx/CMakeLists.txt
+++ b/src/ddscxx/CMakeLists.txt
@@ -20,6 +20,7 @@ set(sources
     src/dds/core/status/State.cpp
     src/dds/domain/discovery.cpp
     src/dds/domain/find.cpp
+    src/dds/topic/BuiltinTopic.cpp
     src/dds/pub/pubdiscovery.cpp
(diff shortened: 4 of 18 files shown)


t0ny-peng:
Hi @reicheratwork. I was trying to use the builtin datareader you added in this MR. Now my waitset can get message when a new Participant/Subscription/Publication is created, and I can `take` the message from the builtin DataReader with no problem.

However, when the other Publisher goes offline, my `builtin_pub_datareader->take()` in the first node has a segment fault, and the problem happened on [this line](https://github.com/reicheratwork/cyclonedds-cxx/blob/be5bb2066641a2f341e3df56bbd538175722fb02/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp#L1519).

Here `c_data->topic_name` and `c_data->type_name` (both converted from `c_sample`) were `NULL` and that's where the segfault happens. 

image

I guess a proper fix is to check if the value is NULL before calling the `topic_name(const char*)` method in the underlying CycloneDDS, but could you confirm if it is normal to receive a "destruction" message on the builtin topic `DCPSPublication`? Thanks!

t0ny-peng:
@reicheratwork Another problem I'm facing is that, all the messages of type `dds::topic::PublicationBuiltinTopicData` that I got from builtin topic `DCPSPublication` have the same guid(`sample->data()->key()->value()`). Therefore, it is impossible to tell them apart. I thought each message should represent a unique datawriter with a unique guid. Could you confirm that? THanks.

```
Receieved Pub samples, length: 10
Get datawriter on topic: ros_discovery_info
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rt/rosout
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rr/talker/get_parametersReply
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rr/talker/get_parameter_typesReply
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rr/talker/set_parametersReply
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rr/talker/set_parameters_atomicallyReply
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rr/talker/describe_parametersReply
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rr/talker/list_parametersReply
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rt/parameter_events
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
Get datawriter on topic: rt/chatter
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
bb 0c 10 01 17 42 af b7 74 ff 23 b5 bb 0c 10 01 17 42 af b7 74 ff 23 b5
```

reicheratwork:
@t0ny-peng Thanks for being my informal beta tester!

You are correct with regarding the segfault issue, this is an oversight on my part by assumming that all builtin topic samples are valid, when they may not be. I will have a fix and some unittests for this ready soon.

I am still looking in your second observation.

t0ny-peng:
@reicheratwork Cool. Sumanth said we had an NDA signed with ADLink so maybe a live talk would be quick for debugging? Let me know if you have time this week. Thanks

reicheratwork:
@t0ny-peng I just pushed an update fixing some of the issues that you reported, please poke around and find things that i might have missed.

This branch is not yet up to date with the current master of CycloneDDS-C, I compile it against the releases/0.8.x branch.
You will also need to cherry pick the commit in https://github.com/eclipse-cyclonedds/cyclonedds/pull/969 and apply it to your  clone of CycloneDDS-C to get the fix for the byte swapping.

t0ny-peng:
Hi @reicheratwork! Btw we finally finished the update of graph implementation internally. The change in your PR works without any problem. Do we have any plan merging it to the CycloneDDS repository?

Thanks

reicheratwork:
> Hi @reicheratwork! Btw we finally finished the update of graph implementation internally. The change in your PR works without any problem. Do we have any plan merging it to the CycloneDDS repository?
> 
> Thanks

Hello @t0ny-peng ,

The plan is to merge it to the CycloneDDS repository, but since the review process is stalled at this moment, I cannot progress to move towards making this a true PR as that would look a lot like marking my own homework.
Maybe you can have a look over the current state of the changes? That would certainly accelerate the process.

P.S.:
I also rebased the commits onto the latest version of CycloneDDS-cxx::master (d9234aa) to bring it in line with the latest changes to CycloneDDS::master (66926cf), maybe check how these impact your work?

eboasson:
@e-hndrks will you have some time in the really near future to look into this further? I can see why you feel there must be a better way — I certainly don't fancy the combination of macros and templates, but I don't have an immediate suggestion how to do it with templates alone (there is no doubt in my mind that can be done).

Some progress on this would be nice, though, so absent any near-term practical proposals, I am inclined towards merging it.

src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicTraits.hpp
```diff
@@ -16,6 +16,142 @@
...
+  }
+
+  static size_t getSampleSize()
+  {
+    return sizeof(dds::topic::ParticipantBuiltinTopicData);
+  }
+
+  static bool isSelfContained()
+  {
+    return false;
```
e-hndrks, line 55: Why is the ParticipantBuiltinTopicData type not considered selfContained? Is that because it has an unbounded sequence which forces it to have an indirection to a separately allocated part? How exactly do you define the term SelfContained in this context? Is this only relevant in the context of the IceOryx integration, where only SelfContained types may be communicated through shared memory?
(The same questions also applies to the other Builtin topic types, but I am not going to repeat it for every one.)

src/ddscxx/src/dds/topic/BuiltinTopic.cpp
```diff
@@ -0,0 +1,94 @@
...
+namespace dds
+{
+
+namespace topic
+{
+
+class default_builtin_topic_qos: public qos::TopicQos {
+    public:
+        default_builtin_topic_qos(): qos::TopicQos() {
+            const char *partition = "__BUILT-IN PARTITION__";
```
e-hndrks, line 29: What is the partition qosPolicy doing in a TopicQos? Topics don't attach to a partition, Publishers and Subscribers do.

src/ddscxx/src/dds/topic/BuiltinTopic.cpp
```diff
@@ -0,0 +1,94 @@
...
+        }
+};
+
+template <>
+Topic<ParticipantBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<ParticipantBuiltinTopicData> >(new detail::Topic<ParticipantBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSParticipant",
```
e-hndrks, line 46: From a code maintenance perspective, Isn't it better to invoke the static getTypeName() from the TopicTraits\<ParticipantBuiltinTopicData\> here instead of duplicating its type-name in this piece of code?

src/ddscxx/src/dds/topic/BuiltinTopic.cpp
```diff
@@ -0,0 +1,94 @@
...
+    this->delegate()->init(this->impl_);
+}
+
+template <>
+Topic<TopicBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<TopicBuiltinTopicData> >(new detail::Topic<TopicBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSTopic",
```
e-hndrks, line 59: From a code maintenance perspective, Isn't it better to invoke the static getTypeName() from the TopicTraits\<TopicBuiltinTopicData\> here instead of duplicating its type-name in this piece of code?

src/ddscxx/src/dds/topic/BuiltinTopic.cpp
```diff
@@ -0,0 +1,94 @@
...
+    this->delegate()->init(this->impl_);
+}
+
+template <>
+Topic<SubscriptionBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<SubscriptionBuiltinTopicData> >(new detail::Topic<SubscriptionBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSSubscription",
```
e-hndrks, line 72: From a code maintenance perspective, Isn't it better to invoke the static getTypeName() from the TopicTraits\<SubscriptionBuiltinTopicData\> here instead of duplicating its type-name in this piece of code?

src/ddscxx/src/dds/topic/BuiltinTopic.cpp
```diff
@@ -0,0 +1,94 @@
...
+    this->delegate()->init(this->impl_);
+}
+
+template <>
+Topic<PublicationBuiltinTopicData, detail::Topic>::Topic(const dds::domain::DomainParticipant& dp,
+                          const std::string& topic_name) :
+::dds::core::Reference< detail::Topic<PublicationBuiltinTopicData> >(new detail::Topic<PublicationBuiltinTopicData>(
+              dp,
+              topic_name,
+              "org::eclipse::cyclonedds::builtin::DCPSPublication",
```
e-hndrks, line 85: From a code maintenance perspective, Isn't it better to invoke the static getTypeName() from the TopicTraits\<PublicationBuiltinTopicData\> here instead of duplicating its type-name in this piece of code?

src/ddscxx/src/org/eclipse/cyclonedds/core/EntityDelegate.cpp
```diff
@@ -62,6 +62,12 @@ org::eclipse::cyclonedds::core::EntityDelegate::~EntityDelegate()
 
 void org::eclipse::cyclonedds::core::EntityDelegate::enable()
 {
+  if (DDS_BUILTIN_TOPIC_DCPSPARTICIPANT == this->ddsc_entity
+   || DDS_BUILTIN_TOPIC_DCPSTOPIC == this->ddsc_entity
+   || DDS_BUILTIN_TOPIC_DCPSSUBSCRIPTION == this->ddsc_entity
+   || DDS_BUILTIN_TOPIC_DCPSPUBLICATION == this->ddsc_entity)
+    return;
```
e-hndrks, line 69: Why do I prematurely return if the Entity I want to enable belongs to a builtin topic? I guess the builtin topics are already enabled by default, but shouldn't the enable() function be idempotent according to the spec? Why the need for an intercept here?

e-hndrks (changes requested):
Generally, I see that the approach in this PullRequest makes the BuiltinTopics accessible in the C++ API, and that people could conveniently work with this. I have some small concrete remarks on the existing code within this existing approach, but have a more fundamental problem with the approach itself. 

That is fundamentally caused by the fact that the read/take operations of the C-API handle builtin topics very differently from user defined topics, because it doesn't use the registered ser_data functions in the same way.  This means that in the C++ API you have to work around this by having lots of additional boilerplate code that transforms a C sample into a C++ sample and to avoid duplication you heavily rely on using macro's to limit the amount of boilterplate needed. 
My problem is that this makes the code not very readable, and also hard to debug. The fundamental solution in my opinion is that we should provide a more dedicated read/take call on the C API that allows us to treat builtin topics much in the same way as we do normal topics, and that can also handle Selectors in a more convenient way. This would not only benefit the C++ API for now, but also any other language binding that we may add in the future. 
I spoke to @eboasson this afternoon, and he generally agrees with this concern and with the need for better dedicated read/take calls in the C API. Together with him I will try to write a separate issue for this on the Cyclone core, which covers our concerns and the proposed APIs, and we both think this new issue should be treated with urgency.
That means that all in all I would not suggest merging this PullRequest in right now, although we might want to process the comments in this review for the people currently depending on this PullRequest. Our suggestion is to wait for the new read/take API on C, and then built a more generic BuiltinTopic implementation on top of those, that behaves much more like the ordinary read/take calls (which will then also forward to the new C calls) thus removing the need for all the additional boilerplate code.

reicheratwork:
@t0ny-peng due to the @e-hndrks saying that the ultimate solution that cyclonedds-cxx will be going for to support builtin topics is to make the cyclonedds usage of builtin topics more in line with the "ordinary" topics, I will not be merging this PR to cyclonedds-cxx repo.
As you said that it was working for you in a satisfactory manner, I propose that you keep using this modification until the more fundamental fix (using the new cyclonedds builtin topics) is in.

joakimono:
I'm trying to use the work in this pull request, but I'm having issues when I try to run the bundled helloworld example. Running both publisher and subscriber, they 'hang' forever, with the condition `writer.publication_matched_status().current_count() == 0` in publisher never being met. After looking around in the commits I've managed to get the helloworld working if I replace `sertype_typeid<T>` with `nullptr` of `ddscxx_sertype_ops` in include/org/eclipse/cyclonedds/topic/datatopic.hpp. Then, of course, builtin topics will not work. 
I built against cyclonedds tagged version 0.9.0a1 with options topic_discovery and type_discovery enabled.

I must admit I don't understand what is going on, and looking at the log output with both _fine_ and _finest_ does not make me any wiser as to what is going on. I've attached publisher and subscriber log with _fine_ verbosity for reference.

Can someone provide some hints to where I should look, @reicheratwork  perhaps?

[pub.log](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/7912667/pub.log)
[sub.log](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/7912671/sub.log)

reicheratwork:
> I'm trying to use the work in this pull request, but I'm having issues when I try to run the bundled helloworld example. Running both publisher and subscriber, they 'hang' forever, with the condition `writer.publication_matched_status().current_count() == 0` in publisher always being met. After looking around in the commits I've managed to get the helloworld working if I replace `sertype_typeid<T>` with `nullptr` of `ddscxx_sertype_ops` in include/org/eclipse/cyclonedds/topic/datatopic.hpp. Then, of course, builtin topics will not work. I built against cyclonedds tagged version 0.9.0a1 with options topic_discovery and type_discovery enabled.
> 
> I must admit I don't understand what is going on, and looking at the log output with both _fine_ and _finest_ does not make me any wiser as to what is going on. I've attached publisher and subscriber log with _fine_ verbosity for reference.
> 
> Can someone provide some hints to where I should look, @reicheratwork perhaps?
> 
> [pub.log](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/7912667/pub.log) [sub.log](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/7912671/sub.log)

First:
This PR is closed, and no further work will be done on it.

Second:
I Can confirm this issue, compiling cyclone WITHOUT type/topic discovery allows the C++ examples to work as expected, but WITH it, they hang and do not complete.
The missing function pointer is a callback function used to get the typeid for the topic. Which is called during the type/topic discovery step.

I do not know whether this function is even necessary for the builtin topics to work, as these topics do not use the serdata/sertype callback functions / as they are built-in and do not originate from usual paths.

eboasson:
The only thing bit in the logs that trigger me is that there are publisher and the subscriber use different type ids. I haven't had the time yet to look into the possible cause of that.

> This PR is closed, and no further work will be done on it.

This is true, and a new one that depends on improvements in the C API will be forthcoming, but for the time being this is best way to get the built-in topics in C++. So I think we should be nice to those willing to bear with us as we redo it and use it as a workaround in the meantime.

>  Can confirm this issue, compiling cyclone WITHOUT type/topic discovery allows the C++ examples to work as expected, but WITH it, they hang and do not complete.
> The missing function pointer is a callback function used to get the typeid for the topic. Which is called during the type/topic discovery step.

I know those bits not being there yet can cause a problem because of the issue addressed by https://github.com/eclipse-cyclonedds/cyclonedds/pull/1116 — but that one came about playing around with the HEAD of `cyclonedds` and with #178 applied ...

UranusTsing:
I'm trying to use the work in this pull request, I'm having issues when I try to build.  cyclonedds-cxx Interface : ddscxx_sertype had update, how do I need to do?

t0ny-peng:
@UranusTsing What's the version of CycloneDDS you are using, and what's the build command. It helps reproduce the error.

UranusTsing:
> @UranusTsing您使用的 CycloneDDS 版本是什么，构建命令是什么。它有助于重现错误。
我是用的 CycloneDDS master最新的代码，构建命令是：cmake -Bbuild -DCMAKE_PREFIX_PATH=<c_install_path> c_install_path是CycloneDDS 编译后install的路径

UranusTsing:
I want to use the CycloneDDS_C++ builtin topics  based on CycloneDDS master, but build error， src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicTraits.hpp：
  static ddsi_sertype *getSerType()
  {
    auto *st = new ddscxx_sertype<dds::topic::ParticipantBuiltinTopicData>();
    return static_cast<ddsi_sertype*>(st);
  }
ddscxx_sertype() param is 1, but should be 2

reicheratwork:
> I want to use the CycloneDDS_C++ builtin topics based on CycloneDDS master, but build error， src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicTraits.hpp： static ddsi_sertype *getSerType() { auto _st = new ddscxx_sertypedds::topic::ParticipantBuiltinTopicData(); return static_cast<ddsi_sertype_>(st); } ddscxx_sertype() param is 1, but should be 2

@UranusTsing 
It could very well be that there are some inconsistencies between this branch's expected interfaces and the current interfaces on CycloneDDS-C, as the latter has had some significant additions in the last months.

Could you give us some more exact feedback on:
- the current version of CycloneDDS you are compiling against (git checkout hash would be best)
- which commands you are using to compile CycloneDDS-CXX
- what the feedback is you are getting from the compiler (especially where it goes wrong)

Thanks.

PR #136, opened 2021-07-16
Serialization improvements

reicheratwork:
Improved CDR serialization performance by:
- only checking whether byte swaps need to be done at the start of serialization of the entire message, and then invoking the correct streaming function, in stead of checking for byte swapping at each primitive
- copying contiguous blocks of primitives (sequences/array) is now no longer done by iterating over each entry

Added unittests for serialization and byteswapping.

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
@@ -190,6 +190,8 @@ inline void write_many(basic_cdr_stream& str, const T* in, size_t N)
   if (str.abort_status())
     return;
 
+  str.align(sizeof(T), true);
+
   T *out = static_cast<T*>(str.get_cursor());
 
   memcpy(out, in, sizeof(T)*N);
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
@@ -41,10 +41,9 @@ public:
    *
    * Basically a pass through for the cdr_stream base class.
    *
-   * @param[in] end The endianness to set for the data stream, default to the local system endianness.
    * @param[in] ignore_faults Bitmask for ignoring faults, can be composed of bit fields from the serialization_status enumerator.
    */
-  basic_cdr_stream(endianness end = native_endianness(), uint64_t ignore_faults = 0x0) : cdr_stream(end, 8, ignore_faults) { ; }
+  basic_cdr_stream(uint64_t ignore_faults = 0x0) : cdr_stream(8, ignore_faults) { ; }
 };
 
 /**
@@ -62,7 +61,7 @@ public:
  *
  * Aligns the stream to the alignment of type T.
  * Reads the value from the current position of the stream str into
- * toread, will swap bytes if necessary.
+ * toread.
  * Moves the cursor of the stream by the size of T.
  * This function is only enabled for arithmetic types and enums.
  *
@@ -70,17 +69,65 @@ public:
  * @param[out] toread The variable to read into.
  */
 template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-void read(basic_cdr_stream &str, T& toread)
+inline void read(basic_cdr_stream &str, T& toread)
 {
   if (str.abort_status())
     return;
 
   str.align(sizeof(T), false);
 
-  char *cursor = str.get_cursor();
-  assert(cursor);
-  transfer_and_swap(*(reinterpret_cast<const T*>(cursor)), toread, str.swap_endianness());
+  toread = *static_cast<T*>(str.get_cursor());
+
   str.incr_position(sizeof(T));
+}
+
+/**
+ * @brief
+ * Primitive type read function, byteswap version.
+ *
+ * Same as read(), with an additional byteswap at the end.
+ *
+ * @param[in, out] str The stream which is read from.
+ * @param[out] toread The variable to read into.
+ */
+template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
+inline void read_swapped(basic_cdr_stream &str, T& toread)
+{
+  read(str, toread);
+
+  byte_swap(toread);
+}
+
+template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
+inline void read_many(basic_cdr_stream &str, T *out, size_t N)
+{
+  if (str.abort_status())
+    return;
+
+  str.align(sizeof(T), false);
+
+  T *in = static_cast<T*>(str.get_cursor());
+
+  memcpy(out, in, sizeof(T)*N);
+
+  str.incr_position(sizeof(T)*N);
+}
+
+template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
+inline void read_many_swapped(basic_cdr_stream &str, T *out, size_t N)
+{
+  if (str.abort_status())
+    return;
+
+  str.align(sizeof(T), false);
+
+  T *in = static_cast<T*>(str.get_cursor());
+  for (size_t i = 0; i < N; i++, out++, in++) {
+    *out = *in;
+    byte_swap(*out);
+  }
+
+  str.incr_position(sizeof(T)*N);
 }
 
 /**
@@ -97,17 +144,74 @@ void read(basic_cdr_stream &str, T& toread)
  * @param[in] towrite The variable to write.
  */
 template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-void write(basic_cdr_stream& str, const T& towrite)
+inline void write(basic_cdr_stream& str, const T& towrite)
 {
   if (str.abort_status())
     return;
 
   str.align(sizeof(T), true);
 
-  char *cursor = str.get_cursor();
-  assert(cursor);
-  transfer_and_swap(towrite, *(reinterpret_cast<T*>(cursor)), str.swap_endianness());
+  auto out = static_cast<T*>(str.get_cursor());
+
+  *out = towrite;
+
   str.incr_position(sizeof(T));
+}
+
+/**
+ * @brief
+ * Primitive type write function, byteswap version.
+ *
+ * Same as write(), with an additional byteswap at the end.
+ *
+ * @param[in, out] str The stream which is read from.
+ * @param[in] towrite The variable to write.
+ */
+template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
+inline void write_swapped(basic_cdr_stream& str, const T& towrite)
+{
+  if (str.abort_status())
+    return;
+
+  str.align(sizeof(T), true);
+
+  auto out = static_cast<T*>(str.get_cursor());
+
+  *out = towrite;
+
+  byte_swap(*out);
+
+  str.incr_position(sizeof(T));
+}
+
+template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
+inline void write_many(basic_cdr_stream& str, const T* in, size_t N)
+{
+  if (str.abort_status())
+    return;
+
+  T *out = static_cast<T*>(str.get_cursor());
+
+  memcpy(out, in, sizeof(T)*N);
+
+  str.incr_position(sizeof(T)*N);
+}
+
+template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
+inline void write_many_swapped(basic_cdr_stream& str, const T* in, size_t N)
+{
+  if (str.abort_status())
+    return;
+
+  str.align(sizeof(T), true);
+
+  T *out = static_cast<T*>(str.get_cursor());
+  for (size_t i = 0; i < N; i++, out++, in++) {
+    *out = *in;
+    byte_swap(*out);
+  }
+
+  str.incr_position(sizeof(T)*N);
 }
 
 /**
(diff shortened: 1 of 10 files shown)


reicheratwork:
The failing macOS test may be flaky in some way, as it did not fail in my build of the exactly same checkout:
https://dev.azure.com/martijnreicher/cyclonedds-cxx/_build/results?buildId=48&view=results

sumanth-nirmal:
The changes in the MR gives a significant performance improvement when sending/receiving large data.

For details, I used `rosbag2_performance_benchmarking` with a custom config for data producer with data from multiple sensors like lidar, camera, radar, GPS, imu, etc. Before this change, I only used to receive ~35% of the messages and with the changes in this MR, I can get ~99.7% of the messages.

src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
```diff
@@ -312,6 +613,42 @@ void write_string(basic_cdr_stream& str, const T& towrite, size_t N)
...
+
+  size_t string_length = towrite.length();
+  if (N &&
+      string_length > N) {
+    if (str.status(serialization_status::write_bound_exceeded))
+      return;
+  }
+
+  //add 1 extra for terminating NULL
+  string_length++;
```
k0ekk0ek, line 640: This invalidates the previous boundary check. It's unlikely it'll happen, but still.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -37,6 +42,10 @@ bool to_key(streamer& str, const T& tokey, ddsi_keyhash_t& hash)
   std::vector<unsigned char> buffer(sz + padding);
   memset(buffer.data() + sz, 0x0, padding);
   str.set_buffer(buffer.data());
+  /* TODO: what is key endianness to be used here?
+   * since, this may be different between nodes, and if this value is used
+   * for global lookups or the like, this
+   * may cause discrepancies. */
```
k0ekk0ek, line 48: I'd say, if streams can be both big and little endian, we should settle on what prefer to write out and use that. Seems like something @eboasson might want to weigh in on and then document it somewhere.

src/ddscxx/tests/DataReaderSelector.cpp
```diff
@@ -56,8 +57,11 @@ class DataReaderSelector : public ::testing::Test
         dds::topic::qos::TopicQos topic_qos = this->participant.default_topic_qos();
         topic_qos << dds::core::policy::Reliability::Reliable() <<
                      dds::core::policy::History::KeepAll();
+
+        char topicname[128];
+        create_unique_topic_name("selector_test_topic",topicname,sizeof(topicname));
```
k0ekk0ek, line 62: This is not a "real" error, but it diverts from the default of 32 that I used before (small nitpick) and it's inconsistent with the 64 you've chosen somewhere else in the PR. It's ok for now, but we should do a follow-up PR that simply change all occurrences to use the same.

k0ekk0ek (changes requested):
Thanks @reicheratwork! Totally agree with the chosen optimization, but I do have some minor comments.

eboasson:
There's also a real issue:

Type `a00022` generated by my [cxx-cdtest.pl](https://gist.githubusercontent.com/eboasson/62632a6fcfdde7839b3230856d205a38/raw/9bdc80d9fa6aaf60b7248c97aba02ec9920cd116/cxx-cdrtest.pl) run with seed 3179845304:
```
typedef octet a00028[1];
typedef sequence<a00028> a00027;
typedef sequence<unsigned short> a00030;
struct a00022 {
  float a00023;
  float a00024;
  a00027 a00025[3];
  a00030 a00029;
};
#pragma keylist a00022
```
causes a deserialization failure with this updated serializer. Publishing from C and reading back in C++ does work (by reversing the condition in the loop that publishes data), so I think it is only the serializer that is broken. Without this PR, it works fine.

Secondly, same out, type `a00264`:
```
typedef sequence<unsigned long long> a00270;
struct a00266 {
  double a00267[4];
  a00270 a00269;
  unsigned short a00271;
  string a00272[2];
};
typedef double a00277[3];
typedef sequence<a00277> a00276;
typedef sequence<a00276> a00275;
struct a00264 {
  a00266 a00265;
  a00275 a00274;
};
#pragma keylist a00264
```
Also freaks out, getting both serialization and deserialization wrong. That turns out to be aligning for reading/writing a sequence of 64-bit numbers when the sequence is empty.

Patches for the above:
```diff
commit b19ce372ee3698e13d2fec377c2cad6545eb75a5
Author: Erik Boasson <eb@ilities.com>
Date:   Thu Jul 22 10:45:21 2021 +0200

    Skip align in read/write_many for empty sequence
    
    Signed-off-by: Erik Boasson <eb@ilities.com>

diff --git a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
index 7eb637b..ea8f56d 100644
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
@@ -101,7 +101,7 @@ inline void read_swapped(basic_cdr_stream &str, T& toread)
 template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
 inline void read_many(basic_cdr_stream &str, T *out, size_t N)
 {
-  if (str.abort_status())
+  if (str.abort_status() || N == 0)
     return;
 
   str.align(sizeof(T), false);
@@ -116,7 +116,7 @@ inline void read_many(basic_cdr_stream &str, T *out, size_t N)
 template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
 inline void read_many_swapped(basic_cdr_stream &str, T *out, size_t N)
 {
-  if (str.abort_status())
+  if (str.abort_status() || N == 0)
     return;
 
   str.align(sizeof(T), false);
@@ -187,7 +187,7 @@ inline void write_swapped(basic_cdr_stream& str, const T& towrite)
 template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
 inline void write_many(basic_cdr_stream& str, const T* in, size_t N)
 {
-  if (str.abort_status())
+  if (str.abort_status() || N == 0)
     return;
 
   str.align(sizeof(T), true);
@@ -202,7 +202,7 @@ inline void write_many(basic_cdr_stream& str, const T* in, size_t N)
 template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
 inline void write_many_swapped(basic_cdr_stream& str, const T* in, size_t N)
 {
-  if (str.abort_status())
+  if (str.abort_status() || N == 0)
     return;
 
   str.align(sizeof(T), true);

commit 628052226dd7f3a9dc802aa37d0e494cad7529f3
Author: Erik Boasson <eb@ilities.com>
Date:   Thu Jul 22 10:44:53 2021 +0200

    Align input stream in read_many
    
    Signed-off-by: Erik Boasson <eb@ilities.com>

diff --git a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
index d7eb09d..7eb637b 100644
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
@@ -190,6 +190,8 @@ inline void write_many(basic_cdr_stream& str, const T* in, size_t N)
   if (str.abort_status())
     return;
 
+  str.align(sizeof(T), true);
+
   T *out = static_cast<T*>(str.get_cursor());
 
   memcpy(out, in, sizeof(T)*N);
```

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
eboasson: Always write out native endianness (most of the time the endianness is the same throughout the network) so that's definitely the fastest.

(There are exceptions: a DDSI keyhash is based on big-endian, but fortunately those are hardly ever needed. Furthermore DDS Security requires some discovery data to be little-endian, but that's not relevant to a language binding. The OMG is certainly consistent in being inconsistent ...)

src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
```diff
@@ -62,27 +61,75 @@ class basic_cdr_stream : public cdr_stream {
...
+ *
+ * @param[in, out] str The stream which is read from.
+ * @param[out] toread The variable to read into.
+ */
+template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
+inline void read_swapped(basic_cdr_stream &str, T& toread)
+{
+  read(str, toread);
+
+  byte_swap(toread);
```
k0ekk0ek, line 98: This is also incorrect. The read checks for `str.abort_status()` and returns, the `byte_swap()` does not.

eboasson: Good point (reviewing clearly catches some things that testing doesn't 😁).

I think it makes sense to distinguish between behaviour with valid inputs and that with invalid inputs: the code that's currently in master doesn't deal gracefully with invalid inputs, and against that background, I find it acceptable to take a PR that dramatically improves the performance for valid inputs without improving the situation with invalid inputs.

src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
```diff
@@ -182,10 +356,32 @@ void max(basic_cdr_stream& str, const T& max_sz)
  * @param[out] toread The variable to read into.
  */
 template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-void read(basic_cdr_stream& str, T& toread) {
+inline void read(basic_cdr_stream& str, T& toread) {
   read(str, *reinterpret_cast<uint32_t*>(&toread));
 }
```
k0ekk0ek, line 361: I think we'll have to look into this one too eventually as the size of enums can be changed in IDL4? We can probably get away with doing a `sizeof` of `T` and call the proper read function for the primitive.

src/idlcxx/src/streamers.c
```diff
@@ -681,6 +830,11 @@ process_inherit_spec(
   char *type = NULL;
   const char *fmt = "  %2$s(streamer,dynamic_cast<%1$s&>(instance));\n";
   const char *constfmt = "  %2$s(streamer,dynamic_cast<const %1$s&>(instance));\n";
+  if (streams->swapped)
+  {
+    fmt = "  %2$s(streamer,dynamic_cast<%1$s&>(instance));\n";
+    constfmt = "  %2$s(streamer,dynamic_cast<const %1$s&>(instance));\n";
+  }
```
k0ekk0ek, line 837: These formats are incorrect. Second occurrence should be `%2$s_swapped(...`.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -329,7 +345,7 @@ ddsi_serdata *serdata_from_sample(
   d->resize(sz);
   ptr = static_cast<unsigned char*>(d->data());
   memset(ptr, 0x0, 4);
-  if (str.local_endianness() == org::eclipse::cyclonedds::core::cdr::endianness::little_endian)
+  if (native_endianness() == endianness::little_endian)
     *(ptr + 1) = 0x1;
 
   str.set_buffer(calc_offset(d->data(), 4));
```
sumanth-nirmal, line 351: Shouldn't we also call `write()` based on the endianness? (with in the scope of the current changes, assuming endianness is not part of the streamer functions)

```diff
  switch (kind)
  {
  case SDK_KEY:
    key_write(str, msg);
    break;
  case SDK_DATA:
-     write(str, msg);
+    if (native_endianness() == endianness::little_endian)
+      write(str, msg);
+    else
+      write_swapped(str, msg);
    break;
  case SDK_EMPTY:
    assert(0);
  }
```

sumanth-nirmal: Probably this is also true for `move()`, though I think `move()` and `move_swapped()` should be same

eboasson: There is a need for that (assuming I am not overlooking something): while there are some places in the spec where byte order in the CDR stream is mandated (like in computing a DDSI keyhash as well as in some places in DDS Security — and perhaps elsewhere) for normal application data, there is no required byte order. Always serializing to native byte order therefore generally makes the most sense.

eboasson (approved):
After deliberation and discussion with @k0ekk0ek and some further testing specifically for the byte-swapping case on the read side I have come to the conclusion that merging this PR as it stands is in the interest of all users because of the significant performance improvements on large sequences. I also haven't seen any regressions.

That doesn't mean there is no further work to be done specifically on the (de)serializer, with the key points being:
- While it makes sense to generate a "straight" and a swapped variant, it doesn't make sense to push the decision to use the one or the other all the way to the caller. E.g. where before it read `read(str,*t)` now it is `if(swap_necessary(...) read_swapped(str,*t); else read(str,*t);` — but the CDR stream can still track the endianness, and a generic `read` function can still be written that then calls a straight or a swapped variant as needed.
- The generated code for the swapped variants is incorrect for inherited structs as noted by @k0ekk0ek. I have since learnt that inheritance is not actually usefully implemented yet, despite the absence of any error messages when one tries to use it. So it is not actually a regression in practice. I’ve created tickets for this in the C and C++ repositories.
- I image that some more creative use of templates would allow the C++ compiler to instantiate the straight and the swapped versions, eliminating the need to generate them. (That would automatically eliminate the preceding issue.)
- Input validation … but this PR is not a regression on this point.

More generally, there are a variety of other open points related to the IDL compiler C++ backend  and the implementation of `sertype` (e.g., invoking MD5 for computing a keyhash when there are no keys) that came out of reviewing and some other unrelated activities. Some of the things noted during review might well disappear if the suggestion to use templating for the straight/swapped variants work out. Some other points will still need attention. That can all wait until next week (ping me if I seem to forget …).

Requiring all that to be done prior to merging this would, I believe, unnecessarily delay making available the non-trivial performance improvements here that we know are important to some users today.

PR #141, opened 2021-08-20
Add test for checking if the status conditions trigger correctly

sumanth-nirmal:
This MR adds a simple test to verify if the status conditions attached to waitset trigger as expected.

### DIFF
--- a/src/ddscxx/tests/WaitSet.cpp
+++ b/src/ddscxx/tests/WaitSet.cpp
@@ -872,3 +872,60 @@ TEST_F(WaitSet, DISABLED_attach_detach_multiple_during_wait)
 
     ddsrt_thread_join(threadId, NULL);
 }
+
+/**
+ * Check that the status conditions are triggered correctly
+ *
+ * The test checks for trigger of status condition in the following sequence
+ *     1. Status Condition is create and initialized with StatusMask::none()
+ *     2. Specific status mask is enabled on status condition (requested_incompatible_qos())
+ *     3. Status condition is attached to the waitset
+ *     4. The test checks for the trigger for an event which happend before enabling the status mask
+ */
+TEST_F(WaitSet, status_condition_trigger)
+{
+  using SampleType = Space::Type2;
+  auto topic = dds::topic::Topic<SampleType>(this->participant, "space_type2");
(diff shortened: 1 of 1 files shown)


eboasson:
Thanks @sumanth-nirmal, I had to find out what was going wrong so you get a response quicker than might otherwise be expected 🙂 the problem is not that the StatusMask or the triggering doesn't work as expected (https://github.com/eclipse-cyclonedds/cyclonedds/issues/891#issuecomment-902991431), but that the number of incompatible QoS events is different from the expectation.

That is because the `WaitSet` test class initialization already creates a volatile writer at https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/452bb2e6039e2ecd18e07a18b3a8a970b55e8f62/src/ddscxx/tests/WaitSet.cpp#L344. If you change the test to use a different topic (that was the easy way for me), e.g.,:
```C++
TEST_F(WaitSet, status_condition_trigger)
{
  using SampleType = Space::Type1;

  char name[100];
  create_unique_topic_name("WaitSet", name, sizeof(name));
  dds::topic::Topic<Space::Type1> topic = dds::topic::Topic<Space::Type1>(this->participant, name);
  ASSERT_NE(this->topic, dds::core::null);
```
that problem disappears and the test passes.

I don't particularly like fixing the issue in this manner. Another route would be to only create a data reader in the test while relying on the `WaitSet` test class constructor to create the writer, but that I don't like so much because it add unnecessary dependencies.

As the test really is about entity status and status conditions, moving it to `EntityStatus.cpp` is a very practical solution. Would you agree?

eboasson (commented):
I'm fine with the updated test and but it can't be merged yet because it includes #142. If you squash the two test commits implementing the test and remove the commits for the build system, I'll approve it.

sumanth-nirmal:
@eboasson Sorry for the delay, I fixed the commits now and it should be ready for review.

However, I now realized that the CI will still fail with SharedMemory tests because of the `iceoryx` version used in CI

sumanth-nirmal:
@eboasson I have revived the branch, and this adds a simple test to check if the conditions attached to a waitset are triggered as expected, please take a look and let me know if something else is missing

PR #142, opened 2021-08-20
Fix iceoryx dependencies for SHM tests

sumanth-nirmal:


eboasson:
I locally have the same problem building the tests with Iceoryx enabled as the two failing runs have here, I'm pretty sure that the `iceoryx_posh_roudi_environment` package configuration file is missing from a commit, because I can't find anything that remotely looks like it either in iceoryx or in cyclonedds-cxx.

sumanth-nirmal:
> I'm pretty sure that the iceoryx_posh_roudi_environment package configuration file is missing from a commit, because I can't find anything that remotely looks like it either in iceoryx or in cyclonedds-cxx

@eboasson That is correct it has changed in the master of `iceoryx` and the dependency should be on `iceoryx_testing`. However, to avoid these kinds of problems I guess it might be reasonable to pin `iceoryx` version in Cyclone DDS.

In this MR I have pinned `iceoryx` to V1.0.1, @dkroenke @eboasson @k0ekk0ek can you check if that makes sense.

.azure/templates/build-test.yml
```diff
@@ -88,13 +88,13 @@ steps:
       set -e -x
       sudo apt-get install libacl1-dev libncurses5-dev pkg-config
       git clone --single-branch \
-                --branch "${ICEORYX_BRANCH:-master}" \
+                --branch "${ICEORYX_BRANCH:-v1.0.1}" \
```
eboasson, line 91: Pinning it to the `release_1.0` branch seems like a better compromise (it is what I use locally anyway). In my view it would be even better would be if we can handle both the 1.0.x and current master, and it seems like this could be done with a simple version check.

.azure/templates/build-test.yml
```diff
@@ -88,13 +88,13 @@ steps:
...
                 "${ICEORYX_REPOSITORY:-https://github.com/eclipse-iceoryx/iceoryx.git}" \
                 iceoryx
       mkdir iceoryx/build
       cd iceoryx/build
       cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
-            -DCMAKE_BUILD_SHARED_LIBS=on \
+            -DBUILD_SHARED_LIBS=on \
             -DCMAKE_INSTALL_PREFIX=install \
             -DROUDI_ENVIRONMENT=on \
```
eboasson, line 99: We should also check whether Iceoryx was built with `-DROUDI_ENVIRONMENT=ON`: that's the bit that produces the file (and turns out to have been the reason for my not having the testing package locally).

.azure/templates/build-test.yml
sumanth-nirmal: @eboasson I think we can do a version check to determine if it is `1.0.x` or `master`, but if we choose the current master from `iceoryx` then probably a lot has been changed when compared with `1.0.x` and only fixing only this dependency might not be sufficient to make it work! 
Maybe we should sync the updating of `iceoryx` version here with `iceoryx` releases or we have a nightly job that tests with `iceoryx` master and then fix any test failures from this job.

sumanth-nirmal:
Closing this since the actual fix for pinning the iceoryx version is already done here https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/158/

CC @eboasson @k0ekk0ek

Issue #146, opened 2021-08-31
Serialization is performed in the write data path even only with shared memory readers


sumanth-nirmal:
## Required information

**Operating system:**
Ubuntu 18.04 LTS

**Cyclone Version**
- cyclonedds - master branch, https://github.com/eclipse-cyclonedds/cyclonedds/commit/fb0c57ee07cd075f04dad6a7d6add92f170930b6
- cyclonedds-cxx - master , https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/212146f849f93e5dd5980ce60999e58c8ed17a0e

**Observed behaviour:**

When shared memory is enabled and if only shared memory readers are available on the system (no network readers) then serialization is still performed. Ideally, serialization is not required when only shared memory readers are available. 

This is linked to issue https://github.com/eclipse-cyclonedds/cyclonedds/issues/783 

**Expected result or behaviour:**

Performing this unwanted serialization has performance impacts when using shared memory. The expected behavior should be to not have any copies or serialization in the data path with shared memory enabled.

PR #147, opened 2021-08-31
Avoid serialization in data path with shared memory enabled

sumanth-nirmal:
This PR adds the implementation for a `serdata` function which avoids serialization when using shared_memory and when there are no network readers. This will fix the issue https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/146

This PR also has a dependency on https://github.com/eclipse-cyclonedds/cyclonedds/pull/788, and can be merged only when the aforementioned PR in cyclonedds is merged.


sumanth-nirmal:
@eboasson I have revived this branch and modified the `serdata_from_iox_buffer` to avoid serialization in write with SHM. This is in line with the changes here https://github.com/eclipse-cyclonedds/cyclonedds/pull/788. Can you please review the changes?

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -592,9 +592,26 @@ ddsi_serdata * serdata_from_iox_buffer(
 {
   try {
     auto d = new ddscxx_serdata<T>(typecmn, kind);
-    d->iox_chunk = iox_buffer;
-    d->iox_subscriber = sub;
 
+    if (sub == nullptr) {  // serdata from the loaned sample
+#ifndef _WIN32
```
eboasson, line 597: `dds/ddsrt/misc.h` has macros for doing this, `DDSRT_WARNING_GNUC_OFF(old-style-cast)` should do this trick (there's a corresponding `ON` macro as well, of course)

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -592,9 +592,26 @@ ddsi_serdata * serdata_from_iox_buffer(
...
+#ifndef __clang__
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wold-style-cast"
+#endif
+#endif
+      d->iox_chunk = SHIFT_BACK_TO_ICEORYX_HEADER(iox_buffer);
```
eboasson, line 603: This has just been changed, after I merged https://github.com/eclipse-cyclonedds/cyclonedds/pull/943 and https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/159 this morning ...

Fortunately it is a trivial change. I also suspect that updating it to use the new Iceoryx interface will eliminate the need to disable the GCC warning.

eboasson (commented):
Thanks @sumanth-nirmal, I think it looks fine, except that events overtook it one (fairly trivial) point. I suspect a one-line change and deleting a few then-superfluous ones will bring take care of that.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
sumanth-nirmal: Thanks for point it :) I was not aware of this

PR #164, opened 2021-09-30
Verify that iceoryx is really used

elfenpiff:
Please merge after #159 is merged.

A failing CI is caused by the not yet merged iceoryx header PR into cyclonedds:master


src/ddscxx/tests/SharedMemory.cpp
```diff
@@ -60,7 +65,42 @@ void make_sample_(Bounded::Msg & sample, const int32_t cnt)
...
   std::fill(sample.bounded_sequence().begin(), sample.bounded_sequence().begin() + 255, cnt);
 }
+
+template<typename T>
+struct MessageProperty;
+
+template<>
+struct MessageProperty<Space::Type1> 
+{
+  static constexpr char INSTANCE_NAME[] = "Space::Type1";
```
sumanth-nirmal, line 75: We can get this info from the type using such a trait `org::eclipse::cyclonedds::topic::TopicTraits<DDSType>::getTypeName()`

src/ddscxx/tests/SharedMemory.cpp
```diff
@@ -60,7 +65,42 @@ void make_sample_(Bounded::Msg & sample, const int32_t cnt)
...
+  static constexpr bool IS_SHARED_MEMORY_COMPATIBLE = true;
```
sumanth-nirmal, line 76: Similarly we can get this info from the type using such a trait `org::eclipse::cyclonedds::topic::TopicTraits<DDSType>::isSelfContained()`

If we use these traits, we then probably will not need this `MessageProperty` struct

sumanth-nirmal (approved):
Thanks, @elfenpiff, looks good

PR #176, opened 2021-11-12
Fix write_cdr/take_cdr with SHM

### LINKED ISSUE #175: Communication fails with write_cdr when SHM is enabled and dynamic types are used
## Required information

**Operating system:**
Ubuntu 18.04 LTS

- cyclonedds - master branch, https://github.com/eclipse-cyclonedds/cyclonedds/commit/8c44fdd1905d24dabe9f54f9c191a0465085e846
- cyclonedds-cxx - master branch, https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/d9234aaf1c00492f37cb06db412f5f68040f8e44

 The communication fails when using `write_cdr` and `take_cdr` with dynamic types and SHM with publisher and subscriber in different processes

What do you expect to happen?

The communication should happen between publishers and subscribers in different processes with Dynamic types and with SHM when using write_cdr/take_cdr

**Conditions where it occurred / Performed steps:**

Writing data with `write_cdr` and receiving this with `take_cdr` in different processes doesn't work

**Details**

The reason for this could be that when SHM is enabled on the process with the writer and the process with the reader, the locators correctly compute that the communication is handled by the SHM. However, with `write_cdr` when constructing the `serdata` SHM is not used and the `iox_chunk` pointer is not updated. Since the `iox_chunk` is not updated the delivery doesn't consider the path with SHM, and the data is also not delivered using the regular network path because the locators detect this as an SHM reader, which eventually leads to no communication/data.

The fix is to handle the SHM data path with `write_cdr` and `read_cdr` when SHM is enabled.

sumanth-nirmal:
This MR fixes the data path for `write_cdr`/`take_cdr` with SHM

### Details

1. With SHM enabled in `write_cdr`, when constructing the `serdata`, memory is loaned from `iceoryx` and copies the user-provided serialized data into this loaned memory and updates the `serdata` with the loaned chunk
2. While receiving the sample in the `CDRSamplesHolder` if the data is over iceoryx the data is taken from the iceoryx chunk, copied into the `CDRBlob` for the user, and the loaned chunk is released

Closes #175 

This should be merged after https://github.com/eclipse-cyclonedds/cyclonedds/pull/1031

### DIFF
--- a/src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
+++ b/src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
@@ -166,6 +166,7 @@ public:
                 "The received sample over SHM is not in serialized form");
             }
             // release the chunk
+            free_iox_chunk(static_cast<iox_sub_t *>(current_blob->iox_subscriber), &current_blob->iox_chunk);
           }
         } else
 #endif
--- a/src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
+++ b/src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
@@ -143,37 +143,9 @@ public:
         org::eclipse::cyclonedds::topic::CDRBlob &sample_data = (*this->samples_.delegate())[i].delegate().data();
         // update the data kind
         sample_data.kind(static_cast<org::eclipse::cyclonedds::topic::BlobKind>(current_blob->kind));
-#ifdef DDS_HAS_SHM
-        // if the data is transferred using shm
-        if (current_blob->iox_chunk && current_blob->iox_subscriber) {
-            // get the size
-            auto iox_header = iceoryx_header_from_chunk(current_blob->iox_chunk);
-            // if the iox chunk has the data in serialized form
-            if (iox_header->shm_data_state == IOX_CHUNK_CONTAINS_SERIALIZED_DATA) {
-              copy_buffer_to_cdr_blob(reinterpret_cast<uint8_t *>(current_blob->iox_chunk),
-                                      iox_header->data_size, sample_data.kind(), sample_data);
-            } else if (iox_header->shm_data_state == IOX_CHUNK_CONTAINS_RAW_DATA) {
-              // serialize the data
-              auto serialized_size = ddsi_sertype_get_serialized_size(current_blob->type,
-                                                                      current_blob->iox_chunk);
-              // create a buffer to serialize
-              std::vector<unsigned char> buffer(serialized_size);
-              // serialize into the buffer
-              ddsi_sertype_serialize_into(current_blob->type, current_blob->iox_chunk, buffer.data(),
-                                          serialized_size);
-              // update the CDR blob with the serialized data
-              copy_buffer_to_cdr_blob(buffer.data(), serialized_size, sample_data.kind(), sample_data);
-            } else {
-              // this shouldn't never happen
-              ISOCPP_THROW_EXCEPTION(ISOCPP_PRECONDITION_NOT_MET_ERROR,
-                                     "The received sample over SHM is not initialized");
-            }
-            // release the chunk
-            free_iox_chunk(static_cast<iox_sub_t *>(current_blob->iox_subscriber), &current_blob->iox_chunk);
-        } else
-#endif
-        {
-          std::cerr << "CDR sample holder, no shm, copying blob\n";
+
+        // if data is transferred using SHM, update the CDRBlob with iox_chunk
+        if(!update_cdrblob_from_iox_chunk(*current_blob, sample_data)) {
           ddsrt_iovec_t blob_content;
           ddsi_serdata_to_ser_ref(current_blob, 0, ddsi_serdata_size(current_blob), &blob_content);
           copy_buffer_to_cdr_blob(reinterpret_cast<uint8_t *>(blob_content.iov_base),
@@ -181,6 +153,7 @@ public:
           ddsi_serdata_to_ser_unref(current_blob, &blob_content);
           ddsi_serdata_unref(current_blob);
         }
+        // copy sample infos
         org::eclipse::cyclonedds::sub::AnyDataReaderDelegate::copy_sample_infos(info[i], (*samples_.delegate())[i].delegate().info());
       }
     }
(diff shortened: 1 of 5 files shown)


sumanth-nirmal:
@eboasson and @MatthiasKillat Can you please take a look at the changes? I have updated with the current state of https://github.com/eclipse-cyclonedds/cyclonedds/pull/1031

src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
```diff
@@ -139,8 +139,39 @@ class CDRSamplesHolder : public SamplesHolder
...
+            // if the iox chunk has the data in serialized form
+            if (iox_header->shm_data_state == IOX_CHUNK_CONTAINS_SERIALIZED_DATA) {
+              // get the actual data
+              sample_data.payload().assign(
+                reinterpret_cast<uint8_t *>(current_blob->iox_chunk) + 4,
+                reinterpret_cast<uint8_t *>(current_blob->iox_chunk) + iox_header->data_size);
+            } else {
+              // this shouldn't never happen
+              ISOCPP_THROW_EXCEPTION(ISOCPP_PRECONDITION_NOT_MET_ERROR,
+                "The received sample over SHM is not in serialized form");
```
sumanth-nirmal, line 166: Probably this assumption is not always true!

sumanth-nirmal: The scenario that the iox chunk containing the raw data should also be handled here, I will modify this accordingly

eboasson: It's probably worth splitting this out into a separate function. Also we should do something about the `+4`s ...

src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
```diff
@@ -139,8 +139,39 @@ class CDRSamplesHolder : public SamplesHolder
       uint32_t cpp_sample_size = this->samples_.delegate()->length();
       for (uint32_t i = 0; i < cpp_sample_size; ++i)
       {
+        struct ddsi_serdata * current_blob = cdr_blobs[i];
+#ifdef DDS_HAS_SHM
+        // if the data is over iox_chunk
```
MatthiasKillat, line 144: typo? over

src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
```diff
@@ -139,8 +139,39 @@ class CDRSamplesHolder : public SamplesHolder
...
+          // update the header
+          memcpy(sample_data.encoding().data(), current_blob->iox_chunk, 4);
+          // update the data kind
+          sample_data.kind(
+            static_cast<org::eclipse::cyclonedds::topic::BlobKind>(current_blob->kind));
+          if (sample_data.kind() != org::eclipse::cyclonedds::topic::BlobKind::Empty) {
+            // get the size
+            auto iox_header = iceoryx_header_from_chunk(current_blob->iox_chunk);
+            // if the iox chunk has the data in serialized form
+            if (iox_header->shm_data_state == IOX_CHUNK_CONTAINS_SERIALIZED_DATA) {
```
MatthiasKillat, line 158: The state can also now be obtained with `iox_shm_data_state_t shm_get_data_state(void *iox_chunk);` Recommended if the header is not required for anything else.

src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
MatthiasKillat: Right., the 4 should be the CDR header size constant that should exist somewhere, if not has be created. 

@sumanth-nirmal I am not sure we can get raw (original, unserialized) data here, but if so yes it should be handled This case must have been handled before (fixed-size types) somewhere though ...

src/ddscxx/src/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.cpp
```diff
@@ -22,10 +22,15 @@
...
 #include <dds/dds.h>
-#include <dds/ddsc/dds_data_allocator.h>
+#include <dds/ddsc/dds_loan_api.h>
 
 #include "dds/ddsi/ddsi_sertopic.h"
 #include "dds/ddsi/q_protocol.h"
+#include "dds/features.hpp"
+
+#ifdef DDSCXX_HAS_SHM
+#include <dds/ddsi/shm_transport.h>
```
MatthiasKillat, line 32: Not for this PR but we probably need to expose the functionality in a C header and be able to use it without requiring the `#ifdef` here ...

src/ddscxx/src/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.cpp
```diff
@@ -116,6 +121,29 @@ AnyDataWriterDelegate::write_cdr(
...
+        // request a loan from the shared memory buffer
+        ret = dds_loan_shared_memory_buffer(writer,
+                                            data->payload().size() + data->encoding().size(),  // include the size for the encoding (CDR header)
+                                            &iox_chunk);
+        ISOCPP_DDSC_RESULT_CHECK_AND_THROW(ret, "write_cdr - Loaning of chunk failed");
+        ISOCPP_BOOL_CHECK_AND_THROW(iox_chunk, ISOCPP_NULL_REFERENCE_ERROR, "write_cdr - Loaning of chunk failed");
+        // copy the header
+        memcpy(iox_chunk, data->encoding().data(), data->encoding().size());
+        // copy the actual data
+        memcpy(static_cast<unsigned char *>(iox_chunk) + data->encoding().size(),
```
MatthiasKillat, line 138: Copying the serialized data later after it has been serialized is also a source of severe inefficiencies. Ideally it should already have been serialized to the buffer. I assume that due to the existing architecture it may be needed n other memory of the serialized sample representation as well, but I am not sure.

src/ddscxx/src/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.cpp
```diff
@@ -124,7 +152,18 @@ AnyDataWriterDelegate::write_cdr(
         ret = dds_writecdr(writer, ser_data);
     }
 
+#ifdef DDSCXX_HAS_SHM
+    if (ret != DDS_RETCODE_OK) {
+        if (ser_data->iox_chunk != nullptr) {
+            // write cdr failed, so free the chunk
+            ret = dds_return_loan(writer, &ser_data->iox_chunk, 1);
```
MatthiasKillat, line 159: We need to double check whether return loan works with `dds_loan_shared_memory_buffer`, but it should (it is just delegated to iceoryx) There are more tests needed ...

src/ddscxx/include/dds/sub/detail/SamplesHolder.hpp
sumanth-nirmal: I need the header for the size of the chunk, so I was not using `shm_get_data_state()`


Ideally, the interface might need to have proper setters and getters for the fields in the user header, we can probably do as a followup

src/ddscxx/src/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.cpp
sumanth-nirmal: Since we don't have a user API to loan the memory and fill the serialized sample, we get the serialized sample as `CDRBlob` then copy this into the `iox_chunk`, which will then be eventually transferred using shm

src/ddscxx/src/org/eclipse/cyclonedds/pub/AnyDataWriterDelegate.cpp
sumanth-nirmal: This should be tested indirectly with the tests in the upper layers

Agree, we might need to add some tests for these new shm APIs (at some point)

sumanth-nirmal:
@eboasson @MatthiasKillat I think I have addressed the above issues, please take a look.

Also, I used a temporary commit (which is dropped now) to test the CI with the branch here https://github.com/eclipse-cyclonedds/cyclonedds/pull/1031 and it is green https://dev.azure.com/eclipse-cyclonedds/cyclonedds-cxx/_build/results?buildId=1527&view=results

sumanth-nirmal:
@eboasson I consolidated the code blocks handling the iox_chunk into API's so now it should not have a mix of pre-processing conditionals and branching, please take a look

eboasson (approved):
Thanks @sumanth-nirmal !

Issue #116, opened 2021-06-30
Compilation Error with generated code from idlcxx


dkroenke:
CycloneDDS version: https://github.com/eclipse-cyclonedds/cyclonedds/commit/7bc432683690ba860a72c662395c07835ce0b3a4
CYcloneDDS-CXX version: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/0add05083ef4c560eb2641ada38432f53d0c3cc8

When using the idlcxx generator in iceoryx, there are some compile errors with the generated code:
```bash
In file included from iceoryx/iceoryx_dds/include/iceoryx_dds/dds/cyclone_data_writer.hpp:23,
                 from iceoryx/iceoryx_dds/source/iceoryx_dds/dds/cyclone_data_writer.cpp:18:
iceoryx/build/iceoryx_dds_messages/Mempool.hpp: In static member function ‘static ddsi_sertype* org::eclipse::cyclonedds::topic::TopicTraits<Mempool::Chunk>::getSerType()’:
iceoryx/build/iceoryx_dds_messages/Mempool.hpp:72:20: error: ‘ddscxx_sertype’ does not name a type; did you mean ‘ddsi_sertype’?
   72 |     auto *st = new ddscxx_sertype<::Mempool::Chunk>();
      |                    ^~~~~~~~~~~~~~
      |                    ddsi_sertype
iceoryx/build/iceoryx_dds_messages/Mempool.hpp:72:51: error: expected primary-expression before ‘>’ token
   72 |     auto *st = new ddscxx_sertype<::Mempool::Chunk>();
      |                                                   ^
iceoryx/build/iceoryx_dds_messages/Mempool.hpp:72:53: error: expected primary-expression before ‘)’ token
   72 |     auto *st = new ddscxx_sertype<::Mempool::Chunk>();
      |                                                     ^
iceoryx/build/iceoryx_dds_messages/Mempool.hpp: At global scope:
iceoryx/build/iceoryx_dds_messages/Mempool.hpp:96:8: error: ‘topic_type_name’ is not a class template
   96 | struct topic_type_name<::Mempool::Chunk>
      |        ^~~~~~~~~~~~~~~
iceoryx/build/iceoryx_dds_messages/Mempool.hpp:97:1: error: explicit specialization of non-template ‘dds::topic::topic_type_name’
   97 | {
      | ^
[...]

PR #178, opened 2021-11-19
Xcdrv2

reicheratwork:
Work towards an implementation that supports extended types serialization in C++:

The general concept is as follows:
  - idlcxx will create a tree of "entity_properties" representing the
    data structures
  - Each C++ data structure representation has switchboxes inside the
    streaming function implementations linking entity_properties to
    class member
  - The different streamer implementations will iterate over this
    tree differently and write headers etc. based on the properties
    provided

### DIFF
--- a/src/ddscxx/CMakeLists.txt
+++ b/src/ddscxx/CMakeLists.txt
@@ -35,6 +35,10 @@ set(sources
     src/org/eclipse/cyclonedds/core/EntitySet.cpp
     src/org/eclipse/cyclonedds/core/MiscUtils.cpp
     src/org/eclipse/cyclonedds/core/cdr/cdr_stream.cpp
+    src/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.cpp
+    src/org/eclipse/cyclonedds/core/cdr/entity_properties.cpp
+    src/org/eclipse/cyclonedds/core/cdr/extended_cdr_v1_ser.cpp
+    src/org/eclipse/cyclonedds/core/cdr/extended_cdr_v2_ser.cpp
     src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
     src/org/eclipse/cyclonedds/core/cond/GuardConditionDelegate.cpp
     src/org/eclipse/cyclonedds/core/cond/StatusConditionDelegate.cpp
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
@@ -48,18 +48,17 @@ public:
    */
   void skip_entity(const entity_properties_t &) { status(serialization_status::unsupported_xtypes); }
 
-    /**
-     * @brief
-     * Function declaration for starting a parameter list.
-     *
-     * This function is called by the generated functions for the entity, and will trigger the necessary actions on starting a new struct.
-     * I.E. starting a new parameter list, writing headers.
-     *
-     * @param[in,out] props The entity whose members might be represented by a parameter list.
-     *
-     * @return Whether the operation was completed succesfully.
-     */
-    bool start_struct(entity_properties_t &props);
+  /**
+   * @brief
+   * Start a new struct.
+   *
+   * This function is called by the generated streaming functions, and will start a parameter list, if that is relevant for it.
+   *
+   * @param[in, out] props The entity whose members might be represented by a parameter list.
+   *
+   * @return Whether the operation was completed succesfully.
+   */
+  bool start_struct(entity_properties_t &props);
 
   /**
    * @brief
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
@@ -13,13 +13,6 @@
 #define BASIC_CDR_SERIALIZATION_HPP_
 
 #include "cdr_stream.hpp"
-#include <org/eclipse/cyclonedds/core/type_helpers.hpp>
-#include <dds/core/Exception.hpp>
-#include <array>
-#include <vector>
-#include <string>
-#include <cstring>
-#include <algorithm>
 
 namespace org {
 namespace eclipse {
@@ -33,7 +26,7 @@ namespace cdr {
  *
  * This type of cdr stream has a maximum alignment of 8 bytes.
  */
-class basic_cdr_stream : public cdr_stream {
+class OMG_DDS_API basic_cdr_stream : public cdr_stream {
 public:
   /**
    * @brief
@@ -41,301 +34,65 @@ public:
    *
    * Basically a pass through for the cdr_stream base class.
    *
+   * @param[in] end The endianness to set for the data stream, default to the local system endianness.
    * @param[in] ignore_faults Bitmask for ignoring faults, can be composed of bit fields from the serialization_status enumerator.
    */
-  basic_cdr_stream(uint64_t ignore_faults = 0x0) : cdr_stream(8, ignore_faults) { ; }
+  basic_cdr_stream(endianness end = native_endianness(), uint64_t ignore_faults = 0x0) : cdr_stream(end, 8, ignore_faults) { ; }
+
+  /**
+   * @brief
+   * Starts a member.
+   *
+   * As the basic cdr stream does not have anything that requires delimiting between entities, this function does nothing.
+   */
+  void start_member(entity_properties_t &, bool) {;}
+
+  /**
+   * @brief
+   * Finishes a member.
+   *
+   * As the basic cdr stream does not have anything that requires delimiting between entities, this function does nothing.
+   */
+  void finish_member(entity_properties_t &, bool) {;}
+
+  /**
+   * @brief
+   * Starts a new struct.
+   *
+   * As the basic cdr stream does not have anything that requires delimiting between entities, this function does nothing.
+   */
+  void start_struct(entity_properties_t &) {;}
+
+  /**
+   * @brief
+   * Finishes the current struct.
+   *
+   * As the basic cdr stream does not have anything that requires delimiting between entities, this function does nothing.
+   */
+  void finish_struct(entity_properties_t &) {;}
+
+  /**
+   * @brief
+   * Skips an entity, bypassing the stack.
+   *
+   * As the basic cdr stream does not have anything that requires delimiting between entities, this function does nothing.
+   */
+  void skip_entity(const entity_properties_t &) {;}
+
+  /**
+   * @brief
+   * Returns the next entity to be processed.
+   *
+   * This implementation directly passes through to next_prop, since pops and pushes do nothing.
+   *
+   * @param[in, out] props The property tree to get the next entity from.
+   * @param[in, out] firstcall Whether it is the first time calling the function for props, will store first iterator if true, and then set to false.
+   *
+   * @return The next entity to be processed, or the final entity if the current tree level does not hold more entities.
+   */
+  entity_properties_t& next_entity(entity_properties_t &props, bool &firstcall);
+
 };
-
-/**
- * @brief
- * Primitive type stream manipulation functions.
- *
- * These are "endpoints" for write functions, since composit
- * (sequence/array/constructed type) functions will decay to these
- * calls.
- */
-
-/**
- * @brief
- * Primitive type read function.
- *
- * Aligns the stream to the alignment of type T.
- * Reads the value from the current position of the stream str into
- * toread.
- * Moves the cursor of the stream by the size of T.
- * This function is only enabled for arithmetic types and enums.
- *
- * @param[in, out] str The stream which is read from.
- * @param[out] toread The variable to read into.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void read(basic_cdr_stream &str, T& toread)
-{
-  if (str.abort_status())
-    return;
-
-  str.align(sizeof(T), false);
-
-  toread = *static_cast<T*>(str.get_cursor());
-
-  str.incr_position(sizeof(T));
-}
-
-/**
- * @brief
- * Primitive type read function, byteswap version.
- *
- * Same as read(), with an additional byteswap at the end.
- *
- * @param[in, out] str The stream which is read from.
- * @param[out] toread The variable to read into.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void read_swapped(basic_cdr_stream &str, T& toread)
-{
-  read(str, toread);
-
-  byte_swap(toread);
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void read_many(basic_cdr_stream &str, T *out, size_t N)
-{
-  if (str.abort_status() || N == 0)
-    return;
-
-  str.align(sizeof(T), false);
-
-  T *in = static_cast<T*>(str.get_cursor());
-
-  memcpy(out, in, sizeof(T)*N);
-
-  str.incr_position(sizeof(T)*N);
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void read_many_swapped(basic_cdr_stream &str, T *out, size_t N)
-{
-  if (str.abort_status() || N == 0)
-    return;
-
-  str.align(sizeof(T), false);
-
-  T *in = static_cast<T*>(str.get_cursor());
-  for (size_t i = 0; i < N; i++, out++, in++) {
-    *out = *in;
-    byte_swap(*out);
-  }
-
-  str.incr_position(sizeof(T)*N);
-}
-
-/**
- * @brief
- * Primitive type write function.
- *
- * Aligns str to the type to be written.
- * Writes towrite to str.
- * Swaps bytes written to str if the endiannesses do not match up.
- * Moves the cursor of str by the size of towrite.
- * This function is only enabled for arithmetic types.
- *
- * @param[in, out] str The stream which is written to.
- * @param[in] towrite The variable to write.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void write(basic_cdr_stream& str, const T& towrite)
-{
-  if (str.abort_status())
-    return;
-
-  str.align(sizeof(T), true);
-
-  auto out = static_cast<T*>(str.get_cursor());
-
-  *out = towrite;
-
-  str.incr_position(sizeof(T));
-}
-
-/**
- * @brief
- * Primitive type write function, byteswap version.
- *
- * Same as write(), with an additional byteswap at the end.
- *
- * @param[in, out] str The stream which is read from.
- * @param[in] towrite The variable to write.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void write_swapped(basic_cdr_stream& str, const T& towrite)
-{
-  if (str.abort_status())
-    return;
-
-  str.align(sizeof(T), true);
-
-  auto out = static_cast<T*>(str.get_cursor());
-
-  *out = towrite;
-
-  byte_swap(*out);
-
-  str.incr_position(sizeof(T));
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void write_many(basic_cdr_stream& str, const T* in, size_t N)
-{
-  if (str.abort_status() || N == 0)
-    return;
-
-  str.align(sizeof(T), true);
-
-  T *out = static_cast<T*>(str.get_cursor());
-
-  memcpy(out, in, sizeof(T)*N);
-
-  str.incr_position(sizeof(T)*N);
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void write_many_swapped(basic_cdr_stream& str, const T* in, size_t N)
-{
-  if (str.abort_status() || N == 0)
-    return;
-
-  str.align(sizeof(T), true);
-
-  T *out = static_cast<T*>(str.get_cursor());
-  for (size_t i = 0; i < N; i++, out++, in++) {
-    *out = *in;
-    byte_swap(*out);
-  }
-
-  str.incr_position(sizeof(T)*N);
-}
-
-/**
- * @brief
- * Primitive type cursor move function.
- *
- * Used in determining the size of a type when written to the stream.
- * Aligns str to the size of toincr.
- * Moves the cursor of str by the size of toincr.
- * This function is only enabled for arithmetic types.
- *
- * @param[in, out] str The stream whose cursor is moved.
- * @param[in] toincr The variable to move the cursor by, no contents of this variable are used, it is just used to determine the template.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void move(basic_cdr_stream& str, const T& toincr)
-{
-  if (str.abort_status())
-    return;
-
-  (void)toincr;
-
-  str.align(sizeof(T), false);
-
-  str.incr_position(sizeof(T));
-}
-
-/**
- * @brief
- * Primitive type cursor move function, byteswap version.
- *
- * Same as move(), with an additional byteswap at the end.
- *
- * @param[in, out] str The stream whose cursor is moved.
- * @param[in] toincr The variable to move the cursor by, no contents of this variable are used, it is just used to determine the template.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void move_swapped(basic_cdr_stream& str, const T& toincr)
-{
-  move(str, toincr);
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void move_many(basic_cdr_stream& str, const T* toincr, size_t N)
-{
-  if (str.abort_status())
-    return;
-
-  (void)toincr;
-
-  str.align(sizeof(T), false);
-
-  str.incr_position(sizeof(T)*N);
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void move_many_swapped(basic_cdr_stream& str, const T* toincr, size_t N)
-{
-  move_many(str, toincr, N);
-}
-
-/**
- * @brief
- * Primitive type max stream move function.
- *
- * Used in determining the maximum stream size of a constructed type.
- * Moves the cursor to the maximum position it could occupy after
- * writing max_sz to the stream.
- * Is in essence the same as the primitive type cursor move function,
- * but additionally checks for whether the cursor it at the "end",
- * which may happen if unbounded members (strings/sequences/...)
- * are part of the constructed type.
- * This function is only enabled for arithmetic types.
- *
- * @param[in, out] str The stream whose cursor is moved.
- * @param[in] max_sz The variable to move the cursor by, no contents of this variable are used, it is just used to determine the template.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void max(basic_cdr_stream& str, const T& max_sz)
-{
-  if (str.abort_status())
-    return;
-
-  if (str.position() == SIZE_MAX)
-    return;
-
-  move(str, max_sz);
-}
-
-/**
- * @brief
- * Primitive type max stream move function.
- *
- * Same as max(), with an additional byteswap at the end.
- *
- * @param[in, out] str The stream whose cursor is moved.
- * @param[in] max_sz The variable to move the cursor by, no contents of this variable are used, it is just used to determine the template.
- */
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void max_swapped(basic_cdr_stream& str, const T& max_sz)
-{
-  max(str, max_sz);
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void max_many(basic_cdr_stream& str, const T* max_sz, size_t N)
-{
-  if (str.abort_status())
-    return;
-
-  if (str.position() == SIZE_MAX)
-    return;
-
-  move_many(str, max_sz, N);
-}
-
-template<typename T, std::enable_if_t<std::is_arithmetic<T>::value && !std::is_enum<T>::value, bool> = true >
-inline void max_many_swapped(basic_cdr_stream& str, const T* max_sz, size_t N)
-{
-  max_many(str, max_sz, N);
-}
-
 
 /**
  * @brief
@@ -354,66 +111,24 @@ inline void max_many_swapped(basic_cdr_stream& str, const T* max_sz, size_t N)
  *
  * @param[in, out] str The stream which is read from.
  * @param[out] toread The variable to read into.
+ * @param[in] N The number of entities to read.
  */
 template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void read(basic_cdr_stream& str, T& toread) {
-  read(str, *reinterpret_cast<uint32_t*>(&toread));
-}
-
-/**
- * @brief
- * Reads the byteswapped value of the enum from the stream.
- *
- * @param[in, out] str The stream which is read from.
- * @param[out] toread The variable to read into.
- */
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void read_swapped(basic_cdr_stream& str, T& toread) {
-  read_swapped(str, *reinterpret_cast<uint32_t*>(&toread));
-}
-
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void read_many(basic_cdr_stream& str, T* toread, size_t N) {
-  read_many(str, reinterpret_cast<uint32_t*>(toread), N);
-}
-
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void read_many_swapped(basic_cdr_stream& str, T* toread, size_t N) {
-  read_many_swapped(str, reinterpret_cast<uint32_t*>(toread), N);
+void read(basic_cdr_stream& str, T& toread, size_t N = 1) {
+  read_enum_impl<basic_cdr_stream,T,uint32_t>(str, toread, N);
 }
 
 /**
  * @brief
  * Writes the value of the enum to the stream.
  *
- * @param [in, out] str The stream which is written to.
- * @param [in] towrite The variable to write.
+ * @param[in, out] str The stream which is written to.
+ * @param[in] towrite The variable to write.
+ * @param[in] N The number of entities to write.
  */
 template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void write(basic_cdr_stream& str, const T& towrite) {
-  write(str, uint32_t(towrite));
-}
-
-/**
- * @brief
- * Writes the byteswapped value of the enum to the stream.
- *
- * @param [in, out] str The stream which is written to.
- * @param [in] towrite The variable to write.
- */
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void write_swapped(basic_cdr_stream& str, const T& towrite) {
-  write_swapped(str, uint32_t(towrite));
-}
-
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void write_many(basic_cdr_stream& str, const T* towrite, size_t N) {
-  write_many(str, static_cast<uint32_t*>(towrite), N);
-}
-
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void write_many_swapped(basic_cdr_stream& str, const T* towrite, size_t N) {
-  write_many_swapped(str, static_cast<uint32_t*>(towrite), N);
+void write(basic_cdr_stream& str, const T& towrite, size_t N = 1) {
+  write_enum_impl<basic_cdr_stream,T,uint32_t>(str, towrite, N);
 }
 
 /**
@@ -421,33 +136,11 @@ inline void write_many_swapped(basic_cdr_stream& str, const T* towrite, size_t N
  * Moves the cursor of the stream by the size the enum would take up.
  *
  * @param[in, out] str The stream whose cursor is moved.
- * @param[in] toincr The variable to move the cursor by, no contents of this variable are used, it is just used to determine the template.
+ * @param[in] N The number of entities to move.
  */
 template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void move(basic_cdr_stream& str, const T& toincr) {
-  move(str, uint32_t(toincr));
-}
-
-/**
- * @brief
- * Moves the cursor of the stream by the size the enum would take up (byteswapped version).
- *
- * @param[in, out] str The stream whose cursor is moved.
- * @param[in] toincr The variable to move the cursor by, no contents of this variable are used, it is just used to determine the template.
- */
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void move_swapped(basic_cdr_stream& str, const T& toincr) {
-  move_swapped(str, uint32_t(toincr));
-}
-
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void move_many(basic_cdr_stream& str, const T* toincr, size_t N) {
-  move_many(str, static_cast<uint32_t*>(toincr), N);
-}
-
-template<typename T, std::enable_if_t<std::is_enum<T>::value && !std::is_arithmetic<T>::value, bool> = true >
-inline void move_many_swapped(basic_cdr_stream& str, const T* toincr, size_t N) {
-  move_many_swapped(str, static_cast<uint32_t*>(toincr), N);
+void move(basic_cdr_stream& str, const T&, size_t N = 1) {
+  move(str, uint32_t(0), N);
 }
 
 /**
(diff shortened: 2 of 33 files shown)


ichernev:
What is the status of this PR? I see there is support for optional fields, but also v2 delimited + property-list. Is there much work left?

reicheratwork:
> What is the status of this PR? I see there is support for optional fields, but also v2 delimited + property-list. Is there much work left?

While this PR is still a work in progress, I think that most implementation details have been hammered out enough that most  work to be done is fixing busgs and implementing edge cases.
I think it has progressed to a mature enough stage that I wanted people interested in it to have a look at it, and maybe do a little interactive review?

If you could also try it and give me feedback, that would be much appreciated.

reicheratwork:
> I'll be glad to test it (mostly interested in optionals), and review to the best of my ability.

Thanks!

ichernev:
I managed to compile with latest cyclonedds (https://github.com/eclipse-cyclonedds/cyclonedds/commit/717d9d7504670a20d96f7e74c00fb6ddc7dfc282) and this branch (https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/355310013bfad27f44127e9b251159f1ffbbd22b), and I managed to send a few packets. The receiver complains (malformed packet):

```
IDL file
--------

module HelloWorldData
{
  struct Msg
  {
    long userID;
    @optional string message; // present in even packets, absent in odd packets
  };
};


Log from publisher
------------------

    === [Publisher] Write sample. 1
    === [Publisher] Write sample. 2
    === [Publisher] Write sample. 3
    === [Publisher] Write sample. 4

Log from subscriber
-------------------

    1638961764.703483 [0]     recvUC: malformed packet received from vendor 1.16 state parse:data <52545053 02010110 0d6a1001 45f8fdfd fb354350 09010800 6492b061 9ed904b4 @0x20 15052000 00001000 00000000 03010000 00000000 02000000 00110001 01000000> (note: maybe partially bswap'd) {{15,5,32},0,16,0,103,2}
    1638961765.704346 [0]     recvUC: malformed packet received from vendor 1.16 state parse:data <52545053 02010110 0d6a1001 45f8fdfd fb354350 09010800 6592b061 56f32fb4 @0x20 15053000 00001000 00000000 03010000 00000000 03000000 00110000 02000000> (note: maybe partially bswap'd) {{15,5,48},0,16,0,103,3}
    1638961766.704753 [0]     recvUC: malformed packet received from vendor 1.16 state parse:data <52545053 02010110 0d6a1001 45f8fdfd fb354350 09010800 6692b061 b91e5ab4 @0x20 15052000 00001000 00000000 03010000 00000000 04000000 00110001 03000000> (note: maybe partially bswap'd) {{15,5,32},0,16,0,103,4}
    1638961767.705209 [0]     recvUC: malformed packet received from vendor 1.16 state parse:data <52545053 02010110 0d6a1001 45f8fdfd fb354350 09010800 6792b061 8cf575b4 @0x20 15053000 00001000 00000000 03010000 00000000 05000000 00110000 04000000> (note: maybe partially bswap'd) {{15,5,48},0,16,0,103,5}

Captured
--------

    # packet 1 (packet ID in wireshark: 12)
    # full UDP packet (INFO_TS, DATA, HEARTBEAT)
    0000   00 00 00 00 00 00 00 00 00 00 00 00 08 00 45 00   ..............E.
    0010   00 80 9a 46 40 00 40 11 89 f9 c0 a8 4a 6e c0 a8   ...F@.@.....Jn..
    0020   4a 6e 84 35 81 0d 00 6c 16 ab 52 54 50 53 02 01   Jn.5...l..RTPS..
    0030   01 10 01 10 6a 0d fd fd f8 45 50 43 35 fb 09 01   ....j....EPC5...
    0040   08 00 64 92 b0 61 9e d9 04 b4 15 05 20 00 00 00   ..d..a...... ...
    0050   10 00 00 00 00 00 00 00 01 03 00 00 00 00 02 00   ................
    0060   00 00 00 11 00 01 01 00 00 00 00 00 00 00 07 01   ................
    0070   1c 00 00 00 00 00 00 00 01 03 00 00 00 00 03 00   ................
    0080   00 00 00 00 00 00 02 00 00 00 01 00 00 00         ..............

    # only serialized DATA
    0000   00 11 00 01 01 00 00 00 00 00 00 00               ............


    # packet 2 (packet ID in wireshark: 14)
    # full UDP packet (INFO_TS, DATA, HEARTBEAT)
    0000   00 00 00 00 00 00 00 00 00 00 00 00 08 00 45 00   ..............E.
    0010   00 90 9a c7 40 00 40 11 89 68 c0 a8 4a 6e c0 a8   ....@.@..h..Jn..
    0020   4a 6e 84 35 81 0d 00 7c 16 bb 52 54 50 53 02 01   Jn.5...|..RTPS..
    0030   01 10 01 10 6a 0d fd fd f8 45 50 43 35 fb 09 01   ....j....EPC5...
    0040   08 00 65 92 b0 61 56 f3 2f b4 15 05 30 00 00 00   ..e..aV./...0...
    0050   10 00 00 00 00 00 00 00 01 03 00 00 00 00 03 00   ................
    0060   00 00 00 11 00 00 02 00 00 00 01 00 00 00 0c 00   ................
    0070   00 00 48 65 6c 6c 6f 20 77 6f 72 6c 64 00 07 01   ..Hello world...
    0080   1c 00 00 00 00 00 00 00 01 03 00 00 00 00 04 00   ................
    0090   00 00 00 00 00 00 03 00 00 00 02 00 00 00         ..............

    # only serialized DATA
    0000   00 11 00 00 02 00 00 00 01 00 00 00 0c 00 00 00   ................
    0010   48 65 6c 6c 6f 20 77 6f 72 6c 64 00               Hello world.


    # packet 3 (packet ID in wireshark: 16)
    # full UDP packet (INFO_TS, DATA, HEARTBEAT)
    0000   00 00 00 00 00 00 00 00 00 00 00 00 08 00 45 00   ..............E.
    0010   00 80 9b 7a 40 00 40 11 88 c5 c0 a8 4a 6e c0 a8   ...z@.@.....Jn..
    0020   4a 6e 84 35 81 0d 00 6c 16 ab 52 54 50 53 02 01   Jn.5...l..RTPS..
    0030   01 10 01 10 6a 0d fd fd f8 45 50 43 35 fb 09 01   ....j....EPC5...
    0040   08 00 66 92 b0 61 b9 1e 5a b4 15 05 20 00 00 00   ..f..a..Z... ...
    0050   10 00 00 00 00 00 00 00 01 03 00 00 00 00 04 00   ................
    0060   00 00 00 11 00 01 03 00 00 00 00 00 00 00 07 01   ................
    0070   1c 00 00 00 00 00 00 00 01 03 00 00 00 00 05 00   ................
    0080   00 00 00 00 00 00 04 00 00 00 03 00 00 00         ..............

    # only seriliazed DATA
    0000   00 11 00 01 03 00 00 00 00 00 00 00               ............

    # packet 4 (packet ID in wireshark: 17)
    # full UDP packet (INFO_TS, DATA, HEARTBEAT)
    0000   00 00 00 00 00 00 00 00 00 00 00 00 08 00 45 00   ..............E.
    0010   00 90 9b e8 40 00 40 11 88 47 c0 a8 4a 6e c0 a8   ....@.@..G..Jn..
    0020   4a 6e 84 35 81 0d 00 7c 16 bb 52 54 50 53 02 01   Jn.5...|..RTPS..
    0030   01 10 01 10 6a 0d fd fd f8 45 50 43 35 fb 09 01   ....j....EPC5...
    0040   08 00 67 92 b0 61 8c f5 75 b4 15 05 30 00 00 00   ..g..a..u...0...
    0050   10 00 00 00 00 00 00 00 01 03 00 00 00 00 05 00   ................
    0060   00 00 00 11 00 00 04 00 00 00 01 00 00 00 0c 00   ................
    0070   00 00 48 65 6c 6c 6f 20 77 6f 72 6c 64 00 07 01   ..Hello world...
    0080   1c 00 00 00 00 00 00 00 01 03 00 00 00 00 06 00   ................
    0090   00 00 00 00 00 00 05 00 00 00 04 00 00 00         ..............

    # only serialized DATA
    0000   00 11 00 00 04 00 00 00 01 00 00 00 0c 00 00 00   ................
    0010   48 65 6c 6c 6f 20 77 6f 72 6c 64 00               Hello world.
```

Link to [full capture (Wireshark)](https://www.dropbox.com/s/60inh6d0vfwa960/buggy_receiver_2.pcapng?dl=0).

I'll also try to investigate.

dennis-adlink:
Looks like the RTPS encapsulation identifier in the data (0x00 0x11) is not what the Cyclone C serializer expects. There is an inconsistency in the OMG XTypes specification (https://issues.omg.org/issues/DDSXTY14-20), so that might be the cause of this issue. I think the values from table 60 from the specification are used by other DDS implementations, so in that case we need to change this in the C++ serializer, @e-hndrks can you confirm that table 60 has the correct identifiers?

ichernev:
> Looks like the RTPS encapsulation identifier in the data (0x00 0x11) is not what the Cyclone C serializer expects. There is an inconsistency in the OMG XTypes specification (https://issues.omg.org/issues/DDSXTY14-20), so that might be the cause of this issue. I think the values from table 60 from the specification are used by other DDS implementations, so in that case we need to change this in the C++ serializer, @e-hndrks can you confirm that table 60 has the correct identifiers?

I just want to make it clear that I'm using cyclone as both publisher and subscriber, so as long as it "understands itself" it should work (even outside of the spec). I'm not suggesting not to follow the spec, just clarifying.

reicheratwork:
> > Looks like the RTPS encapsulation identifier in the data (0x00 0x11) is not what the Cyclone C serializer expects. There is an inconsistency in the OMG XTypes specification (https://issues.omg.org/issues/DDSXTY14-20), so that might be the cause of this issue. I think the values from table 60 from the specification are used by other DDS implementations, so in that case we need to change this in the C++ serializer, @e-hndrks can you confirm that table 60 has the correct identifiers?
> 
> I just want to make it clear that I'm using cyclone as both publisher and subscriber, so as long as it "understands itself" it should work (even outside of the spec). I'm not suggesting not to follow the spec, just clarifying.

Completely true, and while it is true that both the C++ and C implementations follow the spec, only the C++ binding used the values in table 39 of the XCDR v1.3 spec and the C binding those from table 60.

As soon as I know which of the two is the "correct" one, I will fix this.

ichernev:
In the mean time this does the trick:

```patch
diff --git a/src/core/ddsi/include/dds/ddsi/q_protocol.h b/src/core/ddsi/include/dds/ddsi/q_protocol.h
index c7266c51..0cbf007e 100644
--- a/src/core/ddsi/include/dds/ddsi/q_protocol.h
+++ b/src/core/ddsi/include/dds/ddsi/q_protocol.h
@@ -193,16 +193,16 @@ typedef struct InfoSRC {
 #define CDR_LE      0x0100u
 #define PL_CDR_BE   0x0200u
 #define PL_CDR_LE   0x0300u
-#define CDR2_BE     0x0600u
-#define CDR2_LE     0x0700u
-#define D_CDR2_BE   0x0800u
-#define D_CDR2_LE   0x0900u
-#define PL_CDR2_BE  0x0a00u
-#define PL_CDR2_LE  0x0b00u
+#define CDR2_BE     0x1000u
+#define CDR2_LE     0x1100u
+#define PL_CDR2_BE  0x1200u
+#define PL_CDR2_LE  0x1300u
+#define D_CDR2_BE   0x1400u
+#define D_CDR2_LE   0x1500u
 
 #define CDR_ENC_LE(x) (((x) & 0x0100) == 0x0100)
 #define CDR_ENC_IS_NATIVE(x) (CDR_ENC_LE ((x)))
-#define CDR_ENC_IS_VALID(x) (!((x) > PL_CDR2_LE || (x) == 0x0400 || (x) == 0x0500))
+#define CDR_ENC_IS_VALID(x) ((x) >= CDR_BE && (x) <= PL_CDR_LE || (x) >= CDR2_BE && (x) <= D_CDR2_LE)
 #define CDR_ENC_TO_NATIVE(x) ((x) | 0x0100)
 #else
 #define CDR_BE      0x0000u
```

ichernev:
https://github.com/reicheratwork/cyclonedds-cxx/pull/5 -- fix generator for optional sequences.

reicheratwork:
@ichernev Thanks for the checks!

I changed the RTPS encapsulation identifiers to match those from CycloneDDS (so table 60) after WireShark also did not recognize those from table 39.

reicheratwork:
> [reicheratwork#5](https://github.com/reicheratwork/cyclonedds-cxx/pull/5) -- fix generator for optional sequences.

Thanks for finding this edge case, though I will not be using your PR as I think the `.value()` approach is more "C++ -like", though I will credit you accordingly in my commit of course.
I also added some checks in the unittests for this case.

ichernev:
> I changed the RTPS encapsulation identifiers to match those from CycloneDDS (so table 60) after WireShark also did not recognize those from table 39.

Great! Something in me says that it should be possible to reuse those values from the C header, but there they have endianness built-in... still some refactoring can bring it to sanity.

> Thanks for finding this edge case, though I will not be using your PR as I think the .value() approach is more "C++ -like", though I will credit you accordingly in my commit of course.

Thank you! I also think `.value()` is better, I just didn't want to change more characters than necessary to make it work :)

About the generated code in general I have a few pointers:
* I don't really understand why you do this loop + switch, when you can simply "paste" each switch case in order, you just need to have an array of properties
```C++
auto props = /* get properties for this type */;
// body of case 1
handleProp(props[1], obj.prop1());
// body of case 2
handleProp(props[2], obj.prop2());
...
```
* Also, all these ifs for control flow, this is begging to be turned to exceptions. The handling can happen "above" (i.e calling code, so the generated code would be shorter and cleaner.
* Optional handling can be handled with a simple template method, same as you have simple type handling via templates, `std::optional<T>` can be implemented to depend on the serialization of `T`, so code is not repeated over and over
* Sequences can also have generic handling (implemented once), that depends on serialization of the inner type and not copy bound checking all over
* All of the traversal methods are pretty much copy-paste (for `read`, `write`, `max` etc). This can also be achieved with only a single template method (the current method is templated already):
```C++
template <typename Observer>
void observe(Observer o, MyType t) {
  auto props = TypeTraits<T>::getProps();
  o.handleField(props[1], t.field1());
  o.handleField(props[2], t.field2());
}
```
The idea being, that `handleField` or whatever the interface can be implemented for all the different traversals (read, write, max), and the traversing code will stay the same, so no need to copy paste the same over and over, let the compiler copy paste via templates :)

The only thing that really needs to happen is for something else (the observer) to be given each field+field description so it can handle it. A bit of template tweaking (so `const MyType &` and `MyType &` are both allowed) might be necessary to have const & non-const versions so writing works with const objects, but that is doable with templates only.

reicheratwork:
> About the generated code in general I have a few pointers:
> 
> * I don't really understand why you do this loop + switch, when you can simply "paste" each switch case in order, you just need to have an array of properties
> 
> ```c++
> auto props = /* get properties for this type */;
> // body of case 1
> handleProp(props[1], obj.prop1());
> // body of case 2
> handleProp(props[2], obj.prop2());
> ...
> ```

The reason switched to a while getting the field information (the `entity_properties` objects) is so that i could have the generated code differ the least between the different functions (read/write/move/max), as the production of the field information whether it is reading/writing/etc can be abstracted into the cdr streamer implementations.
That way, if you get non-sequential fields (such as in the case of reading a parameter list), you can just have the streamer produce the information it derives from the headers and then pass it to the switch cases, which is basically a map from the member_id to the member in the class. In the other cases the streamer just goes over the list of fields and produces them in sequence.

reicheratwork:
> * Optional handling can be handled with a simple template method, same as you have simple type handling via templates, `std::optional<T>` can be implemented to depend on the serialization of `T`, so code is not repeated over and over

I would love to hear more about this, though I chose this implementation, as there is the facility to supply other template implementations in the idl generator (such as vectors/arrays/... and optionals).

ichernev:
>  That way, if you get non-sequential fields (such as in the case of reading a parameter list), you can just have the streamer produce the information it derives from the headers and then pass it to the switch cases, which is basically a map from the member_id to the member in the class. In the other cases the streamer just goes over the list of fields and produces them in sequence.

Ah, I forgot about that. So the for loop+switch makes sense, for reading at least. For all other functions less so :)

ichernev:
> I would love to hear more about this, though I chose this implementation, as there is the facility to supply other template implementations in the idl generator (such as vectors/arrays/... and optionals).

My idea is, to have a single function `handleMember(props, obj.member())` that can be template-specified for integers/doubles/enums (the way it is now) but also std::optional and std::vector, so it depends on the inner T's implementation. Also, because there are many serialization options (simple, delimited, property-list), I am guessing a different implementation of handle member for each of those is in order but the "traverse all fields" method can stay the same.

You can also merge it as is and we can make fine-tuning and adjustments later on, I don't think the generated field-traversal methods are part of the "public API" so they can be changed if necessary later.

ichernev:
Is support for `ostream operator<<` something that is planned/interested in merging?

The way it is handled in RTI is that each field individually can be `ostream << field`-ed, which makes structure support trivial, but they have a custom array type `rti::core::bounded_sequence`, which they can implement << on. To make it work for `vector<T>`, I'm thinking about a small template helper class that will wrap the vector and implement the operator. Same can happen for `std::optional` (or any other built-in types that are used as "carriers" of the DDS data types).

reicheratwork:
> 



> > That way, if you get non-sequential fields (such as in the case of reading a parameter list), you can just have the streamer produce the information it derives from the headers and then pass it to the switch cases, which is basically a map from the member_id to the member in the class. In the other cases the streamer just goes over the list of fields and produces them in sequence.
> 
> Ah, I forgot about that. So the for loop+switch makes sense, for reading at least. For all other functions less so :)

I implemented it the same way all operations (reading/writing/etc.) to simplify the generation of the code and the machinery for the different serialization modes. Less code to check which might contain bugs 👍  .

reicheratwork:
> 


> > I would love to hear more about this, though I chose this implementation, as there is the facility to supply other template implementations in the idl generator (such as vectors/arrays/... and optionals).
> 
> My idea is, to have a single function `handleMember(props, obj.member())` that can be template-specified for integers/doubles/enums (the way it is now) but also std::optional and std::vector, so it depends on the inner T's implementation. Also, because there are many serialization options (simple, delimited, property-list), I am guessing a different implementation of handle member for each of those is in order but the "traverse all fields" method can stay the same.
> 
> You can also merge it as is and we can make fine-tuning and adjustments later on, I don't think the generated field-traversal methods are part of the "public API" so they can be changed if necessary later.

Thank you very much for this suggestion, I also agree that this is something that does not have to be a part of the current API as this is not part of the user facing interface!

After looking into this I am currently working on an approach that would have all the sequence/array handling done through templates, so that all the actions for reading sequence sizes, checking bounds and iterating over members are handled by the template and no code would need to be generated for that. I am still looking in to handling optional fields through templates as well.
This drastically cuts down on the amount of generated code, and hands over a lot of the correctness checking to the compiler, which should reduce the probability of errors.

ichernev:
Hello, I want to report an issue I recently discovered. The keyed/not-keyed property of the message is flipped. I.e if I have a `@key` field in the idl, the message is not-keyed, and if there is no `@key` property, the message *is* keyed. This causes RTI and Cyclone based implementations to not match on topics (they used to match before this MR is applied). If I give one impl `@key` and the other no `@key`, it matches again :)

I can see that `isKeyless` in generated hpp is missing for keyed types, and returns **false** on non-key-types (it should be true).

ichernev:
This seems to fix the issue:

```patch
diff --git a/src/idlcxx/src/traits.c b/src/idlcxx/src/traits.c
index 7b1822e..27ecf7d 100644
--- a/src/idlcxx/src/traits.c
+++ b/src/idlcxx/src/traits.c
@@ -193,7 +193,7 @@ emit_traits(
       idl_fprintf(gen->header.handle, selfcontainedfmt, name) < 0)
     return IDL_RETCODE_NO_MEMORY;
 
-  if (idl_is_keyless(node, pstate->flags & IDL_FLAG_KEYLIST) &&
+  if (!idl_is_keyless(node, pstate->flags & IDL_FLAG_KEYLIST) &&
       idl_fprintf(gen->header.handle, keylessfmt, name) < 0)
     return IDL_RETCODE_NO_MEMORY;
```

If the struct is **not** keyless, then print `isKeyless() { return false; }`

Ah, those double negatives...

reicheratwork:
> Hello, I want to report an issue I recently discovered. The keyed/not-keyed property of the message is flipped. I.e if I have a `@key` field in the idl, the message is not-keyed, and if there is no `@key` property, the message _is_ keyed. This causes RTI and Cyclone based implementations to not match on topics (they used to match before this MR is applied). If I give one impl `@key` and the other no `@key`, it matches again :)
> 
> I can see that `isKeyless` in generated hpp is missing for keyed types, and returns **false** on non-key-types (it should be true).



> This seems to fix the issue:
> 
> ```diff
> diff --git a/src/idlcxx/src/traits.c b/src/idlcxx/src/traits.c
> index 7b1822e..27ecf7d 100644
> --- a/src/idlcxx/src/traits.c
> +++ b/src/idlcxx/src/traits.c
> @@ -193,7 +193,7 @@ emit_traits(
>        idl_fprintf(gen->header.handle, selfcontainedfmt, name) < 0)
>      return IDL_RETCODE_NO_MEMORY;
>  
> -  if (idl_is_keyless(node, pstate->flags & IDL_FLAG_KEYLIST) &&
> +  if (!idl_is_keyless(node, pstate->flags & IDL_FLAG_KEYLIST) &&
>        idl_fprintf(gen->header.handle, keylessfmt, name) < 0)
>      return IDL_RETCODE_NO_MEMORY;
> ```
> 
> If the struct is **not** keyless, then print `isKeyless() { return false; }`
> 
> Ah, those double negatives...

You are completely correct!

Somehow all the negation of negation of negation got tangled in my brain.

I will make a fix for this soon.

ichernev:
I finally managed to take a quick look at the entire code of this PR, and it looks good

eboasson:
@ichernev thanks for finding that one. Rather than the default extensibility, I'm thinking of two seemingly unlikely cases:
* A difference in interpretation of the correct XCDR2 serialisation of the type. The trouble is that I don't see how that could explain the `userId` field always being 0 or 1.
* An XCDR1 vs XCDR2 thing, because we switch to XCDR2 when there are optional fields in there to avoid the horrors of XCDR1 encoding of optional fields (see 7.4.1.2 in the XTypes spec, it won't take much time until you decide you really don't want to read any further). If RTI insists on deserialising it as XCDR1 even though we send XCDR2 (correctly advertised in discovery & encoding headers), perhaps that can be happen. It is not immediately obvious to me, but not inconceivable. That would be a really weird bug in their code.

Anyway, it would be really interesting to see what CDR RTI would generate.

eboasson:
@reicheratwork I also found some issues in the generated code. The first is caused by a variable inside a `case` that makes clang complain that C++ forbids a `goto` past an initializer. This I managed to hack my around with the patch at the end of this message (it is an ugly hack, but I figure it may be useful to have that at hand).

The other is that, having working around this first issue, the attached code generator+run script 
[cxx-cdrtest.zip](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/7908213/cxx-cdrtest.zip)
still errors out in compiling the output from the IDLC C++ backend with the following:
```
# bash ~/C/cxx-cdrtest.bash 3081669912 2>&1 | head -30
In file included from xxx-types.cpp:9:
./xxx-types.hpp:34783:39: error: unexpected namespace name 'x': expected expression
  auto prop = /*ZZ*/ get_type_props<::x>();
                                      ^
./xxx-types.hpp:34807:39: error: unexpected namespace name 'x': expected expression
  auto prop = /*ZZ*/ get_type_props<::x>();
                                      ^
./xxx-types.hpp:34832:39: error: unexpected namespace name 'x': expected expression
  auto prop = /*ZZ*/ get_type_props<::x>();
                                      ^
./xxx-types.hpp:34856:39: error: unexpected namespace name 'x': expected expression
  auto prop = /*ZZ*/ get_type_props<::x>();
                                      ^
./xxx-types.hpp:45882:12: error: redefinition of 'prop'
      auto prop = /*WW*/ get_type_props<::x::a00315>();
           ^
./xxx-types.hpp:45868:12: note: previous definition is here
      auto prop = /*WW*/ get_type_props<::x::a00319>();
           ^
```
Don't mind the `/*ZZ*/` and `/*WW*/`: I added those to the generator to more easily find where that particular bit of code was generated, in the hope that I could quickly and easily fix it. But no such luck, hence this comment rather than a PR.

Also note that you'll probably have to edit some paths in the bash script for it to work on your machine, but the idea should be clear. I'd be happy to supply the generated files if that's more convenient to you. Finally, note that there may well be issues in my code generator, but I think these particular errors are caused by some issues in the IDLC backend. I'm sure you'll agree no sane person would ever write IDL like this, but I believe it is valid IDL and so should result in valid C++ code.

```diff
diff --git a/src/idlcxx/src/streamers.c b/src/idlcxx/src/streamers.c
index 0d5a97f..c879961 100644
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -967,6 +967,7 @@ process_case(
       return IDL_RETCODE_NO_MEMORY;
 
     if (putf(&streams->read, read_start, type, value)
+     || multi_putf(streams, (WRITE | MOVE), "      {\n")
      || multi_putf(streams, ALL, get_props, type)
      || putf(&streams->max, max_start))
       return IDL_RETCODE_NO_MEMORY;
@@ -980,7 +981,7 @@ process_case(
     if (multi_putf(streams, ALL, check_props))
       return IDL_RETCODE_NO_MEMORY;
 
-    if (multi_putf(streams, (WRITE | MOVE), "      break;\n")
+    if (multi_putf(streams, (WRITE | MOVE), "      break; }\n")
      || putf(&streams->read, read_end, name)
      || putf(&streams->max, max_end))
       return IDL_RETCODE_NO_MEMORY;
```

reicheratwork:
> @reicheratwork I also found some issues in the generated code. The first is caused by a variable inside a `case` that makes clang complain that C++ forbids a `goto` past an initializer. This I managed to hack my around with the patch at the end of this message (it is an ugly hack, but I figure it may be useful to have that at hand).

Thanks!
I figured out what was going wrong: the entity properties for constructed types (the structures containing the descriptions of the member fields etc) are necessary when a struct is a union branch, and this was being retrieved at that point, inside the branch, breaking the rule on declaring variables inside the switch case

I added a fix for this bug.

reicheratwork:
> I want to report another issue, with the following IDL:
> 
> ```
> module HelloWorldData
> {
>   struct Msg
>   {
>     long userID;
>     @optional string message;
>   };
> };
> ```
> 
> I get correct communication between cyclone<->cyclone but if I receive in RTI (6.0.1) stuff is really messed up, the integer is either 0 or 1 and the string is always null, and if userId is >= 10 I get decerialization errors:
> 
> ```
> RTIXCdrInterpreter_deserializeString:HelloWorldData::Msg:message deserialization error. Received string length 1936942412 is larger than maximum 255
> ```
> 
> I haven't investigated much, but if the `@optional` is gone, it works fine. I can try to send some tcpdumps tomorrow (including how the cyclone packets look and how the packet looks if sent via RTI).
> 
> What is the default extensibility? RTI reports `extensible` as the default (at least the topics match...)
> 
> EDIT: If I stick `@final` before `struct Msg` it works. So evidently something about default extensibility is at fault here.

I would love to see the tcp dumps of the offending packets, so we can figure out what is going on here.

ichernev:
So I managed to do some testing on the OLD code (before the commits from 3 days ago), and I have some findings:

# Setup

The subscriber sends messages with alternating present/absent `@optional` message, the userID is one above the number in the message (don't ask :)).

# Final 

    module HelloWorldData
    {
      @final struct Msg
      {
        long userID;
        @optional string message;
      };
    };

## Cyclone

    0007 -- CDR2_LE
    3e   -- 62 # userID
    0000   00 07 00 00 3e 00 00 00 01 00 00 00 0b 00 00 00   ....>...........
    0010   4d 65 73 73 61 67 65 20 36 31 00 00               Message 61..

    0007 -- CDR2_LE
    17   -- 23 # userID

    0000   00 01 00 00 17 00 00 00 01 00 00 00               ............

    ! works cyclone->cyclone, cyclone->rti

## RTI

    0001 -- CDR_LE
    16   -- 22 # userID

    0000   00 01 00 00 16 00 00 00 01 00 10 00 0b 00 00 00   ................
    0010   4d 65 73 73 61 67 65 20 32 31 00 00               Message 21..

    0001 -- CDR_LE
    17   -- 23 # userID

    0000   00 01 00 00 17 00 00 00 01 00 00 00               ............

    ! works rti->rti, rti->cyclone


# Appendable

    module HelloWorldData
    {
      @appendable struct Msg
      {
        long userID;
        @optional string message;
      };
    };

NOTE: Appendable extensibility in RTI is called "extensible" extensibility. It
also looks like to be the default.

## Cyclone

    0009 -- D_CDR2_LE
    0e   -- 14 # userID

    0000   00 09 00 00 17 00 00 00 0e 00 00 00 01 00 00 00   ................
    0010   0b 00 00 00 4d 65 73 73 61 67 65 20 31 33 00 00   ....Message 13..

    0009 -- D_CDR2_LE
    0d   -- 13 # userID

    0000   00 09 00 01 05 00 00 00 0d 00 00 00 00 00 00 00   ................

    ! works cyclone->cyclone, cyclone->rti

## RTI

    0001 -- CDR_LE
    14   -- 20 # userID

    0000   00 01 00 00 14 00 00 00 01 00 10 00 0b 00 00 00   ................
    0010   4d 65 73 73 61 67 65 20 31 39 00 00               Message 19..

    0001 -- CDR_LE
    13   -- 19 # userID

    0000   00 01 00 00 13 00 00 00 01 00 00 00               ............

    ! works rti->rti, rti->cyclone

# Mutable

    module HelloWorldData
    {
      @mutable struct Msg
      {
        long userID;
        @optional string message;
      };
    };

## Cyclone

    000b -- PL_CDR2_LE
    14   -- 20 # userID

    0000   00 0b 00 00 23 00 00 00 00 00 00 40 04 00 00 00   ....#......@....
    0010   14 00 00 00 01 00 00 40 0f 00 00 00 0b 00 00 00   .......@........
    0020   4d 65 73 73 61 67 65 20 31 39 00 00               Message 19..

    1643364799.945386 [0]     recvUC: data(application, vendor 1.16): 110a546:a093c8ae:1c12db6:103 #116: deserialization HelloWorldData_Msg/HelloWorldData::Msg failed (for reasons unknown)
    !! fails cyclone->cyclone
    ! works cyclone->rti

    000b -- PL_CDR2_LE
    15   -- 21 # userID


    0000   00 0b 00 00 0c 00 00 00 00 00 00 40 04 00 00 00   ...........@....
    0010   15 00 00 00                                       ....

    ! works cyclone->cyclone

## RTI

    0003 -- PL_CDR_LE
    0000   00 03 00 00 00 00 04 00 43 00 00 00 02 7f 00 00   ........C.......

    ! works rti->rti, rti->cyclone

    0003 -- PL_CDR_LE
    0000   00 03 00 00 00 00 04 00 44 00 00 00 01 00 10 00   ........D.......
    0010   0b 00 00 00 4d 65 73 73 61 67 65 20 36 37 00 00   ....Message 67..
    0020   02 7f 00 00                                       ....

    ! works rti->rti
    ! crashes rti->cyclone
    subscriber: cyclonedds-cxx/src/ddscxx/src/org/eclipse/cyclonedds/core/cdr/cdr_stream.cpp:110: bool org::eclipse::cyclonedds::core::cdr::cdr_stream::bytes_available(size_t, bool): Assertion `m_buffer_end.size()' failed.


# No Annotation

    module HelloWorldData
    {
      struct Msg
      {
        long userID;
        @optional string message;
      };
    };

Cyclone assumes final (wrong):
P80 XTypes v1.3 7.3.1.2.1.8
> In the event that the representation of a given type does not indicate the
> type’s extensibility kind, the type shall be considered appendable.

RTI assumes appendable(extensible) -- correct

# TL;DR

* Cyclone defaults to `@final`, whereas it should default to `@appendable`.
* Cyclone -> cyclone fails with @mutable (if optional is present, if it's absent it works)
* Cyclone crashes for (PL_CDRv1) mutable reception (from RTI) if optional is present (if it's null it works) 

I will test latest code in a bit...

ichernev:
# Final

## Cyclone

    works    Cyclone->Cyclone
    no match Cyclone->RTI (CDR vs CRD2 type repr)

## RTI

    works    RTI->RTI
    no match RTI->Cyclone (CDR vs CDR2)

# Final + data\_representation(XCDR2) for RTI

    module HelloWorldData
    {
      @final @data_representation(XCDR2) struct Msg
      {
        long userID;
        @optional string message;
      };
    };


## Cyclone/RTI

    works Cyclone->Cyclone
    works Cyclone->RTI
    works RTI->RTI
    works RTI->Cyclone (CDR2)

NOTE: Cyclone `idlc` crashes with `@data_representation` annotation.

# Appendable : Cyclone/RTI

Again won't work across-vendor due to CDR/CDR2 but works in-vendor.

# Appendable + data\_repr(XCDR2) for RTI : Cyclone/RTI

Works across vendor.

# Mutable : Cyclone/RTI

    Cyclone->Cyclone fails if optional field is non-null
    Cyclone->RTI fails to match
    RTI->Cyclone fails to match
    RTI->RTI works

# Mutable + data\_repr(XCDR2) for RTI : Cyclone/RTI

    Cyclone->Cyclone fails if optional field is non-null !!
    Cyclone->RTI works
    RTI->Cyclone works
    RTI->RTI works


# TL;DR

* Cyclone still defauls to `@final` instead of `@appendable`
* Cyclone does not understand `@data_representation` annotation (idlc crashes)
* After recent changes, cyclone doesn't understand CDRv1 (or doesn't announce
  it -- that is a regression, it worked before)
* for `@mutable` Cyclone fails to talk to itself if `@optional` field is
  non-null (but in rare cases the first such message is correctly decoded)
  `deserialization HelloWorldData_Msg/HelloWorldData::Msg failed (for reasons
  unknown)`
* RTI doesn't wish to communicate with cyclone via CDRv2 unless
  `data_representation` is present in IDL, which seems to be against spec

# TODO

* check if QoS `data_representation` works for cyclone/rti
  * for cyclone to tell it to support v1
  * for RTI to tell it to support v2 (according to spec XTYPESv2 clients should
    announce CDRv2,

reicheratwork:
> # TL;DR
> * Cyclone still defauls to `@final` instead of `@appendable`
> * Cyclone does not understand `@data_representation` annotation (idlc crashes)
> * After recent changes, cyclone doesn't understand CDRv1 (or doesn't announce
>   it -- that is a regression, it worked before)
> * for `@mutable` Cyclone fails to talk to itself if `@optional` field is
>   non-null (but in rare cases the first such message is correctly decoded)
>   `deserialization HelloWorldData_Msg/HelloWorldData::Msg failed (for reasons unknown)`
> * RTI doesn't wish to communicate with cyclone via CDRv2 unless
>   `data_representation` is present in IDL, which seems to be against spec
> 
> # TODO
> * check if QoS `data_representation` works for cyclone/rti
>   
>   * for cyclone to tell it to support v1
>   * for RTI to tell it to support v2 (according to spec XTYPESv2 clients should
>     announce CDRv2,

I will have a look at these issues, though i think some of these are already known and in the pipeline 👍 .

eboasson:
Hi @ichernev, thanks a lot for all that testing! We are dependent on others to help us out for the interoperability testing, so it is really very much appreciated.

Regarding your observations, I think we can all agree that some are straight out bugs:
* IDLC crashing on a `@data_representation` annotation is a bug
* Cyclone->Cyclone deser fails if optional field is non-null
I'm sure @reicheratwork has already looked into it, as he wrote.

The `@final` / `@appendable` default extensibility is, however, a deliberate choice on our side. The issue is that with a default extensibility of appendable all nested structs will also be appendable and XCDRv1's inability to represent those correctly (*) is the reason XCDRv2 was introduced. That means:
* firstly, doing appendable with XCDRv1 is nonsensical and therefore forces the use of XCDRv2 for the vast majority of unannotated topics;
* secondly, and because of that, appendable-by-default means that if you take a pre-XTypes application and recompile half of it with an XTypes-capable implementation, the two won't work together. We think that's an unacceptable way of treating users. Hence our decision to default to `final` and adding a command-line option to change the default (https://github.com/eclipse-cyclonedds/cyclonedds/pull/989).

The backwards compatibility argument equally means that any type simple enough to be supported by a pre-XTypes implementation must be done in XCDRv1 by default. So that's also what we do, even though it is in violation of the "XTYPESv2 clients should announce CDRv2" in the spec. Again, the consideration is that we want to help users, not make life harder for them.

Anything else, we do XCDRv2. We have some coverage for XCDRv1 `appendable`/`mutable` representation through happenstance, but we have no desire to support that and so we reject XCDRv1 for anything but the simplest cases. Supporting it is not difficult, but it does mean yet more testing, more code to maintain, a bigger attack surface. And for what? Really only interoperability with some old RTI versions from the days of XTypes 1.2 and older. We can see some scenarios in which it could become a problem (e.g., old subsystems that cannot be changed anymore). If we run into such a situation we'll reconsider.

(*) You're fine as long as you don't append something to a nested struct; once you do, the assignability check will pass and the deserialisation will produce nonsense.

ichernev:
@eboasson any notes about:
> After recent changes, cyclone doesn't understand CDRv1 (or doesn't announce it -- that is a regression, it worked before)

So even if rti/cyclone structs are marked as `@final` RTI tries to send CDRv1 (the simple type), but cyclone insists on v2 for some reason. And it wasn't this way like, last week :)

The rest, I mostly agree with you, I'll have to check with my team to see which direction to take on our side.

ichernev:
If Cyclone does not handle `data_representation` what is the right way to force v1/v2 "on the wire"? Can two versions be proposed by the sender? It would make sense to try v2 and if that fails fall back to v1. Also I don't think we've done anything in that regards with RTI, but it does use v1 always (not very old versions only, we're on 6.0.1, which is pretty recent).

Ideally I won't have to modify all my IDLs to put `data_representation` (just for RTI, assuming Cyclone doesn't crash), or put `@final` everywhere (again, just for compatibility). We're ATM trying to have the exact same IDL files for both RTI/Cyclone.

eboasson:
> @eboasson any notes about:
> 
> > After recent changes, cyclone doesn't understand CDRv1 (or doesn't announce it -- that is a regression, it worked before)
> 
> So even if rti/cyclone structs are marked as `@final` RTI tries to send CDRv1 (the simple type), but cyclone insists on v2 for some reason. And it wasn't this way like, last week :)
> 
> The rest, I mostly agree with you, I'll have to check with my team to see which direction to take on our side.

I was surprised to see that there was a change in Cyclone, so I tried not to say anything about it until I understand what exactly is going on there. My attempt at avoiding the question seems to not have worked ...

But for now, "need to check and find out more" is the answer.

eboasson:
> If Cyclone does not handle `data_representation` what is the right way to force v1/v2 "on the wire"?

Cyclone does support the data representation QoS setting (the annotation will come as well). You can always force v2, v1 is dependent on whether the type is "complicated" or not.

> Can two versions be proposed by the sender? It would make sense to try v2 and if that fails fall back to v1. Also I don't think we've done anything in that regards with RTI, but it does use v1 always (not very old versions only, we're on 6.0.1, which is pretty recent).

The way I understand the XTypes spec, the writer really just proposes one. This kind of makes sense, otherwise you could easily end up in a situation where you have to publish multiple copies in different representations, simply because someone adds a reader. It could be that my understanding is wrong (that's rather easy with this spec), but in that case, at least for now, Cyclone requires the writer to choose the representation once.

> Ideally I won't have to modify all my IDLs to put `data_representation` (just for RTI, assuming Cyclone doesn't crash), or put `@final` everywhere (again, just for compatibility). We're ATM trying to have the exact same IDL files for both RTI/Cyclone.

I think the data representation QoS setting should work in both, and the annotation is just a matter of time. https://github.com/eclipse-cyclonedds/cyclonedds/pull/989 should allow you to use appendable everywhere by default. I haven't gotten around to reviewing the latest update of it, but I'll try to do so without delay.

Once that's merged, it'll be straightforward to have everything appendable in both, and by setting the data representation QoS in both to v2 they should also agree on the using XCDRv2 throughout.

reicheratwork:
> @eboasson any notes about:
> 
> > After recent changes, cyclone doesn't understand CDRv1 (or doesn't announce it -- that is a regression, it worked before)
> 
> So even if rti/cyclone structs are marked as `@final` RTI tries to send CDRv1 (the simple type), but cyclone insists on v2 for some reason. And it wasn't this way like, last week :)
> 
> The rest, I mostly agree with you, I'll have to check with my team to see which direction to take on our side.

I think I get where the confusion is coming from:
- the struct you are attempting to send is final
- but it contains an optional field, so it requires XTypes support, either v1 or v2, as optional field are not a part of the basic CDR spec
- as @eboasson said, the choice was made to support the v2 XTypes and not the v1 XTypes
That is why it requires v2 functionality, even though the struct is annotated as being final

reicheratwork:
@ichernev
I did some autopsies on the packets that you posted, and most seem to follow the correct encoding and formatting for the types they claim to represent. The only weird one out is the following:
```
Cyclone
0007 -- CDR2_LE
3e   -- 62 # userID
0000   00 07 00 00 3e 00 00 00 01 00 00 00 0b 00 00 00   ....>...........
0010   4d 65 73 73 61 67 65 20 36 31 00 00               Message 61..

0007 -- CDR2_LE
17   -- 23 # userID

0000   00 01 00 00 17 00 00 00 01 00 00 00               ............

! works cyclone->cyclone, cyclone->rti
```
for the type:
```
module HelloWorldData
{
  @final struct Msg
  {
    long userID;
    @optional string message;
  };
};
```
In my experiments, Cyclonedds-CXX always says it is encoded as PLAIN-CDR2, like in the first packet `00 07 00 00` (PLAIN_CDR2, little-endian with 0 bytes missing on the end to align to 4), never as PLAIN-CDR which the second packet says (`00 01 00 00` : PLAIN_CDR, little-endian with 0 bytes missing on the end to align to 4).
If it is produced by RTI and encoded as PLAIN-CDR (XCDRv1) however, with the last 4 bytes being `01 00 00 00`, then it makes perfect sense. Member id = 1 (message string) and length = 0 (optional not present).

The issues with RTI and Cyclone not "finding" each other may have something to do with the discovery process and datarepresentation QoS policies being exchanged there.

I cannot reproduce some of your reported issues regarding the optionals and mutables, so maybe you can dump some more "incorrect" CDRs, that I may go over.

ichernev:
> In my experiments, Cyclonedds-CXX always says it is encoded as PLAIN-CDR2, like in the first packet 00 07 00 00 (PLAIN_CDR2, little-endian with 0 bytes missing on the end to align to 4), never as PLAIN-CDR which the second packet says (00 01 00 00 : PLAIN_CDR, little-endian with 0 bytes missing on the end to align to 4).
If it is produced by RTI and encoded as PLAIN-CDR (XCDRv1) however, with the last 4 bytes being 01 00 00 00, then it makes perfect sense. Member id = 1 (message string) and length = 0 (optional not present).

Well, copying by hand from WireShark is not the least error-prone exercise. I'll redo the tests (now that you've changed the code again), and try to make it reproducible.

reicheratwork:
I did a force push to update this branch to get it to work with the latest changes in CycloneDDS, I also fixed a number of issues revealed by the efforts of @eboasson , @ichernev (many thanks!) and myself

There will still be communication issues between CycloneDDS and CycloneDDS-CXX (mainly writing from C to CXX) until [PR#1132](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1132) is merged into CycloneDDS

ichernev:
> There will still be communication issues between CycloneDDS and CycloneDDS-CXX (mainly writing from C to CXX) until https://github.com/eclipse-cyclonedds/cyclonedds/pull/1132 is merged into CycloneDDS

Thankfully I don't have a plan to test THAT :)

> check if QoS data_representation works for cyclone/rti 

Ah, the beauty:

```cpp
        auto writer_qos = publisher.default_datawriter_qos();
#if defined(DDS_IMPL_RTI)
        auto val = dds::core::policy::DataRepresentation::xcdr2();
#elif defined(DDS_IMPL_CYCLONE)
        auto val = dds::core::policy::DataRepresentationId::XCDR2;
#endif
        writer_qos.policy(dds::core::policy::DataRepresentation({val}));
```

If you try `writer_qos.policy({val})` like cyclone tests do you get into a weird linker error and a long discussion with an RTI rep, followed by a lot of grepping in both codebases and figuring out stuff on your own...

Now all my tests (struct with different extensibility types) work across vendors and within-vendor (the `@mutable` issue was apparently resolved). Also there is an `idlc` option for default extensibility, which is nice.

eboasson:
> > There will still be communication issues between CycloneDDS and CycloneDDS-CXX (mainly writing from C to CXX) until [eclipse-cyclonedds/cyclonedds#1132](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1132) is merged into CycloneDDS
> 
> Thankfully I don't have a plan to test THAT :)

And I do have a plan for that (using a perl-based monstrosity) 🙂

> > check if QoS data_representation works for cyclone/rti
> 
> Ah, the beauty:
> 
> ```c++
>         auto writer_qos = publisher.default_datawriter_qos();
> #if defined(DDS_IMPL_RTI)
>         auto val = dds::core::policy::DataRepresentation::xcdr2();
> #elif defined(DDS_IMPL_CYCLONE)
>         auto val = dds::core::policy::DataRepresentationId::XCDR2;
> #endif
>         writer_qos.policy(dds::core::policy::DataRepresentation({val}));
> ```

That's an interesting difference. The XTypes spec has:
image
which Cyclone matches on the `DataRepresentationId`. But the `XCDR2` is also interesting, because the C++ API isn't a simple mechanical translation from the IDL and has differences with the various other policies. I would assume the data representation id is supposed to be mapped similarly to the ownership kind, which in the DCPS spec is defined as:
image
and in the C++ PSM ends up as:
image
To me that suggests it is intended to be a method (like in RTI's version) but I don't think it should be all lower-case. Probably it should be `Xcdr2()` to follow the pattern. Ah, the joys of underspecification in ambiguous and self-contradictory specs!

I don't think we'll soon see eye-to-eye with RTI on this. But this is certainly the time to double-check what it is supposed to be. @e-hndrks, any suggestions?

> If you try `writer_qos.policy({val})` like cyclone tests do you get into a weird linker error and a long discussion with an RTI rep, followed by a lot of grepping in both codebases and figuring out stuff on your own...

At least with Cyclone you can always do a PR to try to change it to your liking 😁

> Now all my tests (struct with different extensibility types) work across vendors and within-vendor (the `@mutable` issue was apparently resolved). Also there is an `idlc` option for default extensibility, which is nice. 🎉

Great!

ichernev:
> At least with Cyclone you can always do a PR to try to change it to your liking grin

From what I figured, the policy lets you set an **array** of data representations. Which is different than most other policies where you set a single value. I didn't have the nerve to read the docs, but isn't the array supposed to be multiple supported options, choose the first that matches the other side?

In any case, for some reason cyclone accepts `{val}` (so does it get wrapped in an array and then the array into a DataRepresentation?), and RTI needs at least `DataRepresentation` explicitly specified (I used to have `DataRepresentation(DataRepresentationIdSeq({val}))` but I figured I can drop some.

This almost works:
```cpp
        auto writer_qos = publisher.default_datawriter_qos();
#if defined(DDS_IMPL_RTI)
        auto val = DDS_XCDR2_DATA_REPRESENTATION;
#elif defined(DDS_IMPL_CYCLONE)
        auto val = DDS_DATA_REPRESENTATION_XCDR2;
#endif
        writer_qos.policy(dds::core::policy::DataRepresentation({val}));
```

But I get this `error: no matching function for call to ‘dds::core::safe_enum<dds::core::policy::DataRepresentationId_def>::safe_enum()’
  109 |     { ::new(static_cast<void*>(__p)) _Tp(std::forward<_Args>(__args)...); }`. It would be nice if this smart enum can handle integers in constructors ... but oh safety. Also, as you can see the constant names are different. In RTI the DRId type is a plain short, for cyclone it's a safe_enum. Not sure if that is very spec-y.

BTW enum handling of idl-generated types is also different, but that is a subject for another issue.

reicheratwork:
> From what I figured, the policy lets you set an **array** of data representations. Which is different than most other policies where you set a single value. I didn't have the nerve to read the docs, but isn't the array supposed to be multiple supported options, choose the first that matches the other side?

As far as I understand it, the datarepresentation qos structure is the same between topics, readers, writers etc., which is a sequence of values (DataRepresentationIdSeq) (DDS XTypes spec v1.3 pp. 312-315).
For the reader it represents the types it ACCEPTS, whereas for the writer, it means the DataRepresentation it MUST use when writing, so just one is used, in the spec it says to take the first entry from the sequence, and ignore the rest (spec pp. 211), quote:
_(Because the policy structure includes a sequence, it is technically possible for the writer to offer more than one representation. Implementers of this specification may use this fact in order to offer extended functionality; however, this specification does not specify any meaning for the representation identifiers after the first, and implementations may ignore them.)_

ichernev:
>  For the reader it represents the types it ACCEPTS, whereas for the writer, it means the DataRepresentation it MUST use when writing, so just one is used, in the spec it says to take the first entry from the sequence, and ignore the rest (spec pp. 211), quote:

Thanks for the quote.

BTW, if Cyclone has some extended QoS test cases I can try to run them under RTI to see if there are other mismatches.

eboasson (approved):
Last major xtypes hurdle taken! 🎉 Thank you for, @reicheratwork !

PR #182, opened 2021-12-01
Added some changes to the idlcxx/CMakeLists.txt and to inttypes.h to …

zephod77:
…allow

older systems, ie pre-gcc5, to build.

Removed the obsolete limitation of a maximum of 230 for the domain_id. Now it can
be any uint32_t value.

NOTE: these changes have only been tested on CentOS7.1 and Ubuntu 20.04. They should be tested on systems with other  tool sets. Also the test suite has not been run because GoogleTest required GCC5 or greater and the CentOS7.1 system I have uses GCC4.8.

### DIFF
--- a/src/ddscxx/include/dds/core/detail/inttypes.hpp
+++ b/src/ddscxx/include/dds/core/detail/inttypes.hpp
@@ -27,7 +27,7 @@
 #define __STDC_FORMAT_MACROS
 #endif
 
-//#include <stdint.h>
+#include <stdint.h>
 #include <inttypes.h>
 
 // End of implementation
--- a/src/ddscxx/include/dds/core/detail/inttypes.hpp
+++ b/src/ddscxx/include/dds/core/detail/inttypes.hpp
@@ -23,7 +23,12 @@
  * simply include that one. Under toolchains that do not, this header must
  * provide equivalent definitions.
  */
-#include <stdint.h>
+#if defined(__GNUC__) && __GNUC__ < 5
+#define __STDC_FORMAT_MACROS
+#endif
+
+//#include <stdint.h>
+#include <inttypes.h>
 
 // End of implementation
 
--- a/src/idlcxx/CMakeLists.txt
+++ b/src/idlcxx/CMakeLists.txt
@@ -22,7 +22,7 @@ set_target_properties(idlcxx PROPERTIES
    OUTPUT_NAME "cycloneddsidlcxx"
    VERSION ${PROJECT_VERSION}
    SOVERSION ${PROJECT_VERSION_MAJOR}
-   C_STANDARD 11)
+   C_STANDARD 99)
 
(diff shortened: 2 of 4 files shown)


src/ddscxx/include/dds/core/detail/inttypes.hpp
```diff
@@ -23,7 +23,12 @@
  * simply include that one. Under toolchains that do not, this header must
  * provide equivalent definitions.
  */
-#include <stdint.h>
+#if defined(__GNUC__) && __GNUC__ < 5
+#define __STDC_FORMAT_MACROS
+#endif
+
+//#include <stdint.h>
```
eboasson, line 30: I am not sure whether `inttypes.h` actually implies `stdint.h`. It seems no more than reasonable, but including both seems a practical of avoiding any problems.
```suggestion
#include <stdint.h>
```

src/idlcxx/CMakeLists.txt
```diff
@@ -21,7 +21,8 @@ add_library(
 set_target_properties(idlcxx PROPERTIES
    OUTPUT_NAME "cycloneddsidlcxx"
    VERSION ${PROJECT_VERSION}
-   SOVERSION ${PROJECT_VERSION_MAJOR})
+   SOVERSION ${PROJECT_VERSION_MAJOR}
+   C_STANDARD 11)
```
eboasson, line 25: `C_STANDARD 11` is going to be a problem, there are too many platforms that don't do C11 yet (it is only a decade old ...), but the code is C99 so this
```suggestion
   C_STANDARD 99)
```
instead would do the trick. Even Microsoft has caught up with the new millennium by now.

eboasson (commented):
LGTM, just two minor points where I think problems might arise, both trivially fixable.

eboasson:
@zephod77 the code is fine now, but the Eclipse ECA check fails because you have used two email addresses and only one is registered with the Eclipse Foundation. Sorry that I have to be picky about that, but I have no choice ...

This is what it says:

image

Feel free to "just" [change the author](https://github.community/t/how-to-mark-a-past-commit-as-belonging-to-another-user-why-does-claim-email-address-not-show-for-others/135506/2) and force-push the branch. This is small enough that that won't pose any difficulty.

zephod77:
Done on one commit but I might need to do it again on other commits.

eboasson:
> Done on one commit but I might need to do it again on other commits.

Yes, it still complaining about some of the commits ...

Come to think of it, given that there are only a few commits, what might be easier than editing them commits is to create a new branch, and then do `git cherry-pick X && git commit --reset-author --amend` and then reset the PR branch. But you may well know git better than I do 🤔🙂

eboasson:
@zephod77 I must have overlooked a message from GitHub that you pushed something because it all checks out now. Thanks for the fixes and for your patience with the ECA check!

PR #183, opened 2021-12-07
Add option to skip IDL module building

ichernev:
Similar to cyclonedds (the C project) do not build the IDL compiler
module in case of cross-compiling. In that case the native IDL compiler
(build separately) should be used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,15 @@ if(CMAKE_VERSION VERSION_LESS 3.13)
   endmacro()
 endif()
 
+if(CMAKE_CROSSCOMPILING)
+  set(not_crosscompiling OFF)
+else()
+  set(not_crosscompiling ON)
+endif()
+
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/Modules")
+
+option(BUILD_IDLLIB "Build IDL preprocessor lib" ${not_crosscompiling})
 
 # Make it easy to enable MSVC, Clang's/gcc's analyzers
 set(ANALYZER "" CACHE STRING "Analyzer to enable on the build.")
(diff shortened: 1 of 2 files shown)


k0ekk0ek:
Thanks @ichernev! Definitely in favor of merging this, but it would be great if you can rename the option to `BUILD_IDLCXX` though. I think we'd also need to update `idlcxx_generate` to use it correctly. `cmake/Modules/Generate.cmake` in the *cyclonedds* repository is a good resource there (commit `a614bdebe0626326bad2d40e84adba937ec42a8f`). They're pretty similar. Any chance you're willing to include those changes as well?

ichernev:
> Any chance you're willing to include those changes as well?

I'm willing to help but no earlier than mid-next-week.

k0ekk0ek:
@ichernev, that's awesome! Don't feel obligated though. If you can't find time, I'll be happy add the missing bits. Thanks again for the PR!

ichernev:
@k0ekk0ek there are a few PRs that do similar stuff, have you made up your mind what is the right way forward? Should there be a Generate.cmake installed along the cross-compiled library so the "package" can use idlc (on native) platform with cmake?

k0ekk0ek:
Hi @ichernev. If it'd be me, I'd accept this PR request, but other interested parties have a definite preference to combine all of it. If you don't mind, I prefer they handle this.

eboasson:
@ichernev this was on my list of PRs that were "magically" being take care of without my paying attention to it ... I suppose the world changed and it is now up to me to decide. My apologies for not paying more attention to the changes in the world.

Better than late than never. This is simple enough to add without causing grief for other PRs (that have been lingering for far too long as well); maintaining a `BUILD_IDLLIB` seems perfectly fine. Merging it. Thanks!

PR #184, opened 2021-12-07
Include <limits> for std::numeric_limits<T>

ichernev:
Maybe this worked on older GCCs but on newer ones (11.1.0) it won't.


k0ekk0ek:
Yep, you're right. Getting the same issue on my Fedora workstation :+1: @eboasson, any chance you can hit merge?

Issue #170, opened 2021-10-27
IDLC: segfault in handling of typedef


TumbleOwlee:
Something goes wrong when generating an idl file with any `typedef`. A simple example is given here:
```c
#ifndef DATA_TYPES
#define DATA_TYPES
module some_module
{
    struct SomeIdentifier
    {
      long id;
      long type;
    };

    typedef SomeIdentifier SomeIdentifier_t;

    struct ModeCommand_T
    {
      SomeIdentifier_t rId;
      SomeIdentifier_t sId;
      long num;
    };

#pragma keylist ModeCommand_T rId.id rId.type
};
#endif
```
I skimmed the code and checked the backtrace and it looks like the `typedef` can't be resolved. The backtrace of the segfault looks like this:
```bash
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7bac859 in __GI_abort () at abort.c:79
#2  0x00007ffff7bac729 in __assert_fail_base (fmt=0x7ffff7d42588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x7ffff795db67 "0", 
    file=0x7ffff795d120 ".../source_subfolder/src/idlcxx/src/streamers.c", line=895, function=<optimized out>) at assert.c:92
#3  0x00007ffff7bbdf36 in __GI___assert_fail (assertion=0x7ffff795db67 "0", 
    file=0x7ffff795d120 ".../source_subfolder/src/idlcxx/src/streamers.c", line=895, function=0x7ffff795e428 <__PRETTY_FUNCTION__.4022> "process_key") at assert.c:101
#4  0x00007ffff795811e in process_key (pstate=0x5555557f4c30, streams=0x7fffffffdb60, _struct=0x5555558001f0, key=0x555555801100)
    at .../source_subfolder/src/idlcxx/src/streamers.c:895
#5  0x00007ffff79582f0 in process_keylist (pstate=0x5555557f4c30, streams=0x7fffffffdb60, _struct=0x5555558001f0)
    at .../source_subfolder/src/idlcxx/src/streamers.c:940
#6  0x00007ffff7958930 in process_struct (pstate=0x5555557f4c30, revisit=false, path=0x7fffffffdb10, node=0x5555558001f0, user_data=0x7fffffffdb60)
    at .../source_subfolder/src/idlcxx/src/streamers.c:1068
[...]

PR #186, opened 2021-12-09
Fix keylist processing for typedeffed topics

### LINKED ISSUE #157: Forbidding dynamic allocation of reference types
Just a small question: [this comment in the Reference class](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/b1895b56b2ac07c4814ae965e49bdb2f15191e30/src/ddscxx/include/dds/core/Reference.hpp#L261) says that reference types shouldn't be dynamically allocated. It is however still possible to create shared ptrs from them, which use placement new. Is this intentional? 

Out of curiosity: Why should this type not be dynamically allocated anyway?

k0ekk0ek:
(Originally incorrectly referenced #157 which GitHub now desperately wants to close on merging this one.)

### DIFF
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -861,6 +861,8 @@ process_inherit_spec(
 static const idl_declarator_t*
 resolve_member(const idl_struct_t *type_spec, const char *member_name)
 {
+  type_spec = idl_unalias(type_spec, 0u);
+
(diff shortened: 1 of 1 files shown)


k0ekk0ek:
Correction, this fixes #170 :sweat_smile:

eboasson:
Merging because it fixes a blocking issue for someone, but I adding a testcase would've been a good idea.

PR #187, opened 2021-12-09
Remove redundant template parameter

### LINKED ISSUE #181: client compilation failed under official gcc 11.2.0 container with strict c++20 standard enabled
### Infos: 
- host : official [GCC 11.2.0](https://hub.docker.com/_/gcc/) devcontainer
- libs : 
    - Eclipse Iceoryx v1.0.1
    - Eclipse CycloneDDS v0.8.1
    - Eclipse CycloneDDS-CXX v0.8.1
 - misc libs dependencies : git, cmake, bison, openssl, ... 

### HowTo:

1. clone the repo & switch to tags/0.8.1
2. update the _[cyclonedds-cxx/exemple/helloworld/CMakeLists.txt:37,38](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/120c34915a129c1189f6834b967616c248355b8f/examples/helloworld/CMakeLists.txt#L37)_ to enforce strict c++20 as follow :
```cmake 
set_property(TARGET ddscxxHelloworldPublisher PROPERTY CXX_STANDARD 20 CMAKE_CXX_EXTENSIONS OFF CMAKE_CXX_STANDARD_REQUIRED ON)
set_property(TARGET ddscxxHelloworldSubscriber PROPERTY CXX_STANDARD 20 CMAKE_CXX_EXTENSIONS OFF CMAKE_CXX_STANDARD_REQUIRED ON)
```
3. under the build dir, run
```shell 
cmake -DCMAKE_CXX_FLAGS=-std=c++20 -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_EXAMPLES=ON .. && cmake --build .
```
4. then, you should see these errors (x4)
> cyclonedds-cxx/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp:63:33: error: template-id not allowed for destructor
>    63 | dds::core::Reference<DELEGATE>::~Reference<DELEGATE>()
> [...]
> cyclonedds-cxx/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp:211:31: error: template-id not allowed for destructor
>   211 | dds::topic::detail::Topic<T>::~Topic<T>()
> [...]
> cyclonedds-cxx/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp:514:34: error: template-id not allowed for destructor
>   514 | dds::sub::detail::DataReader<T>::~DataReader<T>()
> [...]
> cyclonedds-cxx/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp:403:34: error: template-id not allowed for destructor
>   403 | dds::pub::detail::DataWriter<T>::~DataWriter<T>()

### Reason to do so: 

Our client library who use cyclonedds-cxx is dependent of some c++20 features 

### Solving: 

Following [this solved issue](https://wg21.cmeerw.net/cwg/issue2237) and especially the _clause 15_, it lead me to the correction of each template-id destructor definition in the previous files by commenting them as follow : 
1. _[cyclonedds-cxx/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp:403](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/120c34915a129c1189f6834b967616c248355b8f/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp#L403)_
```c++
template <typename T>    
dds::pub::detail::DataWriter<T>::~DataWriter/*<T>*/()   
{
    /* some code eluded */
}
```
2. _[cyclonedds-cxx/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp:514](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/120c34915a129c1189f6834b967616c248355b8f/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp#L514)_
```c++
template <typename T>
dds::pub::detail::DataReader<T>::~DataReader/*<T>*/()
{
    /* some code eluded */
}
```
3. _[cyclonedds-cxx/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp:211](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/120c34915a129c1189f6834b967616c248355b8f/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp#L211)_
```c++
template <typename T>
dds::pub::detail::Topic<T>::~Topic/*<T>*/()
{
    /* some code eluded */
}
```
4. _[cyclonedds-cxx/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp:63](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/120c34915a129c1189f6834b967616c248355b8f/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp#L63)_
```c++
template <typename DELEGATE>
dds::pub::detail::Reference<DELEGATE>::~Reference/*<DELEGATE>*/()
{
}
```
Doing so will result in a correct compilation & execution.

### Discussion 

Is there any reason to keep these template-id definitions **_even without the strict c++20 support ?_** Could you explicit their purpose or need ? 

Thanks !

k0ekk0ek:
Fixes #181. Thanks @nkinani!

### DIFF
--- a/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp
+++ b/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp
@@ -60,7 +60,11 @@ dds::core::Reference<DELEGATE>::Reference(const DELEGATE_REF_T& p) : impl_(p)
 }
 
 template <typename DELEGATE>
+#if __cplusplus >= 201402L && (__cplusplus <= 201703L || __clang__)
 dds::core::Reference<DELEGATE>::~Reference<DELEGATE>()
+#else
+dds::core::Reference<DELEGATE>::~Reference()
+#endif
 {
 }
 
--- a/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp
+++ b/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp
@@ -400,7 +400,11 @@ dds::pub::detail::DataWriter<T>::DataWriter(
 }
 
 template <typename T>
+#if __cplusplus >= 201402L && (__cplusplus < 201703L || __clang__)
 dds::pub::detail::DataWriter<T>::~DataWriter<T>()
+#else
+dds::pub::detail::DataWriter<T>::~DataWriter()
+#endif
 {
     if (!this->closed) {
         try {
--- a/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
+++ b/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
@@ -511,7 +511,11 @@ dds::sub::detail::DataReader<T>::common_constructor(
 }
 
 template <typename T>
+#if __cplusplus >= 201402L && (__cplusplus <= 201703L || __clang__)
 dds::sub::detail::DataReader<T>::~DataReader<T>()
+#else
+dds::sub::detail::DataReader<T>::~DataReader()
+#endif
 {
     if (!this->closed) {
         try {
--- a/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp
+++ b/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp
@@ -206,9 +206,12 @@ dds::topic::detail::Topic<T>::Topic(const dds::domain::DomainParticipant& dp,
     this->AnyTopicDelegate::set_sample(&this->sample_);
 }
 
-
 template <typename T>
+#if __cplusplus >= 201402L && (__cplusplus <= 201703L || __clang__)
 dds::topic::detail::Topic<T>::~Topic<T>()
+#else
+dds::topic::detail::Topic<T>::~Topic()
+#endif
 {
     if (!closed) {
         try {


eboasson:
This reminds me of a change, and indeed: 39ffc05dd419ecd661c8f4d090289b4fc6ed9f5f. My local Apple clang (version 13.0.0) insists it was right and this PR makes it wrong (see below). I'm sure there's a good explanation why this warning doesn't show up in CI macOS build, but I don't know it yet. @e-hndrks, @sumanth-nirmal, @reicheratwork anyone who knows C++ well enough to say with certainty that one is right and the other is wrong?

If the code is correct with the proposed change, I have no difficulty with locally disabling a warning. But I don't want to go through a cyclone of adding and removing the same template parameter twice a year.

```
[66/94] Building CXX object src/ddscxx/tests/CMakeFiles/ddscxx_tests.dir/Exception.cpp.o
In file included from /Users/erik/C/cxx/src/ddscxx/tests/Exception.cpp:12:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/dds.hpp:33:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/core/ddscore.hpp:28:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/core/Entity.hpp:22:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/core/detail/Entity.hpp:22:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/core/detail/TEntityImpl.hpp:22:
/Users/erik/C/cxx/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp:63:31: warning: ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [-Wdtor-name]
dds::core::Reference<DELEGATE>::~Reference()
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
                              ::Reference
In file included from /Users/erik/C/cxx/src/ddscxx/tests/Exception.cpp:12:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/dds.hpp:35:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/topic/ddstopic.hpp:25:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/topic/ContentFilteredTopic.hpp:22:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp:31:
In file included from /Users/erik/C/cxx/src/ddscxx/include/org/eclipse/cyclonedds/sub/AnyDataReaderDelegate.hpp:36:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/topic/BuiltinTopic.hpp:22:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/topic/detail/BuiltinTopic.hpp:23:
/Users/erik/C/cxx/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp:211:29: warning: ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [-Wdtor-name]
dds::topic::detail::Topic<T>::~Topic()
~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
                            ::Topic
In file included from /Users/erik/C/cxx/src/ddscxx/tests/Exception.cpp:12:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/dds.hpp:37:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/sub/ddssub.hpp:125:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/sub/detail/ddssub.hpp:15:
/Users/erik/C/cxx/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp:514:32: warning: ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [-Wdtor-name]
dds::sub::detail::DataReader<T>::~DataReader()
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
                               ::DataReader
In file included from /Users/erik/C/cxx/src/ddscxx/tests/Exception.cpp:12:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/dds.hpp:38:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/pub/ddspub.hpp:30:
In file included from /Users/erik/C/cxx/src/ddscxx/include/dds/pub/detail/ddspub.hpp:15:
/Users/erik/C/cxx/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp:403:32: warning: ISO C++ requires the name after '::~' to be found in the same scope as the name before '::~' [-Wdtor-name]
dds::pub::detail::DataWriter<T>::~DataWriter()
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
                               ::DataWriter
4 warnings generated.
```

k0ekk0ek:
I suspected something like this would happen. A `#ifdef` if `-std=c++20` or something would work I guess(?) @eboasson, since it's relatively easy for you, can you test what happens if you set `set_property(TARGET ddscxxHelloworldSubscriber PROPERTY CXX_STANDARD 20)` in the helloworld example?

eboasson:
Same ...

k0ekk0ek:
In that case I'm kind of leaning towards an `#ifdef` in case of `__clang__`. I know it's not pretty, but in think GCC is actually right? I'll have another look to determine the which compiler is right and make the appropriate changes. Thanks for giving it another run.

k0ekk0ek:
@eboasson, I've updated the PR to use the older signature in case of `__cplusplus <= 201703L || __clang__`. The check can probably be adjusted to `__cplusplus < 20xxxxL` after a while. I suspect we'll know exactly when that is once clang starts throwing warnings/errors again :slightly_smiling_face: If any of the aforementioned people (@e-hndrks, @reicheratwork, @sumanth-nirmal) can pitch in, that'd be appreciated.

reicheratwork:
> This reminds me of a change, and indeed: [39ffc05](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/39ffc05dd419ecd661c8f4d090289b4fc6ed9f5f). My local Apple clang (version 13.0.0) insists it was right and this PR makes it wrong (see below). I'm sure there's a good explanation why this warning doesn't show up in CI macOS build, but I don't know it yet. @e-hndrks, @sumanth-nirmal, @reicheratwork anyone who knows C++ well enough to say with certainty that one is right and the other is wrong?
> ...

It seems that CLang is correct in this case and other people in the LLVM project have run into the same issue as us: [https://bugs.llvm.org/show_bug.cgi?id=46979#c1](url). It looks like a difference in the way qualified name lookup is done between destructors and other functions.

While CLang is correct to the letter of the spec, I think the spec is wrong in the spirit of this case as it results in "ugly" code for destructors which require duplicate template parameters.
As there is a specific compiler switch to suppress this error explicitly: `-Wdtor-name` it seems that the people of CLang have expected this issue, and I suggest using this switch to suppress this error.

eboasson:
Thanks @reicheratwork for figuring out what exactly is going on. @k0ekk0ek can you change the PR once again, please?Disabling a nuisance warning is better than adding ifdefs. The following change to this PR does the trick for me locally:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 5e97ea1..8bf806c 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,8 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR
   add_compile_options(
     -Wall -Wextra -Wconversion -Wstrict-prototypes
     -Wunused -Winfinite-recursion -Wassign-enum -Wcomma -Wdocumentation
-    -Wconditional-uninitialized -Wshadow -Wsign-conversion -Wpedantic)
+    -Wconditional-uninitialized -Wshadow -Wsign-conversion -Wpedantic
+    -Wno-dtor-name)
   add_compile_options("$<$<STREQUAL:$<TARGET_PROPERTY:LINKER_LANGUAGE>,CXX>:-Wold-style-cast>")
   if(WERROR)
     add_compile_options(-Werror)
diff --git a/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp b/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp
index 92892de..081341e 100644
--- a/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp
+++ b/src/ddscxx/include/dds/core/detail/ReferenceImpl.hpp
@@ -60,11 +60,7 @@ dds::core::Reference<DELEGATE>::Reference(const DELEGATE_REF_T& p) : impl_(p)
 }

 template <typename DELEGATE>
-#if __cplusplus <= 201703L || __clang__
-dds::core::Reference<DELEGATE>::~Reference<DELEGATE>()
-#else // C++20 (CWG 2237)
 dds::core::Reference<DELEGATE>::~Reference()
-#endif
 {
 }

diff --git a/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp b/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp
index 92ff492..4707b61 100644
--- a/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp
+++ b/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp
@@ -400,11 +400,7 @@ dds::pub::detail::DataWriter<T>::DataWriter(
 }

 template <typename T>
-#if __cplusplus <= 201703L || __clang__
-dds::pub::detail::DataWriter<T>::~DataWriter<T>()
-#else // C++20 (CWG 2237)
 dds::pub::detail::DataWriter<T>::~DataWriter()
-#endif
 {
     if (!this->closed) {
         try {
diff --git a/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp b/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
index c880686..8a99caf 100644
--- a/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
+++ b/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
@@ -511,11 +511,7 @@ dds::sub::detail::DataReader<T>::common_constructor(
 }

 template <typename T>
-#if __cplusplus <= 201703L || __clang__
-dds::sub::detail::DataReader<T>::~DataReader<T>()
-#else // C++20 (CWG 2237)
 dds::sub::detail::DataReader<T>::~DataReader()
-#endif
 {
     if (!this->closed) {
         try {
diff --git a/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp b/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp
index b6f3b48..4bbec9c 100644
--- a/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp
+++ b/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp
@@ -207,11 +207,7 @@ dds::topic::detail::Topic<T>::Topic(const dds::domain::DomainParticipant& dp,
 }

 template <typename T>
-#if __cplusplus <= 201703L || __clang__
-dds::topic::detail::Topic<T>::~Topic<T>()
-#else // C++20 (CWG 2237)
 dds::topic::detail::Topic<T>::~Topic()
-#endif
 {
     if (!closed) {
         try {
```

k0ekk0ek:
Well, the problem isn't with clang. The problem is the specification. Other compilers, at least the ones we test against, simply don't throw warnings. I'd be willing to include just the command line option, but that doesn't solve the problem for applications because the compiler option isn't propagated through CMake if added to `add_compile_options`. I could add it to the target's compiler options, but that means it'll only every work when using CMake and impacts the options passed to the application (I suspect) the user is compiling and I'm not sure that's what the user wants.

Having talked to @reicheratwork offline, my preferred solution is: `#if __cplusplus >= 201402L && (__cplusplus <= 201703L || __clang__)`.

If that one's a bit too much, I think the second-best solution is to disable the warning for clang specifically in the header.

Third-best: Suppressing the warning through the target's options.

Feel free to state your preference and I'll update the PR accordingly, just wanted to make sure the options are weighted.

reicheratwork:
> Well, the problem isn't with clang. The problem is the specification. Other compilers, at least the ones we test against, simply don't throw warnings. I'd be willing to include just the command line option, but that doesn't solve the problem for applications because the compiler option isn't propagated through CMake if added to `add_compile_options`. I could add it to the target's compiler options, but that means it'll only every work when using CMake and impacts the options passed to the application (I suspect) the user is compiling and I'm not sure that's what the user wants.
> 
> Having talked to @reicheratwork offline, my preferred solution is: `#if __cplusplus >= 201402L && (__cplusplus <= 201703L || __clang__)`.
> 
> If that one's a bit too much, I think the second-best solution is to disable the warning for clang specifically in the header.
> 
> Third-best: Suppressing the warning through the target's options.
> 
> Feel free to state your preference and I'll update the PR accordingly, just wanted to make sure the options are weighted.

I personally think this is also the option with the smallest probability for unintended side-effects, though maybe a comment explaining the "why" behind the existence of the compiler version check macros would be a good addition.
A one line comment should be more than enough.
As sometime in the future an overeager contributor might undo this fix without knowing why it is there.

sumanth-nirmal:
@eboasson @reicheratwork @k0ekk0ek Apologies for the delayed response, as I was on holiday.

The code looks valid, However, after some reading and from this bug in LLVM https://bugs.llvm.org/show_bug.cgi?id=46979 I agree with @reicheratwork that the Clang compiler is just strictly following the C++ standard.
Quick look at the C++ standard [C++11 3.4.3 [basic.lookup.qual]](https://timsong-cpp.github.io/cppwp/n3337/basic.lookup.qual#6), [C++14 3.4.3 [basic.lookup.qual]](https://timsong-cpp.github.io/cppwp/n4140/basic.lookup.qual#6), [C++17 6.4.3 [basic.lookup.qual]](https://timsong-cpp.github.io/cppwp/n4659/basic.lookup.qual#6), [C++20 6.5.3 [basic.lookup.qual]](https://timsong-cpp.github.io/cppwp/n4861/basic.lookup.qual#6) says that the destructor should be written as `nested-name-specifier[opt] type-name :: ~ type-name` which is strictly followed by the Clang compiler but not with the other compilers we use here. However, this only started to change in the current C++23 working standard [draft 6.5.5 [basic.lookup.qual]](https://timsong-cpp.github.io/cppwp/basic.lookup.qual#general-4) which means this check `#if __cplusplus >= 201402L && (__cplusplus <= 201703L || __clang__)` will also not really solve the issue (unless we also include C++11 in this check).

I think the best is to disable this error with `Wdtor-name` diagnostic flag for Clang in the headers.

PR #191, opened 2021-12-12
idlcxx: add cmake argument to search additional include directories

mtudan:
Signed-off-by: Matija Tudan <tudan.matija@gmail.com>

### DIFF
--- a/src/idlcxx/Generate.cmake
+++ b/src/idlcxx/Generate.cmake
@@ -11,7 +11,7 @@
 #
 function(IDLCXX_GENERATE)
   set(one_value_keywords TARGET)
-  set(multi_value_keywords FILES FEATURES)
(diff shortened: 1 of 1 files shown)


eboasson (commented):
Makes sense to me, but it makes equal sense for other language bindings and so either this needs to be copied, or there should be a common IDL preprocessing function for all language bindings. The differences are really small, so it seems feasible.

It also would've ensured the cross-compilation PR #189 would have been unnecessary.

Would you be willing to have a look at a making a parameterized version of the generator function, add it to the `cyclonedds` and then use that one in the `Generate.cmake` there and here?

mtudan:
Sure, I can take a look into that. Could you just elaborate in more details what actually to change so that we are on the same page?

PR #193, opened 2021-12-16
Add support for StatusMask | StatusMask

ichernev:
Even though `StatusMask` inherits from `std::bitset`, if you attempt to do
`StatusMask|StatusMask` it complains because the resulting type is
a bitset, not a StatusMask. Also the expected operator supported for
a StatusMask should be OR not bitshift that accumulates.

    error: could not convert ‘std::operator|<32>(std::operator|<32>(std::operator|<32>(std::operator|<32>(dds::core::status::StatusMask::none(), dds::core::status::StatusMask::offered_deadline_missed()), dds::core::status::StatusMask::dds::core::status::StatusMask::liveliness_lost()), dds::core::status::StatusMask::publication_matched())’ from ‘std::bitset<32>’ to ‘dds::core::status::StatusMask’

### DIFF
--- a/src/ddscxx/include/dds/core/status/State.hpp
+++ b/src/ddscxx/include/dds/core/status/State.hpp
@@ -208,6 +208,13 @@ public:
         return *this;
     }
 
+    inline StatusMask operator | (const dds::core::status::StatusMask& mask) const
+    {
+        auto res = StatusMask(*this);
+        res |= mask;
+        return res;
+    }
+
     /**
      * Get all StatusMasks
      *


eboasson (commented):
Thanks @ichernev, that looks entirely sensible to me. @e-hndrks as you are my go-to guy for this kind of changes to the C++ API, do you agree?

e-hndrks:
Hi @eboasson, I agree that it would be nice to be able to use the | symbol to add bits to the StatusMask besides the << operator which is currently your only available mechanism to add them. That said, this proposed addition is a non-spec compliant operation and using it would make your application no longer portable to the C++ API of other DDS products. 

The spec does offer  a standardized way to add vendor specific additions (by putting them on the vendor specific Delegates and not on the spec-compliant wrapper classes and invoking them with the '->' instead of with the '.'), but this doesn't apply to the StatusMask since the latter doesn't have a vendor specific Delegate nor would it make sense to invoke the | operator through the -> operator.

Of course the Royal Road would be to first file an issue with the OMG specification, ask them to add this operation to the spec and when adopted make it available to our product. I am willing to file the issue with the OMG now, and put it on the agenda for the next Revision Task Force meeting. These task force meetings have been stalled for a while due to other commitments of its members, but I am hoping to host them again at the start of the new year. I could put this proposal on the agenda as a sensible and non-complicated hammer piece that we might all quickly agree on. If that is indeed the case I would be willing to prematurely add it to our API, in anticipation of the formal release of the next revised specification, whose release date is not yet carved in stone but would take at least a year before it has gone through all its official hoops.

Hope that gives some insight in where we are at right now. For now I would be hesitant to put it in already before having spoken with the other Task Force members.

ichernev:
Oh, wow. The reason I suggested the change was exactly for compatibility with other implementations, namely the RTI Connext implementation, which does support `|` but not `<<`. Also, from my humble dev experience `|` is a better way to merge `masks` than `<<`. I haven't read the OMG DDS spec in that regard, I was mostly doing it for RTI compat, and common sense, but I'll understand if you don't want to merge it for spec reasons.

eboasson:
This comment is in two parts: first a bit of venting, then my conclusion based on the information I got from both of you, @ichernev and @e-hndrks.

I am (only ever so slightly) fed up with the OMG DDS spec process. There are missing bits, various parties do as they please (sometimes literally ignoring what they wrote themselves), issues remain open for years, there are inconsistencies in the spec, IDL files that are invalid, stupid defaults, etc., etc. While I agree with @e-hndrks that the royal road would be ask the DDS PSIG at the OMG, I am not really prepared to wait for a response.

Besides which, I couldn't care less about the possibility of someone using a Cyclone-specific extension to the API and then finding another implementation doesn't support it. Why I do care about is when a conflict would arise, either with a future version of the specification or with another implementation that takes a different view of what a certain function or operator should do. But I don't see an obvious alternative way to handle the `|` operator on `StatusMask`.

> I was mostly doing it for RTI compat

And we already know the answer anyway ...

e-hndrks:
Hi @ichernev, I filed a ticker for this issue in the OMG portal (see https://issues.omg.org/issues/spec/DDS-PSM-Cxx/1.0?view=ALL#issue-48649 ). If others already do this they would probably not object, in which case we could quickly adopt it too. Will get back to you as soon as I know more.

PR #206, opened 2022-01-28
Preprocess documentation input at build time

k0ekk0ek:
Like it says in the title. Doing it at build time saves some configure time and makes it so that changes to files don't require reconfiguration.


eboasson (approved):
Thanks @k0ekk0ek !

PR #214, opened 2022-02-17
Add external field handling

reicheratwork:
According to XTypes v1.3 7.5.1.2.3.3 added the dds::core::external
template class
Added unittests to test dds::core::external functionality
Added parsing of @external fields in generated C++ code

### DIFF
--- a/src/ddscxx/include/dds/core/External.hpp
+++ b/src/ddscxx/include/dds/core/External.hpp
@@ -1,0 +1,114 @@
+#ifndef OMG_DDS_CORE_EXTERNAL_HPP_
+#define OMG_DDS_CORE_EXTERNAL_HPP_
+
+/*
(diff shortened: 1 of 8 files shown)


eboasson (commented):
I'd feel more comfortable if there were also some tests that use `@external` in IDL and then verify that it works. It seems like a self-evident thing to do, so perhaps you forgot to `git add` it to the commit?

reicheratwork:
> I'd feel more comfortable if there were also some tests that use `@external` in IDL and then verify that it works. It seems like a self-evident thing to do, so perhaps you forgot to `git add` it to the commit?

That completely skipped my mind!
I was just thinking of the C++ implementation, and it obeying the specs, not the eventual places it might be used, my bad.

eboasson (approved):
Thanks @reicheratwork !

PR #215, opened 2022-02-18
Add skipping hashing of keyless topics

reicheratwork:
As keyless topics have the same key value each time, "calculating"
it can be short-circuited significantly

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
@@ -63,10 +63,12 @@ bool to_key(const T& tokey, ddsi_keyhash_t& hash)
   if (!move(str, tokey, true))
     return false;
   size_t sz = str.position();
-  size_t padding = 16 - sz % 16;
-  if (sz != 0 && padding == 16) padding = 0;
+  size_t padding = 0;
+  if (sz < 16)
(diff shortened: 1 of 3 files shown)


src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -53,7 +53,14 @@ using org::eclipse::cyclonedds::topic::extensibility;
 using org::eclipse::cyclonedds::topic::encoding_version;
 using org::eclipse::cyclonedds::topic::TopicTraits;
 
-template<typename T>
+template<typename T, std::enable_if_t<TopicTraits<T>::isKeyless(), bool> = true >
+bool to_key(const T&, ddsi_keyhash_t& hash)
+{
+  memset(&(hash.value), 0x0, sizeof(hash.value));
+  return false;
```
eboasson, line 60: Doesn't this `false` imply that `populate_hash` still runs MD5 over it?

reicheratwork: It does, yes.

And with all instances having the same key value for a keyless topic, it should not matter whether this is hashed or not.

reicheratwork:
Added a fix that was uncovered by the tests added by @eboasson 
Thanks by the way!

eboasson (approved):
Thanks @reicheratwork 🙂

PR #223, opened 2022-03-03
Variant duplicate types fix

### LINKED ISSUE #127: Unions with identical types in different cases don't work
This:
```
union U switch(long)
{
  case 1:
    string aap;
  case 2:
    string noot;
};
```
causes (Apple) clang 12 in C++17 mode to error out with the following:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/variant:1440:59: error: no member named 'value' in 'std::__1::__find_exactly_one_t<std::__1::basic_string<char>, std::__1::basic_string<char>, std::__1::basic_string<char>>'
  return _VSTD::get<__find_exactly_one_t<_Tp, _Types...>::value>(__v);
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
x.hpp:75:17: note: in instantiation of function template specialization 'std::__1::get<std::__1::basic_string<char>, std::__1::basic_string<char>, std::__1::basic_string<char>>' requested here
    return std::get<std::string>(m__u);
```

So it seems that the generated `std::variant` should not have duplicates. IDL allows it (and it is also perfectly sensible to allow this once there are typedefs and complex types).

reicheratwork:
Added checking that the types supplied to the union container template class do not contain duplicates as std::variant can not handle this.

This fixes #127.

### DIFF
--- a/src/ddscxx/tests/Regression.cpp
+++ b/src/ddscxx/tests/Regression.cpp
@@ -201,3 +201,8 @@ TEST_F(Regression, propagate_xtypes_version_reqs)
   ASSERT_EQ(TopicTraits<td_3>::getExtensibility(), extensibility::ext_final);
   ASSERT_EQ(TopicTraits<td_3>::minXCDRVersion(), encoding_version::xcdr_v2);
 }
+
+TEST_F(Regression, union_duplicate_types)
+{
+  duplicate_types_union d_t;
+}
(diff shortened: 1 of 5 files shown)


eboasson (commented):
It looks good to me, but I think it would be wise to add some test cases, both to prove that it makes `std::variant` happy and to exercise the de-duplication code a bit

reicheratwork:
> It looks good to me, but I think it would be wise to add some test cases, both to prove that it makes `std::variant` happy and to exercise the de-duplication code a bit

I will add some tests which check whether the code generation works correctly.

eboasson (approved):
Thanks @reicheratwork !

PR #225, opened 2022-03-04
Update to use iceoryx 2.0

### LINKED ISSUE #224: Update cyclonedds-cxx to be compatible with iceoryx2.0
Update the current version of cyclonedds-cxx to use the iceoryx 2.0 version.

As `cyclonedds-cxx` depends on `iceoryx` directly only for using the `RoudiGtest` framework for testing. The difference should be only with the following changes in dependencies
- `iceoryx_posh_testing` instead of `iceoryx_posh_roudi_environment`
- `iceoryx_posh` instead of `iceoryx_utils`

sumanth-nirmal:
Fixes #224

This should be merged after https://github.com/eclipse-cyclonedds/cyclonedds/pull/1175

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -88,7 +88,7 @@ steps:
       set -e -x
       sudo apt-get install libacl1-dev libncurses5-dev pkg-config
       git clone --depth 1 \
-                --branch "${ICEORYX_BRANCH:-v1.0.1}" \
+                --branch "${ICEORYX_BRANCH:-release_2.0}" \
                 "${ICEORYX_REPOSITORY:-https://github.com/eclipse-iceoryx/iceoryx.git}" \
                 iceoryx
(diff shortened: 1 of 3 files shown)


.azure/templates/build-test.yml
```diff
@@ -88,7 +88,7 @@ steps:
       set -e -x
       sudo apt-get install libacl1-dev libncurses5-dev pkg-config
       git clone --depth 1 \
-                --branch "${ICEORYX_BRANCH:-v1.0.1}" \
+                --branch "${ICEORYX_BRANCH:-v1.90.0}" \
```
MatthiasKillat, line 91: This tag will probably not work with https://github.com/eclipse-cyclonedds/cyclonedds/pull/1175/files since the storage is not used anymore...

I.e. it will not work with iceoryx master (2.0) and 1.x at the same time.

PR #226, opened 2022-03-04
Union as topic and nested annotations

reicheratwork:
Add support for unions as topics
Add support for nested/topic annotations

### DIFF
--- a/src/idlcxx/src/traits.c
+++ b/src/idlcxx/src/traits.c
@@ -120,6 +120,11 @@ emit_topic_type_name(
   (void)revisit;
   (void)path;
 
+  if (idl_is_struct(node) && ((const idl_struct_t *)node)->nested.value)
+    return IDL_RETCODE_OK;
+  else if (idl_is_union(node) && ((const idl_union_t *)node)->nested.value)
+    return IDL_RETCODE_OK;
+
   struct generator *gen = user_data;
   char *name = NULL;
   if (IDL_PRINTA(&name, get_cpp11_fully_scoped_name, node, gen) < 0)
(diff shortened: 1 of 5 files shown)


src/idlcxx/src/traits.c
```diff
@@ -207,26 +207,28 @@ emit_traits(
     " };\n"
     "template<> constexpr const unsigned char TopicTraits<%1$s>::type_info_blob[] = {\n";
   const idl_struct_t *_struct = node;
+  const idl_union_t *_union = node;
```
eboasson, line 215: It is a bit of a nitpick, but it appears that `_struct` and `_union` serve no other purpose than in defining `selfcontained` and `ext`. Given that `node` can't be both a struct and a union at the same time, having these two defined all the time seems like asking for confusion.

Would you mind rewriting it a little bit, e.g., like:
```C
bool selfcontained;
idl_extensibility_t ext;
if (idl_is_struct(node)) {
  const idl_struct_t *_struct = node;
  selfcontained = sc_struct(_struct);
  ext = _struct->extensibility.value;
} else {
  assert(idl_is_union(node));
  ...
}
```
That would make me sleep a little better at night 😀

Otherwise it LGTM

eboasson, line 215: It is a bit of a nitpick, but it appears that `_struct` and `_union` serve no other purpose than in defining `selfcontained` and `ext`. Given that `node` can't be both a struct and a union at the same time, having these two defined all the time seems like asking for confusion.

Would you mind rewriting it a little bit, e.g., like:
```C
bool selfcontained;
idl_extensibility_t ext;
if (idl_is_struct(node)) {
  const idl_struct_t *_struct = node;
  selfcontained = sc_struct(_struct);
  ext = _struct->extensibility.value;
} else {
  assert(idl_is_union(node));
  ...
}
```
That would make me sleep a little better at night 😀

Otherwise it LGTM

reicheratwork: Will rewrite it into a function.

reicheratwork:
After conversing with @eboasson , I will refactor the requested changes into their own functions in a new PR, which will also include other improvements.

reicheratwork:
The refactoring is done in PR #227

PR #235, opened 2022-03-11
Array fixes

reicheratwork:
A number of fixes related to array handling:
- added batch copying of primitive types in arrays which should improve performance
- fixed union cases not taking array declarators into account
- added unittests to check these additions

### DIFF
--- a/src/ddscxx/tests/Regression.cpp
+++ b/src/ddscxx/tests/Regression.cpp
@@ -204,6 +204,8 @@ TEST_F(Regression, union_duplicate_types)
 {
   duplicate_types_union d_t;
   d_t.l_1(456789);
+  duplicate_types_union_2 d_t_2;
+  d_t_2.b_1({1,2,3,4,5});
   duplicate_sequences_union d_s;
   d_s.c_1({1,2,3,4,5,6});
 }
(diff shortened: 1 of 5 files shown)


eboasson (approved):

eboasson:
I don't feel quite confident fixing the conflict, so I tried fuzzing my guess a bit and it comes up with:
```
In file included from xxx-types.cpp:9:
./xxx-types.hpp:7725:33: error: use of undeclared identifier 'prop'; did you mean 'props'?
      if (!write(streamer, a_1, prop))
                                ^~~~
                                props
./xxx-types.hpp:7695:75: note: 'props' declared here
``` 
For a little bit of context:
```C+++
template<typename T, std::enable_if_t<std::is_base_of<cdr_stream, T>::value, bool> = true >
bool write(T& streamer, const ::x::a00016& instance, entity_properties_t &props) {
  (void)instance;
  if (!streamer.start_struct(props))
    return false;
  const auto d = instance._d();
  if (!write(streamer, d))
    return false;
  switch(d)
  {
    case 'A':
      {
      if (!write(streamer, instance.a00017()))
        return false;
      break; }
    case 'B':
      {
      auto prop = get_type_props<std::remove_cv_t<std::remove_reference_t<decltype(instance.a00029())>>>();
      if (!write(streamer, instance.a00029(), prop))
        return false;
      break; }
    case 'D':
      {
      if (!write(streamer, instance.a00030()))
        return false;
      break; }
    case 'E':
      {
      if (!streamer.start_consecutive(true, false))
        return false;
      for (const auto & a_1:instance.a00036()) {  //array depth 1
LINE 7725:      if (!write(streamer, a_1, prop))
        return false;
      }  //array depth 1
      if (!streamer.finish_consecutive())
        return false;
      break; }
  }
  return streamer.finish_struct(props);
}
```
WIth a bit of luck that's enough for you to understand what the problem is. Otherwise I can upload the full case (but it is quite large). In case, I guess I was too optimistic in approving it already ...

PR #236, opened 2022-03-11
Add warnings for unsupported annotations

reicheratwork:
The C++ backend will give warnings when it encounters annotations
it does not support (yet):
- @try_construct
- @default_literal


reicheratwork:
This relies on PR #1194 on CycloneDDS being merged.

PR #237, opened 2022-03-11
Add support for default parameters in constructors

reicheratwork:
Allows setting of default values in constructed types through the
@default annotation
At this point this only supports primitive non-enum types and strings
due to limitations in the idl parser
Add unittests to check whether this works as intended

### DIFF
--- a/src/idlcxx/src/types.c
+++ b/src/idlcxx/src/types.c
@@ -62,10 +62,10 @@ emit_member(
     value = "{ }";
   } else if (idl_is_enum(type_spec) || idl_is_base_type(type_spec) || idl_is_string(type_spec) || idl_is_bitmask(type_spec)) {
     if (mem->value.annotation) {
-      if ((idl_is_base_type(type_spec) || idl_is_string(type_spec)) && IDL_PRINTA(&value, get_cpp11_value, mem->value.value, gen) < 0)
+      if ((idl_is_base_type(type_spec) || idl_is_string(type_spec) || idl_is_bitmask(type_spec)) && IDL_PRINTA(&value, get_cpp11_value, mem->value.value, gen) < 0)
         return IDL_RETCODE_NO_MEMORY;
-      else if (idl_is_enum(type_spec) || idl_is_bitmask(type_spec))
-        return IDL_RETCODE_UNSUPPORTED;  //implement writing enum/bitmask value
+      else if (idl_is_enum(type_spec))
+        return IDL_RETCODE_UNSUPPORTED;  //implement writing enum value
     } else if (!idl_is_string(type_spec) && IDL_PRINTA(&value, get_cpp11_default_value, type_spec, gen) < 0) {
       return IDL_RETCODE_NO_MEMORY;
     }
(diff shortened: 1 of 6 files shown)


src/idlcxx/src/types.c
```diff
@@ -45,16 +45,26 @@ emit_member(
...
+
   name = get_cpp11_name(node);
   if (IDL_PRINTA(&type, get_cpp11_type, type_spec, gen) < 0)
     return IDL_RETCODE_NO_MEMORY;
   type_spec = idl_strip(type_spec, IDL_STRIP_ALIASES | IDL_STRIP_FORWARD);
-  if (idl_is_array(type_spec))
+  value = NULL;
+  if (idl_is_array(type_spec)) {
     value = "{ }";
-  else if (!idl_is_enum(type_spec) && !idl_is_base_type(type_spec))
```
eboasson, line 54: I get a conflict here because of the `bitmask` that appears straightforward to resolve, however ...

src/idlcxx/src/types.c
```diff
@@ -45,16 +45,26 @@ emit_member(
...
-    value = NULL;
-  else if (IDL_PRINTA(&value, get_cpp11_default_value, type_spec, gen) < 0)
-    return IDL_RETCODE_NO_MEMORY;
+  } else if (idl_is_enum(type_spec) || idl_is_base_type(type_spec) || idl_is_string(type_spec)) {
```
eboasson, line 58: ... this is a little too convoluted for me!

Isn't it simply:
* if mem->value.annotation != nullptr (i.e. no default value annotation)
  * if is_base_type || is_enum
    * if print default value fails: return NO_MEMORY
  * (else nothing to be done)
* else
  * if is_base_type || is_string
    * if print custom default value fails: return NO_MEMORY
  * else return UNSUPPORTED

?

I am not actually sure where the bitmask would end up, so maybe the conflict is not so straightforward after all!

eboasson, line 58: ... this is a little too convoluted for me!

Isn't it simply:
* if mem->value.annotation != nullptr (i.e. no default value annotation)
  * if is_base_type || is_enum
    * if print default value fails: return NO_MEMORY
  * (else nothing to be done)
* else
  * if is_base_type || is_string
    * if print custom default value fails: return NO_MEMORY
  * else return UNSUPPORTED

?

I am not actually sure where the bitmask would end up, so maybe the conflict is not so straightforward after all!

eboasson:
I think it'll do what it says, so that's good, but I was hoping that it would be possible to explain in a few words why `enum`s are complicated enough to be left out for now?

I am asking because I (perhaps naively) expect that `@default` takes a constant expression, that surely there's a way to convert a constant expression to an integer if it is of some integral type or enum, and that there must be a way to initialise an `enum class` from an integer. If all of that is the case, then it seems like it would be simple enough.

reicheratwork:
The issue is that defaults now store everthing as an idl_literal_t, which is not the type that enumerators resolve to, and enumerators cannot be stored in this type.

If we want to also allow enums in default annotations, we need to change the type of the default annotation is stored as, and then do more checking what type is stored when consuming it.

eboasson:
Thanks for explaining that, @reicheratwork. Wrong expectations lead to wrong conclusions 🙂

PR #241, opened 2022-03-17
Fixes and improvements regarding type properties

reicheratwork:
- Improved thread safety of get_type_props functions
- Simplified enumerator streaming functions
- Add enumerator type properties generations


eboasson (approved):
Great, thanks @reicheratwork !

PR #248, opened 2022-03-22
Fix for arrays of unions/structs in union cases

reicheratwork:
Arrays of unions/structs in union cases, did not retrieve the entity
properties for the streaming function, resulting in a failure to
compile


eboasson (approved):
Thanks @reicheratwork !

Issue #233, opened 2022-03-11
Header for array is not included in the generated code


sumanth-nirmal:
## Required information

**Operating system:**
Ubuntu 20.04 LTS

**Cyclone Version**

- cyclonedds - master branch, https://github.com/eclipse-cyclonedds/cyclonedds/commit/cf9ad3dc404ca8ba31973f8ce056ac19f980dd9b
- cyclonedds-cxx - master branch, 6d6467fd24b521250031dc211ebb6f1b15f06642

**Compiler**

- gcc `gcc version 8.4.0 (Ubuntu 8.4.0-3ubuntu2) `

**Observed result or behavior:**
If the IDL has an array of nested types, then the header for the array is not included

**Expected result or behavior:**
The headers should be correctly included in the generated code for all the used types

**Reproducer**
Having a simple IDL file like below

```hpp
module test_msgs {
module msg {

struct int32_ {
unsigned long uint_long;
};
};  // module msg
};  // module test_msgs
```
which is included in another IDL as below

```hpp
#include "int32.idl"

module test_msgs {
module msg {

struct StaticArray_ {
int32_ bool_values_[3];
};
};  // module msg
};  // module test_msgs
```

The generated code for this looks like below with out any header for arrays `<array>`

```hpp
...
#include "int32.hpp"


namespace test_msgs
{
namespace msg
{
class StaticArrays_
{
private:
 std::array<::test_msgs::msg::int32_, 3> bool_values__ = { };

public:
...
```

eboasson:
@reicheratwork is this as straightforward to fix as I think it is?

reicheratwork:
> @reicheratwork is this as straightforward to fix as I think it is?

Looks like the check for whether to include arrays isn't done if the type the array is on is from an include.
I will have a fix ready soon.

reicheratwork:
PR 249 will fix this issue

reicheratwork:
@sumanth-nirmal I think this issue is fixed in a commit (df7acb62aaf5c3be1ebc89787e04b096b7d6f72f) which has been merged to master.

So this issue can be closed, right?

sumanth-nirmal:
@reicheratwork Yes, can be closed

PR #249, opened 2022-03-24
Fix array include statement not being generated correctly (issue #233)

reicheratwork:
The check on whether or not to include the array statement was not
being done, because this check was being skipped if the type the
array was on was from included from outside the current file
By doing the array check on the declarator before the type check
the array include is now being done correctly


eboasson (approved):
Thanks @reicheratwork !

PR #259, opened 2022-04-06
Improve templating approach

reicheratwork:
By using improved templating models, we no longer need to generate code to handle the template types for instance for sequences, optionals, etc.
This should have the following advantages:
- fewer bugs due to the compiler handling unrolling/de-aliasing of containers/typedefs
- better legibility of generated code
- give the compiler better handles on optimizing the code as all the sequence/etc. code is only in the compiler's "imagination"

Any feedback/corrections/suggestions is very welcome


reicheratwork:
closed because this was way out of sync with the current master

PR #262, opened 2022-04-11
Fix issue 260

reicheratwork:
Fixes problems with compiling under MSVC15 (2017)

- Changed check for TopicTraits::isKeyless in tokey function to inside function block, as this is a constexpr this will result in the same optimized bytecode, as MSVC15 seems to resolve the template classes (TopicTraits and datatopic in the wrong order)
- Added include for <algorithm> which MSVC15 could not find

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/Missing.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/Missing.hpp
@@ -20,7 +20,7 @@
 
 //since these helpers are only introduced from c++14 onwards
 
-#if __cplusplus == 201103L
+#if __cplusplus < 201402L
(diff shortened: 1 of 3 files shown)


eboasson (commented):
LGTM, thanks @reicheratwork for figuring this out. I do have a request though: would you mind rebasing it on master and updating the PR to target master? It makes my life easier if I don't have to also keep track of fixes that need to be ported from release branches to master.

reicheratwork:
> LGTM, thanks @reicheratwork for figuring this out. I do have a request though: would you mind rebasing it on master and updating the PR to target master? It makes my life easier if I don't have to also keep track of fixes that need to be ported from release branches to master.

it is done

Issue #265, opened 2022-04-13
Not receiving instance state when using multiple (instance) writers 


kjrusscher:
### Situation
The following IDL file is used:
```
module RTP {
    @appendable
    struct Offer {
        @key string videoSourceID;
        string sdp;
    }; 
};
```
Two writers are started on the same topic but they are writing on separate instances; one uses `videoSourceID = "writer 1"`, the other uses `videoSourceID = "writer 2"`.   
There is one reader that does a take on the topic once every second, the code below is looped every second:
```
samplesRTPOffer = rtpOfferReader.take();
for (auto sample = samplesRTPOffer.begin(); sample < samplesRTPOffer.end(); ++sample) {
    dds::sub::status::InstanceState instanceState = sample->info().state().instance_state();
    if (instanceState == dds::sub::status::InstanceState::alive()) {
        // option1
        std::cout << sample->data().videoSourceid() << std::endl;
    } else if ((instanceState & dds::sub::status::InstanceState::not_alive_mask()) != 0) {
        // option2
        std::cout << sample->data().videoSourceid() << std::endl;
    }
}
````
The writers and reader are started in 3 separate applications.
### Problem
First I start the reader application. When I start both writers `option1` (see code above) is called twice. However, when I close both writers `option2` is only called once, when the last running writer application closes. 

What is even stranger to me is that `option2` prints the `videoSourceID` of the writer that started first, so not of the one that stopped last. I did a small test, 4 runs of starting 2 writers then closing them, the results are shown below.
| test # | start    |   start     | stop | stop | which videoSourceID is printed? |
|--|--------------|-----------|------------|-------------|---|
| 1 | writer 1 | writer 2 | writer 1 | writer 2 | writer 1 |
| 2 | writer 1 | writer 2 | writer 2 | writer 1 | writer 1 |
| 3 | writer 2 | writer 1 | writer 1 | writer 2 | writer 2 |
| 4 | writer 2 | writer 1 | writer 2 | writer 1 | writer 2 |

### Question
Is this expected behavior? I would expect to receive from both writers a `not-alive` 'notification'. But if this is expected behavior, I would expect to be notified of the last writer that stops, not the first one that started.

**Comment by kjrusscher** [comment]
I don't know if it matters, but for the topic I'm using the qos:
```
dds::topic::qos::TopicQos topicQos =
    domainParticipant.default_topic_qos() 
    << dds::core::policy::Reliability::Reliable() 
[...]

PR #267, opened 2022-04-20
Fix for issue #265

reicheratwork:
This issue was caused by the to_key function to not stream the key
values correctly when encountering a type which had keys and required
xtypes functionality
The streaming operations (move/write) failed when they shouldnt and
the error was just being passed as a non-hashing of the key value
contents
Fixed the streaming of keys in this case, and the passing of failures
to stream in the to_key function
Added regression tests to check for the correct generation of keys in
this scenario


reicheratwork:
@kjrusscher , if you could give this a look and check whether this fixes the issue on your end, then that would be much appreciated :)

kjrusscher:
@reicheratwork, I have tested it and I can confirm that this commit fixes issue #265.

eboasson (approved):

PR #268, opened 2022-05-02
add -Werror and -Wnull-dereference to examples

ghorn:
I'm trying to trigger an error I get locally when using cyclonedds-cxx as a third party dependency.


ghorn:
This shouldn't be merged

ghorn:
I wasn't able to reproduce

Issue #261, opened 2022-04-10
Performance degradation with serialization of nested structs


budrus:
The serialization performance at least for nested structs seems to be degraded with the latest cyclones-cxx. 

Used commits: 
- commits close to 0.9 release 
- CycloneDDS b3c386e328fc2015bfa5494d6840be973fce2e7e
- CycloneDDS-CXX 725cc2487818268f351303e5bc3576f4fac89483

Tests that reproduce this:
- [Open source performance test tool ](https://gitlab.com/ApexAI/performance_test)
   - [Used branch](https://gitlab.com/ApexAI/performance_test/-/tree/galactic-with-cyclonedds-0.9ish)
   
Expected behavior: 
- Performance test setup: Struct32k, 1 pub, 1 sub, 100 Hz, BEST_EFFORT, VOLATILE, intraprocess, shmem disabled 
- 100 samples are sent and received per second

Actual behavior:
- Only 1-2 samples are sent and received per second

reicheratwork:
Looking in to this case.

reicheratwork:
Can confirm this issue.
It seems that the entity_properties_t meta data container is not that good at being recursively created, lots of creating/copying of std::list going on.

I am working on a different (a flat in stead of a tree-like) approach for storing the sample metadata.

ErikAtApex:
I ran some more experiments. @reicheratwork Thanks for working on this; your fixes made a big difference. The performance for this deeply-nested structure is only 25% as fast as 0.8.1, but  much better than the 1-2 samples per second.

- Environment:
   - Erik's laptop
   - [Open source performance test tool](https://gitlab.com/ApexAI/performance_test)
   - [CycloneDDS-CXX docker container](https://gitlab.com/ApexAI/performance_test/-/blob/master/dockerfiles/Dockerfile.CycloneDDS-CXX)
   - Struct32k, 1 pub, 1 sub, 1000 Hz, BEST_EFFORT, VOLATILE
   - pub and sub are in the same process, but different threads
   - shmem disabled (CYCLONEDDS_URI environment variable is unset)
- Experiment 1:
   - CycloneDDS 0.8.1 tag
   - CycloneDDS-CXX 0.8.1 tag
   - 1ms mean latency
   - 1000 samples sent / received per second
- Experiment 2: (0.9ish, before this fix)
   - CycloneDDS `b3c386e328fc2015bfa5494d6840be973fce2e7e`
[...]

PR #271, opened 2022-05-06
Fix for issue #261

reicheratwork:
This fixes issue #261

Improved the performance of (de)serialization through the following changes:
- flattened the entity_properties tree as the std::list approach was not that quick
- created a custom stack which is faster than std::stack
- moved adminstration of the streaming status of entity offset (e_off) and entity size (e_sz) out the entity_properties and into the streamer class
- added skipping of size determination through streaming functions when serialized type has constant size
- removed member start/stop checking for basic cdr streams
- simpified a number of frequently used streaming functions
- changed from a range based for-loop to just running over the bare pointers in get_type_props as this vector will not change in size
- removed a number of unnecessary checks for abort_status, the streaming function will exit with a false status before
- only calculate whether byte swapping is necessary once, as the endiannesses will not change during the lifetime of the stream
- optimized reading/writing of single entry basic types, no longer an unaligned memcpy

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/basic_cdr_ser.hpp
@@ -41,15 +41,6 @@ public:
 
   /**
    * @brief
-   * Skips an entity, bypassing the stack.
-   *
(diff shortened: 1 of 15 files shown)


eboasson:
Apologies for my having forgotten about it once I decided to not rush it into the 0.9 release ...

PR #273, opened 2022-05-10
ADLINK Technology Ltd became ZettaScale Technology

eboasson:
This updates the copyright comments at the beginning of the files to reflect this change.

Companion to https://github.com/eclipse-cyclonedds/cyclonedds/pull/1254: I confess to completely forgetting to update them also in the C++ and Python repositories ...

### DIFF
--- a/.azure/codecov.yml
+++ b/.azure/codecov.yml
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2021 ADLINK Technology Limited and others
+# Copyright(c) 2021 ZettaScale Technology and others
 #
 # This program and the accompanying materials are made available under the
 # terms of the Eclipse Public License v. 2.0 which is available at
--- a/.azure/coverity-scan.yml
+++ b/.azure/coverity-scan.yml
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2021 ADLINK Technology Limited and others
+# Copyright(c) 2021 ZettaScale Technology and others
 #
 # This program and the accompanying materials are made available under the
 # terms of the Eclipse Public License v. 2.0 which is available at
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -1,5 +1,5 @@
 #
-# Copyright(c) 2021 ADLINK Technology Limited and others
+# Copyright(c) 2021 ZettaScale Technology and others
 #
 # This program and the accompanying materials are made available under the
 # terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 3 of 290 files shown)


reicheratwork (approved):
Looks good to me, apart from some inclarity coming from the differences between the years in some files (for instance 2020 changing to 2021 or 2022) e.g. TCorePolicyImpl.hpp or basic_cdr_ser.hpp.
Is there a specific reason for these choices, or could it be a better idea to update all years to 2022 or keep all years the same, just to head off any possible confusion?

eboasson:
Thanks for reviewing @reicheratwork ☺️ Regarding your question:

> apart from some inclarity coming from the differences between the years in some files (for instance 2020 changing to 2021 or 2022) e.g. TCorePolicyImpl.hpp or basic_cdr_ser.hpp.
> Is there a specific reason for these choices, or could it be a better idea to update all years to 2022 or keep all years the same, just to head off any possible confusion?

There is a hint in the commit message:

> It turns out in many cases the years weren't updated when editing, so I've tried to get the correct years in there based on the git commit log.

I built a script that for each file, inspects the git log, figures out the range of dates and authors, and updates accordingly. For example, no-one paid attention when editing `TCorePolicyImpl.hpp` and so it kept the 2020 year even though it did get changed in 2021. The case of `basic_cdr_ser.hpp` is similar, except there the most recent change was actually in 2022. Simply always including 2022 would perhaps raise fewer questions, but I have a problem with that when nothing changed.

In short, this is the best compromise that I could think of.

reicheratwork:
> Thanks for reviewing @reicheratwork ☺️ Regarding your question:
> 
> > apart from some inclarity coming from the differences between the years in some files (for instance 2020 changing to 2021 or 2022) e.g. TCorePolicyImpl.hpp or basic_cdr_ser.hpp.
> > Is there a specific reason for these choices, or could it be a better idea to update all years to 2022 or keep all years the same, just to head off any possible confusion?
> 
> There is a hint in the commit message:
> 
> > It turns out in many cases the years weren't updated when editing, so I've tried to get the correct years in there based on the git commit log.
> 
> I built a script that for each file, inspects the git log, figures out the range of dates and authors, and updates accordingly. For example, no-one paid attention when editing `TCorePolicyImpl.hpp` and so it kept the 2020 year even though it did get changed in 2021. The case of `basic_cdr_ser.hpp` is similar, except there the most recent change was actually in 2022. Simply always including 2022 would perhaps raise fewer questions, but I have a problem with that when nothing changed.
> 
> In short, this is the best compromise that I could think of.

@eboasson Right! Well then I have no reservations whatsoever regarding this PR :P

PR #280, opened 2022-05-29
feature c++11 compatibility

### LINKED ISSUE #185: C++ standard version?
Hello cyclonedds-cxx,

we have some legacy platforms which require the standard c++11.
The generated code from idlc seems to contain std::variant which would require c++17.
I did not found a statement which the lowest support c++ standard version is.

May you could add this information in the README file for more clearance?

### LINKED ISSUE #278: Broken C++11 since Mar 18 2022
Broken C++11 compatibility since Mar 18 2022 see here https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/b6310e37d325b48d1c5b14484e938b250d878660
Prior to this i was able to use cyclonedds_cxx  with c++11 projects, see here ref #185

So far i investigated two problems:

#### Problem 1: static_assert
(static_assert syntax changed since c++17)
https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/fdbef1f77433cdac9957ab3001f50af6ef42b08e/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/entity_properties.hpp#L256
Possible fix:
```
  #if __cplusplus >= 201703L // c++17 or higher
  static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8);
  #else
  static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8, "");
  #endif
```
#### Problem 2: std::enable_if_t
(Its a c++14 feature)
https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/fdbef1f77433cdac9957ab3001f50af6ef42b08e/src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/entity_properties.hpp#L253

### Questions

Hey @eboasson 😊 i wanted to ask you if c++11 compatibility is still "something we intend to be"?
https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/185#issuecomment-989684242

@reicheratwork may you know a workaround to solve Problem 2 for me? It would be awesome for me 😃

trittsv:
This introduces a c++11 legacy mode - meaning the support of c++11.
Boost is added as a dependency (components variant, optional) if the legacy mode is enabled.
C++11 seems like a good tradeoff between the old and the new world.

Background:
Some projects/products can't use the latest compiler features because they have to support some limited platforms (e.g. QNX 7.0)

closes #185 
closes #278 

Example usage:
```
mkdir build
cd build
cmake -G "Unix Makefiles" -DENABLE_LEGACY=ON ..
cmake --build .
```


trittsv:
The remaining build error is not related with this PR. It also happens on the master branch.

eboasson:
I'll review, but I can't resist saying: perhaps I should've looked at this before looking at #278 ... 😁

src/ddscxx/tests/CMakeLists.txt
```diff
@@ -83,7 +94,13 @@ if(DEFINED ENV{SYSTEM_TEAMFOUNDATIONSERVERURI})
   endif()
 endif()
 
-set_property(TARGET ddscxx_tests PROPERTY CXX_STANDARD 17)
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  # Microsoft does not specify the __cplusplus macro by default
+  # https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus
```
eboasson, line 99: `entity_properties.hpp` gets included by application code, so doesn't this lovely "feature" of MSVC++ now break C++ application code?

I suppose we could force `/Zc:__cplusplus` into the CMake config files for cyclonedds-cxx, but then simply including Cyclone would likely change the definition for the application as well.

Altogether I think it would make more sense to make sure `features.hpp.in` adds an entry for `ENABLE_LEGACY` (obviously it would have to be a slightly different name, like `DDSCXX_USE_BOOST` or so), and then use that instead of checking `__cplusplus`.

.travis.yml
```diff
@@ -182,6 +182,8 @@ jobs:
       env: [ ARCH=x86_64, BUILD_TYPE=Debug, SANITIZER=address ]
     - <<: *macos1015_xcode11_5
       env: [ ARCH=x86_64, BUILD_TYPE=Release ]
+    - <<: *macos1015_xcode11_5
```
eboasson, line 185: It seems like we forgot to remove `.travis.yml`, it hasn't been maintained for quite some time. You can throw it away, or I'll do it ...

examples/helloworld/CMakeLists.txt
```diff
@@ -16,7 +16,16 @@ if(NOT TARGET CycloneDDS-CXX::ddscxx)
...
+set(features_list "")
+if(ENABLE_LEGACY)
+  list(APPEND features_list "union-template=boost::variant")
+  list(APPEND features_list "union-include=\\\"boost/variant.hpp\\\"")
+
+  find_package(Boost)
+  include_directories(${Boost_INCLUDE_DIR})
+endif()
+
+idlcxx_generate(TARGET helloworlddata FILES HelloWorldData.idl WARNINGS no-implicit-extensibility FEATURES ${features_list})
```
eboasson, line 28: Is this still needed with the changes `src/idlcxx/src/generator.c`?

Ideally the output of IDLC would be completely target independent, but that would mean having to make this decision in the generated output. That should be possible, but perhaps it is not worth the effort for a "legacy" mode!

src/ddscxx/tests/CMakeLists.txt
```diff
@@ -11,6 +11,15 @@
 #
 find_package(GTest REQUIRED)
 
+set(features_list "")
+if(ENABLE_LEGACY)
+  list(APPEND features_list "union-template=boost::variant")
```
eboasson, line 16: Same comment as for `examples/helloworld/CMakeLists.txt`

src/ddscxx/include/org/eclipse/cyclonedds/core/cdr/entity_properties.hpp
```diff
@@ -18,7 +18,14 @@
 #include <map>
 #include <atomic>
 #include <mutex>
+#if __cplusplus >= 201703L // c++17
+#define STD_IMPL std
```
eboasson, line 22: IIRC this file ends up in application code so we need to be carefully with macro names. `DDSCXX_STD_IMPL` seems a lot safer.

Alternatively (but here my limited knowledge of C++ puts me on rather shaky ground), perhaps we could do
```C++
#if __cplusplus >= 201703L
namespace std_impl = std;
#else
namespace std_impl = boost;
#endif
```
inside the `org::eclipse::cyclonedds::core::cdr` namespace? (So roughly where `decl_ref_type` is defined — and this makes me think that `decl_ref_type` itself probably slipped through the net ... but that's a different story.)

eboasson (commented):
Thanks @trittsv! It looks good to me in principle, but there are few details that I have some questions on.

trittsv:
Thanks for your review @eboasson - you made some really good points.

I implemented your suggestions, may you can review it again? 😇

trittsv:
Hi @eboasson, dont worry, the merge conflict is solved now😊

eboasson (approved):
Thanks!

Issue #276, opened 2022-05-20
memory_order_release idl cxx code generator


nate800:
The code generator produces "std::memory_order::memory_order_release", which will only be part of the C++ standard until C++20. 

https://en.cppreference.com/w/cpp/atomic/memory_order

reicheratwork:
> The code generator produces "std::memory_order::memory_order_release", which will only be part of the C++ standard until C++20.
> 
> https://en.cppreference.com/w/cpp/atomic/memory_order

@nate800  you are correct, seems like the memory order is only an enum class since C++20.
But at the same time, there are mappings of the old (non-scoped) enumerator names to the new scoped values, which should make it very easy to have code generated that is independent on whether the compiler follows C++11 or C++20.

I will reference this issue in a PR which will contain a fix.

PR #282, opened 2022-05-30
Fix issue 276 and type/topic discovery issues

reicheratwork:
This PR fixes:
- issue #276 by replacing the scoped enumerator `std::memory_order::memory_order_release`  with its non-scoped alias `std::memory_order_release`
- issues uncovered by the newer builds of CycloneDDS which now supports topic/type discovery by default
  - defining static members in template classes to contain the type/topic discovery information was not accepted by compilers other than GCC
  - to work around this this data is now accessed by static accessor functions
  - added unittests to check the correct generation of topic/type discovery information
  
@nate800 could you give this a try (and a review)?


ghorn:
Thank you!

eboasson (approved):

PR #285, opened 2022-06-02
Replace CMAKE_PROJECT_NAME with PROJECT_NAME

vectorsli:
CMAKE_PROJECT_NAME always refers to the root project while PROJECT_NAME is the current project.

Using CMAKE_PROJECT_NAME causes problems when using `add_subdirectory` to include CycloneDDS-CXX.


vectorsli:
@eboasson Build passes now

eboasson (approved):
LGTM, Thanks @vectorsli !

PR #290, opened 2022-06-14
Update the SHM data path with content filtered topics

### LINKED ISSUE #288: Segfault in read/take path with SHM when using ContentFilteredTopic
## Required information

**Operating system:**
Ubuntu 20.04 LTS

**Cyclone Version**

- cyclonedds-cxx - master branch, 9a0ebe219c0c0112f550cfb23867af08deccbbc5

**Observed result or behavior:**

When using a content topic filter on the reader, then read/take with SHM enabled causes a segfault

**Expected result or behavior:**

There should be no segfault and all the data paths with SHM should be handled gracefully.

**Details**

When a content filtered topic is used, then the filter mode is set to `DDS_TOPIC_FILTER_SAMPLE_*` here https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/6ec026863a8a738d4980bfbfed13f35c55f3aed0/src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp#L411

Because of this when a sample is received over SHM using iceoryx, when the sample is being stored in RHC, we check the content filter [content_filter_accepts](https://github.com/eclipse-cyclonedds/cyclonedds/blob/a1f1f9cad6ecc96b149d538c07cc8c47fd6b814c/src/core/ddsc/src/dds_rhc_default.c#L1433) and since the content filter is configured, we call `ddsi_serdata_to_sample` [here](https://github.com/eclipse-cyclonedds/cyclonedds/blob/a1f1f9cad6ecc96b149d538c07cc8c47fd6b814c/src/core/ddsc/src/dds_rhc_default.c#L988) this calls `serdata_to_sample` https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/9a0ebe219c0c0112f550cfb23867af08deccbbc5/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L497 in which we call `deserialize_sample_from_buffer` with `ptr->data()` which is a `nullptr` since the actual data is in the `iox_chunk` https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/9a0ebe219c0c0112f550cfb23867af08deccbbc5/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L506

sumanth-nirmal:
Fixes #288

### DIFF
--- a/src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
+++ b/src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
@@ -275,6 +275,9 @@ public:
                 /* Empty: the exception throw should have already traced an error. */
             }
         }
+        // delete the functor
+        delete myFunctor;
+        myFunctor = nullptr;
     }
 
     virtual void close()
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
@@ -510,10 +510,14 @@ bool serdata_to_sample(
 {
   (void)bufptr;
   (void)buflim;
-  auto ptr = static_cast<const ddscxx_serdata<T>*>(dcmn);
 
-  auto& msg = *static_cast<T*>(sample);
-  return deserialize_sample_from_buffer(ptr->data(), ptr->size(), msg);
+  auto typed_sample_ptr = static_cast<T*>(sample);
+  // cast away const, with the reasoning that we don't modify the underlying ddsi_serdata which
+  // is actually const, we only modify the ddscxx_serdata non const contents
+  auto d = const_cast<ddscxx_serdata<T>*>(static_cast<const ddscxx_serdata<T>*>(dcmn));
+  *typed_sample_ptr = *d->getT();
+
+  return true;
 }
 
 template <typename T, class S>
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
@@ -515,8 +515,12 @@ bool serdata_to_sample(
   // cast away const, with the reasoning that we don't modify the underlying ddsi_serdata which
   // is actually const, we only modify the ddscxx_serdata non const contents
   auto d = const_cast<ddscxx_serdata<T>*>(static_cast<const ddscxx_serdata<T>*>(dcmn));
-  *typed_sample_ptr = *d->getT();
 
+  auto t_ptr = d->getT();
+  if (!t_ptr)
+    return false;
+
+  *typed_sample_ptr = *t_ptr;
   return true;
 }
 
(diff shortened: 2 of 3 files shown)


src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
```diff
@@ -275,6 +275,9 @@ class ContentFilteredTopic  :
                 /* Empty: the exception throw should have already traced an error. */
             }
         }
+        // delete the functor
+        delete myFunctor;
+        myFunctor = nullptr;
```
sumanth-nirmal, line 280: This fixes a leak with the filter function. As we didn't have tests with the content-filtered topics, we didn't catch this!

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -500,10 +510,14 @@ bool serdata_to_sample(
...
   (void)bufptr;
   (void)buflim;
-  auto ptr = static_cast<const ddscxx_serdata<T>*>(dcmn);
 
-  auto& msg = *static_cast<T*>(sample);
-  return deserialize_sample_from_buffer(ptr->data(), ptr->size(), msg);
+  auto typed_sample_ptr = static_cast<T*>(sample);
+  // cast away const, with the reasoning that we don't modify the underlying ddsi_serdata which
+  // is actually const, we only modify the ddscxx_serdata non const contents
+  auto d = const_cast<ddscxx_serdata<T>*>(static_cast<const ddscxx_serdata<T>*>(dcmn));
```
sumanth-nirmal, line 517: I thought of using some free functions to get the sample from `serdata` (either from `iox_chunk` or actually deserialize from the buffer) instead of using `const_cast`. However, with that approach, we will need always to deserialize since we don't store the sample as part of `ddscxx_serdata`. So I took the approach to cast away the const on `ddscxx_serdata` and then modify the non-const fields in `ddscxx_serdata`. 

There is also another place in the code here https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/653e97e2ba5bc3a81ecf9c4edba3493c0f08bdb8/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L516 where this is done, so I guess this should be okay.

CC @eboasson @reicheratwork

reicheratwork: This looks like it does what it's supposed to and with fewer lines of code and without the complete deserialization step that deserialize_sample_from_buffer introduced in all cases, here the deserialization will only be done if the sample is not yet deserialized in the ddscxx_serdata entity. So this should be faster.

The only issue i see is it not checking the result of any deserialization. While it is a small probability, something could have gone wrong during deserialization and then the getT function will return a nullptr.
I suggest the following small change from:
```c++
 *typed_sample_ptr = *d->getT();

  return true;
```
to:
```c++
  auto t_ptr = d->getT();
  if (!t_ptr)
    return false;

  *typed_sample_ptr = *t_ptr;
  return true;
```

src/ddscxx/include/dds/topic/detail/ContentFilteredTopic.hpp
reicheratwork: Good catch!

reicheratwork (changes requested):
Looks good to me, though I would like the extra check in `serdata_to_sample` to be implemented before this is merged.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
sumanth-nirmal: @reicheratwork Good point, fixed it.

sumanth-nirmal:
@reicheratwork I fixed it as per your suggestions, please take a look again.

eboasson:
Thanks a lot @sumanth-nirmal and @reicheratwork

eboasson (approved):

PR #291, opened 2022-06-14
Disable inheritance for interface classes

reicheratwork:
As the interface classes (DataWriter, Topic, DataReader, etc.) in the
C++ binding make liberal use of virtual inheritance, they can not be
inherited from in the normal sense, as calling the inherited
constructor in the derived constructor will cause the Reference class
contained within to be initialized before the derived constructor is
called, producing an exception
To this end, these interface classes are made final, which disables
inheriting from them


eboasson (approved):
LGTM. It can't really oversee all the consequences, but it seems like a really stupid thing to try to inherit from these classes and so forbidding it altogether seems sensible. All the more so if it solves a real problem. Thanks @reicheratwork !

PR #294, opened 2022-07-14
Update api for changed dds_find_topic

dpotman:
This updates the c++ binding for the changes in the dds_find_topic function from [Cyclone PR 1308](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1308): find_topic expects a type_info, which is retrieved from the topic traits.

Note: this PR is in draft, because I have to find out why the test `DataWriter.qos_nondefault_set` failed unexpectedly, and was fixed removing the data representation from the `lifespan` QoS that is used for this test.

### DIFF
--- a/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp
+++ b/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp
@@ -307,7 +307,8 @@ dds::topic::detail::Topic<T>::discover_topic(
         const dds::core::Duration& timeout)
 {
     dds::topic::Topic<T> found = dds::core::null;
-    dds_entity_t ddsc_topic = dp.delegate()->lookup_topic(name, timeout);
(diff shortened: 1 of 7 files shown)


dpotman:
The call to `dds_ensure_valid_data_representation` that was added to `dds_create_topic_impl` was causing the above mentioned test to fail. In that test, the topic will get both XCDR1 and XCDR2 representations in `dds_ensure_valid_data_representation`, and the writer inherits this qos from the topic. Setting a qos that has only XCDR1 after the writer was created fails because the qos is immutable. 

Although it seems logical to ensure the correct data representation is set in `dds_create_topic_impl` (so in fact for every topic that is created), I'm not completely sure if this is correct in all cases. For topics created from a topic descriptor the qos is already set in `dds_create_topic`, and the call in `dds_create_topic_impl` (which uses the data representation from the sertype, that is copied from the topic descriptor) is a no-op. For topics created from RMW the sertype implementation forces XCDR1 only, so I think that will also work as expected. @eboasson what do you think about this?

eboasson (approved):
LGTM and it all checks out locally, so it must be good :)

PR #297, opened 2022-07-25
Properly propagate dependency to Boost if ENABLE_LEGACY is true

vectorsli:
We got missing boost include errors when testing the latest master in legacy C++11 mode. This PR should fix the issue and correctly declare Boost as a transitive dependency of CycloneDDS-CXX.

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -144,7 +144,8 @@ steps:
       INSTALLPREFIX="$(pwd)/install"
       mkdir helloworld
       cd helloworld
-      cmake -DCMAKE_PREFIX_PATH="${BUILD_SOURCESDIRECTORY}/cyclonedds/build/install;${BUILD_SOURCESDIRECTORY}/iceoryx/build/install;${BUILD_SOURCESDIRECTORY}/build/install" \
+      conan install -b missing -s arch=${ARCH} -s build_type=${BUILD_TYPE} -g cmake_find_package_multi ../..
+      cmake -DCMAKE_PREFIX_PATH="${BUILD_SOURCESDIRECTORY}/cyclonedds/build/install;${BUILD_SOURCESDIRECTORY}/iceoryx/build/install;${BUILD_SOURCESDIRECTORY}/build/install;${BUILD_SOURCESDIRECTORY}/build/helloworld" \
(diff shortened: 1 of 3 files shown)


vectorsli:
It looks like the CI is failing because one of the configurations builds with `ENABLE_LEGACY` but `Boost` isn't available when building the example. Can you guys please have a quick look? @eboasson?

EttoreVidotto:
@vectorsli I tried to build my project based on your PR, and it worked for me. Thanks for your effort!

eboasson:
@vectorsli 

> It looks like the CI is failing because one of the configurations builds with `ENABLE_LEGACY` but `Boost` isn't available when building the example. Can you guys please have a quick look? @eboasson?

I think boost gets installed via conan, so perhaps this:
```
diff --git a/.azure/templates/build-test.yml b/.azure/templates/build-test.yml
index f2fdc402..1aaaf777 100644
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -140,6 +140,7 @@ steps:
   - bash: |
       set -e -x
       cd build
+      conan install -b missing -s arch=${ARCH} -s build_type=${BUILD_TYPE} ..
       ctest --output-on-failure -T test -C ${BUILD_TYPE}
       INSTALLPREFIX="$(pwd)/install"
       mkdir helloworld
```
is what's needed?

I'd be much obliged if you could give a quick try.

EttoreVidotto:
@eboasson I tried to include the suggested line, but it did not solve the problem. @vectorsli Any idea?

eboasson:
@vectorsli I guess that file is in a different location when building the examples than I imagined ... sorry about that ...

vectorsli:
I have a feeling this time it works :)

eboasson:
Right, so now conan's happy but it still doesn't fix it. There must be a bit of magic involved to make CMake look in the right place ...

vectorsli:
Yeah, setting the `CMAKE_PREFIX_PATH` is the final step. I'll have a look

vectorsli:
Somehow the CI didn't like the last commit and didn't start. I squashed everything together. Let's see.

eboasson (approved):
LGTM, @vectorsli, thanks a lot!

PR #298, opened 2022-07-29
Update documentation build

thijsmie:
This slightly restructures the docs build for C++. After some struggling I decided to remove the CMake procedure that pre-processes the headers to hide the `DELEGATE` templating. It does not always yield correct output and breaks some referencing. Now you need some more clicks in the documentation to actually find what you are looking for, but it always correctly reflects the real code, which I think is more valuable. (And my opinion is that we should get rid of the DELEGATE's altogether anyway, it just makes our lives and our users lives harder under the guise of API portability, which can also be achieved without DELEGATE's anyway.)


eboasson (approved):
I haven't re-generated the documentation locally, but based on the changes that I see, I do think it makes sense to merge them — thanks @thijsmie !

PR #307, opened 2022-08-22
Update waitset to detach conditions on cleanup

### LINKED ISSUE #306: Detach conditions from waitset automatically when closed
## Required information

**Operating system:**
Ubuntu 20.04 LTS

**Cyclone Version**
- cyclonedds - master branch, https://github.com/eclipse-cyclonedds/cyclonedds/commit/bcdaf065083da970b5a976bf29fd76bb68e943ac
- cyclonedds-cxx - master branch, https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/b01b1b3cec5c3fef756e74af6842341de7f023ca

**Observed result or behavior:**

When conditions that are attached to a waitset are closed, then these are not automatically detached from waitset, and it's not even possible to detach these closed conditions from the waitset (related to #172)

**Expected result or behavior:**

Any attached conditions to waitset should be detached automatically when these conditions are closed (or cleaned up)

**Related Issues**
[Closed conditions can not be detached from the Waitset](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/172)

sumanth-nirmal:
This MR changes the behavior of waitset to detach the conditions automatically when the conditions are closed or cleaned up.

Fixes #306

### DIFF
--- a/src/ddscxx/src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
+++ b/src/ddscxx/src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
@@ -47,24 +47,33 @@ void
 org::eclipse::cyclonedds::core::cond::ConditionDelegate::detach_from_waitset(
     const dds_entity_t entity_handle) {
     std::vector<WaitSetDelegate *> waitset_list_tmp;
-    org::eclipse::cyclonedds::core::ScopedMutexLock scopedLock(this->waitSetListUpdateMutex);
+    org::eclipse::cyclonedds::core::ScopedMutexLock scopedLockForCopy(this->waitSetListUpdateMutex);
     waitset_list_tmp.assign(this->waitSetList.begin(), this->waitSetList.end());
-    scopedLock.unlock();
+    scopedLockForCopy.unlock();
 
     for (auto waitset : waitset_list_tmp) {
         org::eclipse::cyclonedds::core::ScopedObjectLock scopedWaisetLock(*waitset);
-        if (this->waitSetList.erase(waitset)) {
-            waitset->remove_condition_locked(this, entity_handle);
-       }
+        {
+            org::eclipse::cyclonedds::core::ScopedMutexLock scopedLock(this->waitSetListUpdateMutex);
+            // remove the waitset from the list and detach the condition
+            if (this->waitSetList.erase(waitset)) {
+                waitset->remove_condition_locked(this, entity_handle);
+            }
+        }
     }
 }
 
 void
-org::eclipse::cyclonedds::core::cond::ConditionDelegate::close(
+org::eclipse::cyclonedds::core::cond::ConditionDelegate::detach_and_close(
     const dds_entity_t entity_handle)
 {
-    // detach the condition from any waitsets
     detach_from_waitset(entity_handle);
+    org::eclipse::cyclonedds::core::cond::ConditionDelegate::close();
+}
+
+void
+org::eclipse::cyclonedds::core::cond::ConditionDelegate::close()
+{
     // close the condition
     DDScObjectDelegate::close();
 
--- a/src/ddscxx/src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
+++ b/src/ddscxx/src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
@@ -39,17 +39,38 @@ org::eclipse::cyclonedds::core::cond::ConditionDelegate::~ConditionDelegate()
     // sure that functor is delete here
     if (this->myFunctor) {
         delete this->myFunctor;
+        this->myFunctor = nullptr;
     }
 }
 
 void
-org::eclipse::cyclonedds::core::cond::ConditionDelegate::close()
+org::eclipse::cyclonedds::core::cond::ConditionDelegate::detach_from_waitset(
+    const dds_entity_t entity_handle) {
+    std::vector<WaitSetDelegate *> waitset_list_tmp;
+    org::eclipse::cyclonedds::core::ScopedMutexLock scopedLock(this->waitSetListUpdateMutex);
+    waitset_list_tmp.assign(this->waitSetList.begin(), this->waitSetList.end());
+    scopedLock.unlock();
+
+    for (auto waitset : waitset_list_tmp) {
+        org::eclipse::cyclonedds::core::ScopedObjectLock scopedWaisetLock(*waitset);
+        if (this->waitSetList.erase(waitset)) {
+            waitset->remove_condition_locked(this, entity_handle);
+       }
+    }
+}
+
+void
+org::eclipse::cyclonedds::core::cond::ConditionDelegate::close(
+    const dds_entity_t entity_handle)
 {
+    // detach the condition from any waitsets
+    detach_from_waitset(entity_handle);
+    // close the condition
     DDScObjectDelegate::close();
 
     if (this->myFunctor) {
         delete this->myFunctor;
-        this->myFunctor = NULL;
+        this->myFunctor = nullptr;
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/ddscxx/src/org/eclipse/cyclonedds/core/cond/StatusConditionDelegate.cpp
```diff
@@ -53,7 +53,15 @@ org::eclipse::cyclonedds::core::cond::StatusConditionDelegate::close()
...
     /* todo: cannot close this->ddsc_entity because this is the parent EntityDelegate;
         find better solution to get correct feedback when closed StatusCondition is used */
-    this->ddsc_entity = 0;
+
+    // detach the condition from any waitset
+    this->check();
+    org::eclipse::cyclonedds::core::ScopedObjectLock scopedLock(*this);
+    ConditionDelegate::detach_from_waitset(ddsc_entity);
+    // set the entity to 0
+    this->ddsc_entity = DDS_HANDLE_NIL;
```
sumanth-nirmal, line 62: Can't really close the status condition because of the reason above, so currently it detaches the condition from any waitset but doesn't close the entity.

However, since the entity handle is reset to nil, it's anyway not usable. I think I didn't make this any worse if I didn't improve the behavior.

src/ddscxx/src/org/eclipse/cyclonedds/core/cond/WaitSetDelegate.cpp
```diff
@@ -66,7 +73,9 @@ org::eclipse::cyclonedds::core::cond::WaitSetDelegate::wait(
     const dds::core::Duration& timeout)
 {
     dds_duration_t c_timeout = org::eclipse::cyclonedds::core::convertDuration(timeout);
+    org::eclipse::cyclonedds::core::ScopedObjectLock scopedLock(*this);
     const size_t sz = conditions_.size();
```
sumanth-nirmal, line 77: I assume this size should be fine even if some conditions are removed later, since its upper bound size for the conditions. Is this assumption correct?

src/ddscxx/tests/WaitSet.cpp
```diff
@@ -817,10 +829,8 @@ TEST_F(WaitSet, destructor)
 
 /**
  * Check that it is possible to remove and attach multiple conditions while waiting
- *
- * TODO: disabled because the conditions are currently not removed from the waitset
  */
-TEST_F(WaitSet, DISABLED_attach_detach_multiple_during_wait)
+TEST_F(WaitSet, attach_detach_multiple_during_wait)
```
sumanth-nirmal, line 833: This test should work now as we detach conditions during cleanup

src/ddscxx/tests/WaitSet.cpp
```diff
@@ -928,4 +947,131 @@ TEST_F(WaitSet, status_condition_trigger)
...
+  // can't really close a status condition, enable the test after the fix
+// close status condition
+//  {
+//    waitSet = dds::core::cond::WaitSet();
+//    auto statusCondition = dds::core::cond::StatusCondition(reader);
+//    statusCondition.enabled_statuses(dds::core::status::StatusMask::none());
+//    waitSet.attach_condition(statusCondition);
+//
+//    // close the read condition while waiting
+//    actionThreadArgs.action = TA_CLOSE_STATUS_CONDITION;
```
sumanth-nirmal, line 989: Closing a status condition will make it unusable, so commented this block until its fixed

src/ddscxx/src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
```diff
@@ -39,17 +39,47 @@ org::eclipse::cyclonedds::core::cond::ConditionDelegate::~ConditionDelegate()
...
+void
+org::eclipse::cyclonedds::core::cond::ConditionDelegate::detach_and_close(
+    const dds_entity_t entity_handle)
+{
+    detach_from_waitset(entity_handle);
+    org::eclipse::cyclonedds::core::cond::ConditionDelegate::close();
+}
+
 void
 org::eclipse::cyclonedds::core::cond::ConditionDelegate::close()
```
sumanth-nirmal, line 75: The reason for adding a new API `detach_and_close()` is to get the entity handle without calling `get_ddsc_entity()` as it will lead to a double lock, and can't modify the `ConditionDelegate::close()` with some default parameter as it will hide the virtual interface function.

sumanth-nirmal:
@eboasson @reicheratwork The issue with conditions not getting correctly detached (or not possible to even detach) was causing some inconvenient issues (or crashes). \
So I took some inspiration from OpenSplice and tried to fix the issue with the waitset when conditions are closed or cleaned up. Currently, each condition has a list of waitsets to which it's connected and when the condition is cleaned up, it is detached from any waitset to which it is connected and then the condition is closed. However, the status condition is still special as it can not be really closed, so currently it will only be detached from any waitset when it is cloeanup, but not really closed (which is also the behavior earlier).

There are some locks introduced which should only affect the performance when attaching/detaching the conditions (and this will ideally happen only during the init/shutdown state of the application). 

I hope I didn't make the behaviour worse, but it's at least a step in the correct direction. Can you please review the changes?

src/ddscxx/src/org/eclipse/cyclonedds/core/cond/ConditionDelegate.cpp
eboasson: Perfectly sensible. I suspect there is still a window for a race condition where one can sneak in an `attach` in between copying the list of waitsets and looping over them. Given the set of problems this PR addresses and the obvious API abuse involved, I can't be too fussed about not closing that tiny hole in this PR.

src/ddscxx/src/org/eclipse/cyclonedds/core/cond/StatusConditionDelegate.cpp
```diff
@@ -53,7 +53,15 @@ org::eclipse::cyclonedds::core::cond::StatusConditionDelegate::close()
 {
     /* todo: cannot close this->ddsc_entity because this is the parent EntityDelegate;
```
eboasson, line 54: I am quite surpirsed that a `StatusCondition` can be closed at all. In the spec each Entity *has* a StatusCondition, so what could be the possible meaning of closing it? Maybe I misremembering or overlooking something in the spec, if so, perhaps @e-hndrks will correct me.

src/ddscxx/src/org/eclipse/cyclonedds/core/cond/WaitSetDelegate.cpp
eboasson: You could also add one more while waiting, but too it is not a problem: it simply means that the call to `wait` can return that more conditions triggered then it could return in the `attach` array. That's supported in the C code, and the code below allows for it as well (line 91 in the new line numbering).

src/ddscxx/tests/WaitSet.cpp
eboasson: As closing a StatusCondition is dubious in the first place, I'm fine with this approach.

eboasson (approved):
Wow, thanks @sumanth-nirmal! I'm really happy when someone helps out with the C++ API 😀

src/ddscxx/src/org/eclipse/cyclonedds/core/cond/StatusConditionDelegate.cpp
sumanth-nirmal: @eboasson I am also confused with this and wondering what is the correct behavior. [DDS C++ PSM](https://www.omg.org/spec/DDS-PSM-Cxx/1.0/PDF) section 7.5.1.2 says "Users of this PSM are recommended to call close on objects of all reference types once they are finished using them", at least from this it might be possible to close even status condition, or ..?

@e-hndrks

Issue #308, opened 2022-08-25
v0.10.1 compile error on vs2017


trittsv:
Hey, there is a compile error on vs2017 in the new CycloneDDS v0.10.1 release.

```
[2022-08-25T05:13:11.071Z] -- The C compiler identification is MSVC 19.16.27034.0
[2022-08-25T05:13:12.002Z] -- The CXX compiler identification is MSVC 19.16.27034.0
[2022-08-25T05:13:12.259Z] -- Detecting C compiler ABI info
[2022-08-25T05:13:13.630Z] -- Detecting C compiler ABI info - done
[2022-08-25T05:13:13.630Z] -- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Professional/VC/Tools/MSVC/14.16.27023/bin/Hostx64/x86/cl.exe - skipped
[2022-08-25T05:13:13.630Z] -- Detecting C compile features
[2022-08-25T05:13:13.630Z] -- Detecting C compile features - done
[2022-08-25T05:13:13.630Z] -- Detecting CXX compiler ABI info
[2022-08-25T05:13:15.524Z] -- Detecting CXX compiler ABI info - done
[2022-08-25T05:13:15.524Z] -- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Professional/VC/Tools/MSVC/14.16.27023/bin/Hostx64/x86/cl.exe - skipped
[2022-08-25T05:13:15.524Z] -- Detecting CXX compile features
[2022-08-25T05:13:15.525Z] -- Detecting CXX compile features - done
[2022-08-25T05:13:15.525Z] -- Compiling with type discovery support
[2022-08-25T05:13:15.525Z] -- Compiling with topic discovery support
[2022-08-25T05:13:15.525Z] -- Performing Test COMPILER_HAS_DEPRECATED_ATTR
[2022-08-25T05:13:16.454Z] -- Performing Test COMPILER_HAS_DEPRECATED_ATTR - Failed
[2022-08-25T05:13:16.454Z] -- Performing Test COMPILER_HAS_DEPRECATED
[2022-08-25T05:13:17.823Z] -- Performing Test COMPILER_HAS_DEPRECATED - Success
[2022-08-25T05:13:18.080Z] -- Configuring done
[2022-08-25T05:13:18.644Z] -- Generating done
[2022-08-25T05:13:18.644Z] CMake Warning:
[...]

PR #310, opened 2022-08-25
Fix for issue #308

reicheratwork:
The include for the C++ string library was missing, though this was not causing
issues in other compilation scenarioes, the order of compilation or parsing of
includes with Visual Studio 2017 caused this issue to be exposed


eboasson (approved):
LGTM, thanks @reicheratwork !

PR #312, opened 2022-08-26
Free type map/info only if successfully generated

eboasson:
On failure generating the type info and type map, the code using these
blobs when writing out the traits would still free the typemap and
typeinfo blobs, even though the contents of the memory would be
undefined.  This changes the error handling so this is done only iff the
type info and type map were generated correctly.


reicheratwork:
LGTM!

Though this needs the changes proposed in [this PR](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1389) to be merged as well to cover all possible paths.

PR #322, opened 2022-09-15
Feature/optional ddscxx

trittsv:
This makes the build of `DDSCXX` optional.
Useful in cross-compiling scenarios where you only need the generator and use the DDS lib from another build.


eboasson (approved):
I can imagine this being useful and I can't think of any downsides. @thijsmie do you agree?

thijsmie:
I think it is also okay to merge this now, although my mind immediately goes to "but then everything should be modularized and optional with standardized cmake variables!". I think it is wise to limit ourselves with what is actually useful for people so lets stick with this for now 😉

PR #325, opened 2022-09-27
Jira cp28

reicheratwork:
This PR will implement 2 example programs from CycloneDDS-C into C++:
- roundtrip
- throughput

### DIFF
--- a/examples/roundtrip/ping.cpp
+++ b/examples/roundtrip/ping.cpp
@@ -1,0 +1,296 @@
+/**
+*  Copyright(c) 2022 ZettaScale Technology and others
+*
+*   This program and the accompanying materials are made available under the
+*   terms of the Eclipse Public License v. 2.0 which is available at
+*   http://www.eclipse.org/legal/epl-2.0, or the Eclipse Distribution License
+*   v. 1.0 which is available at
+*   http://www.eclipse.org/org/documents/edl-v10.php.
+*
+*   SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+*/
+
+#include <vector>
+#include <chrono>
+#include <algorithm>
+#include <iomanip>
+#include <iostream>
+#include <thread>
+#include <string>
+
+#include "dds/dds.hpp"
+#include "dds/dds.h"
+#include "RoundTrip.hpp"
+#include "roundtrip_common.hpp"
+
+#define US_IN_ONE_SEC 1000000LL
+
+using times = std::vector<dds_time_t>;
+
+static times
+  roundTrip,
+  writeAccess,
+  readAccess,
+  roundTripOverall,
+  writeAccessOverall,
+  readAccessOverall;
+
+static dds_time_t
+  startTime,
+  preWriteTime,
+  postWriteTime,
+  preTakeTime,
+  postTakeTime;
+
+static bool warmUp = true;
+
+static RoundTripModule::DataType roundtrip_msg;
+
+static unsigned long payloadSize = 0,
+                     elapsed = 0,
+                     numSamples = 0;
+static const size_t column_width = 8;
+
+static void print_stats(times &to_print)
+{
+  std::sort(to_print.begin(), to_print.end());
+
+  std::cout << std::setw(column_width) << to_print[to_print.size()/2] << std::setw(column_width) << to_print.front() << std::setw(column_width) << to_print[static_cast<size_t>(static_cast<double>(to_print.size())*0.99)] << std::setw(column_width) << to_print.back() << std::flush;
+
+  to_print.clear();
+}
+
+static void print_all(times &rt, times &wa, times &ra)
+{
+    std::cout << std::setw(column_width) << rt.size() << " | ";
+    print_stats(rt);
+    std::cout << " | ";
+    print_stats(wa);
+    std::cout << " | ";
+    print_stats(ra);
+    std::cout << std::endl << std::flush;
+}
+
+static void print_header()
+{
+  std::cout <<
+    "# Warm up complete.\n"
+    "\n"
+    "#                 |            Latency [us]          |        Write-access time [us]    |     Read-access time [us]\n"
+    "#    N |    Count |   median   min     99%     max   |   median   min     99%     max   |   median   min     99%     max\n" << std::flush;
+}
+
+static void print_usage(void)
+{
+  std::cout <<
+    "Usage (parameters must be supplied in order):\n"
+    "./cxxRoundtripPong [-l/-h] [payloadSize (bytes, 0 - 100M)] [numSamples (0 = infinite)] [timeOut (seconds, 0 = infinite)]\n"
+    /*"./cxxRoundtripPong quit - ping sends a quit signal to pong.\n"*/
+    "Defaults:\n"
+    "./cxxRoundtripPong 0 0 0\n" << std::flush;
+  exit(EXIT_FAILURE);
+}
+
+template<typename T>
+bool data_available(dds::sub::DataReader<T>& rd, dds::pub::DataWriter<T>& wr)
+{
+  if (done)
+    return true;
+
+  dds_time_t difference = 0;
+
+  /* Take sample and check that it is valid */
+  try {
+    preTakeTime = dds_time ();
+    auto samples = rd.take();
+    postTakeTime = dds_time ();
+
+    if (samples.length() == 0)
+      return false;
+
+    auto info = samples.begin()->info();
+    if (info.valid() == false)
+      return false;
+    else
+      timedOut = false;
+
+    if (warmUp) {
+      if ((postTakeTime - startTime) > DDS_SECS(5)) {
+        warmUp = false;
+        startTime = dds_time ();
+        print_header();
+      }
+    } else {
+      difference = (postWriteTime - preWriteTime)/DDS_NSECS_IN_USEC;
+      writeAccess.push_back(difference);
+      writeAccessOverall.push_back(difference);
+
+      difference = (postTakeTime - preTakeTime)/DDS_NSECS_IN_USEC;
+      readAccess.push_back(difference);
+      readAccessOverall.push_back(difference);
+
+      difference = (postTakeTime - info.timestamp().nanosec() - info.timestamp().sec()*DDS_NSECS_IN_SEC)/(2*DDS_NSECS_IN_USEC);
+      roundTrip.push_back(difference);
+      roundTripOverall.push_back(difference);
+
+      /* Print stats each second */
+      if ((postTakeTime - startTime)/DDS_NSECS_IN_SEC > 1 && roundTrip.size() > 0)
+      {
+
+        std::cout << "# " << std::setw(4) << ++elapsed << " | "  << std::flush;
+        print_all(roundTrip, writeAccess, readAccess);
+
+        startTime = dds_time ();
+      }
+    }
+
+    preWriteTime = dds_time();
+    wr.write(roundtrip_msg, dds::core::Time(preWriteTime/DDS_NSECS_IN_SEC, static_cast<uint32_t>(preWriteTime%DDS_NSECS_IN_SEC)));
+    postWriteTime = dds_time();
+  } catch (const dds::core::TimeoutError &) {
+    std::cout << "Timeout encountered.\n" << std::flush;
+    return false;
+  } catch (const dds::core::Exception &e) {
+    std::cout << "Error: \"" << e.what() << "\".\n" << std::flush;
+    return false;
+  } catch (...) {
+    std::cout << "Generic error.\n" << std::flush;
+    return false;
+  }
+
+  return true;
+}
+
+static bool parse_args(int argc, char *argv[])
+{
+  if (argc >= 2)
+  {
+    if (0 == strcmp(argv[1], "-l"))
+      use_listener = true;
+    else if (0 == strcmp(argv[1], "-h"))
+      return false;
+  }
+
+  try {
+    for (int a = 1 + (use_listener ? 1 : 0); a < argc; a++) {
+      switch (a - (use_listener ? 1 : 0)) {
+        case 3:
+          timeOut = std::stoul(argv[a]);
+          break;
+        case 2:
+          numSamples = std::stoul(argv[a]);
+          break;
+        case 1:
+          payloadSize = std::stoul(argv[a]);
+          break;
+        default:
+          std::cout << "invalid number of command line parameters supplied\n" << std::flush;
+          return false;
+      }
+    }
+  } catch (...) {
+    std::cout << "invalid command line parameter supplied\n" << std::flush;
+    return false;
+  }
+
+  std::cout << "# payloadSize: " << payloadSize << " | numSamples: " << numSamples << " | timeOut: " << timeOut  << " | using " << (use_listener ? "listener" : "waitset") << " method\n" << std::flush;
+
+  return true;
+}
+
+int main (int argc, char *argv[])
+{
+  /* Register handler for Ctrl-C */
+  signal (SIGINT, sigint);
+
+  if (!parse_args(argc, argv)) {
+    print_usage();
+    return EXIT_FAILURE;
+  }
+
+  dds::domain::DomainParticipant participant(domain::default_id());
+
+  dds::topic::Topic<RoundTripModule::DataType> topic(participant, "RoundTrip");
+
+  dds::pub::qos::PublisherQos pqos;
+  pqos.policy(dds::core::policy::Partition("ping"));
+
+  dds::pub::Publisher publisher(participant, pqos);
+
+  dds::pub::qos::DataWriterQos wqos;
+  wqos.policy(
+    dds::core::policy::Reliability(
+      dds::core::policy::ReliabilityKind::Type::RELIABLE,
+      dds::core::Duration::from_secs(10)));
+  wqos.policy(
+    dds::core::policy::WriterDataLifecycle(true));
+
+  dds::pub::DataWriter<RoundTripModule::DataType> writer(publisher, topic, wqos);
+
+  dds::sub::qos::SubscriberQos sqos;
+  sqos.policy(dds::core::policy::Partition("pong"));
+
+  dds::sub::Subscriber subscriber(participant, sqos);
+
+  dds::sub::qos::DataReaderQos rqos;
+  rqos.policy(
+    dds::core::policy::Reliability(
+      dds::core::policy::ReliabilityKind::Type::RELIABLE,
+      dds::core::Duration::from_secs(10)));
+
+  RoundTripListener<RoundTripModule::DataType> list(writer, &data_available<RoundTripModule::DataType>);
+
+  dds::sub::DataReader<RoundTripModule::DataType>
+    reader(
+      subscriber,
+      topic,
+      rqos,
+      use_listener ? &list : NULL,
+      use_listener ? dds::core::status::StatusMask::data_available() : dds::core::status::StatusMask::none());
+
+  dds::core::Duration waittime = dds::core::Duration::from_secs(static_cast<double>(timeOut ? timeOut : 5));
+  if (!match_readers_and_writers(reader, writer, waittime))
+    return EXIT_FAILURE;
+
+  roundtrip_msg.payload(std::vector<uint8_t>(payloadSize, 'a'));
+
+  startTime = dds_time ();
+  preWriteTime = dds_time ();
+  printf ("# Waiting for startup jitter to stabilise\n");
+
+  //starting write
+  if (use_listener) {
+    writer.write(roundtrip_msg);
+    while (!timedOut && !done && (!numSamples || elapsed < numSamples)) { 
+      timedOut = true;
+      std::this_thread::sleep_for(std::chrono::seconds(timeOut ? timeOut : 1));
+    }
+  } else {
+    dds::core::cond::WaitSet waitset;
+
+    dds::core::cond::StatusCondition rsc(reader);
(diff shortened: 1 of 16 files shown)


examples/throughput/subscriber.cpp
```diff
@@ -0,0 +1,323 @@
...
+    pollingDelay = std::chrono::milliseconds(atoi(argv[2])); /* The number of ms to wait between reads (0 = waitset, -1 = listener) */
+  }
+  if (argc > 3)
+  {
+    partitionName = argv[3]; /* The name of the partition */
+  }
+  return EXIT_SUCCESS;
+}
+
+unsigned long long do_take(dds::sub::DataReader<ThroughputModule::DataType>& rd)
```
poetinger, line 85: We don't need to return anything since it is tracked via global.

examples/throughput/subscriber.cpp
```diff
@@ -0,0 +1,323 @@
...
+    if (!s.info().valid())
+      continue;
+
+    auto ih = s.info().publication_handle();
+    auto ct = s.data().count();
+    auto it = mp.find(ih);
+    /*check whether the received sequence number matches that which we expect*/
+    if (it != mp.end() &&
+        it->second != ct)
+      outOfOrder++;
```
poetinger, line 100: Is the intention to show out of order samples as a rolling total? Or should it get reset between batches of samples?

poetinger (approved):
I only found a couple nitpicking things that don't really need to change. I ran each example on a Mac as well, and all seemed to function as intended. I think this is okay to merge, but we need the final E-proval.

poetinger:
After speaking to @reicheratwork I found out I had not tested the waist functionality. It worked fine for me, but maybe that should be added to the documentation and '-h' info.

examples/roundtrip/ping.cpp
```diff
@@ -0,0 +1,294 @@
...
+  std::cout <<
+    "Usage (parameters must be supplied in order):\n"
+    "./cxxRoundtripPong [-l/-h] [payloadSize (bytes, 0 - 100M)] [numSamples (0 = infinite)] [timeOut (seconds, 0 = infinite)]\n"
+    /*"./cxxRoundtripPong quit - ping sends a quit signal to pong.\n"*/
+    "Defaults:\n"
+    "./cxxRoundtripPong 0 0 0\n" << std::flush;
+  exit(EXIT_FAILURE);
+}
+
+static bool data_available(dds::sub::DataReader<RoundTripModule::DataType>& rd, dds::pub::DataWriter<RoundTripModule::DataType>& wr)
```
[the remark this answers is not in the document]
reicheratwork: Is a good idea, I will do this.

examples/throughput/subscriber.cpp
reicheratwork: It was implemented this way because the C throughput example also shows cumulative received samples, and I thought it would be better to have the C++ throughput example to mirror this approach and head off possible questions about why C and C++ give different output :)

hansvanthag: I'd agree with keeping it the same as the C example (and with that also with the OSPL-alternatives as those examples are also used 'in combinations' to compare both DDS implementations)

examples/throughput/subscriber.cpp
reicheratwork: It is actually used in the polling method in the function `process_samples` to see whether we received messages in the interval, so we still need this return value.

examples/throughput/subscriber.cpp
poetinger: I'm all for keeping things consistent. But the expected behaviour would be per batch (especially when it is part of a function that operates per batch), and not a rolling total that needs to be broken apart to see the numbers per batch.

If it was a tally of the out of order *batches* then it makes sense, but this is a count of the out of order *samples* in a single batch. If you miss any samples, it will look like you miss the same number of samples for every batch following that batch.

examples/throughput/subscriber.cpp
poetinger: Ah yes, I see now. I missed that in the browser search function. You mean this:
```
while (do_take (reader))
         ;
```
Nevermind then :)

examples/roundtrip/ping.cpp
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+
+  if (!warmUp)
+  {
+    std::cout << "# Overall" << std::flush;
+    print_all(roundTripOverall, writeAccessOverall, readAccessOverall);
+  }
+
+  //send quit?
```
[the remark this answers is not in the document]
reicheratwork: This is a piece of functionality that is in the C example, but still missing in this C++ example.
I will implement this.

examples/throughput/subscriber.cpp
hansvanthag: ok

examples/throughput/subscriber.cpp
```diff
@@ -0,0 +1,327 @@
...
+      const dds::core::status::LivelinessChangedStatus&) { }
+
+  void on_subscription_matched(
+      dds::sub::DataReader<T>&,
+      const dds::core::status::SubscriptionMatchedStatus&) { }
+
+  void on_sample_lost(
+      dds::sub::DataReader<T>&,
+      const dds::core::status::SampleLostStatus&) { }
+};
```
e-hndrks, line 267: Instead of extending from the DataReaderListener, where all methods need to be overridden, why not extend from NoOpDataReaderListener, where only the one that you are going to implement needs to be overridden?

examples/throughput/subscriber.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+  if (wait_for_writer (reader))
+  {
+    std::cout << subprefix << "Waiting for samples...\n" << std::flush;
+
+    process_samples(reader, maxCycles);
+  }
+
+  return EXIT_SUCCESS;
+}
```
e-hndrks, line 323: Basically all the comments I gave for the Publisher regarding the use of command line parameters and the use of Qos also apply to the Subscriber.

examples/throughput/subscriber.cpp
```diff
@@ -0,0 +1,327 @@
...
+      dds::core::policy::ReliabilityKind::Type::RELIABLE,
+      dds::core::Duration::from_secs(10)));
+  rqos.policy(
+    dds::core::policy::History(
+      dds::core::policy::HistoryKind::Type::KEEP_ALL,
+      0));
+  rqos.policy(
+    dds::core::policy::ResourceLimits(MAX_SAMPLES));
+
+  ThroughputListener<ThroughputModule::DataType> list;
```
e-hndrks: The name "list" for a listener might be a bit confusing. Ususally you mean something else when referring to a list.

examples/throughput/subscriber.cpp
```diff
@@ -0,0 +1,327 @@
...
+    auto ct = s.data().count();
+    auto it = mp.find(ih);
+    /*check whether the received sequence number matches that which we expect*/
+    if (it != mp.end() &&
+        it->second != ct)
+      outOfOrder++;
+
+    valid_samples++;
+    total_bytes += s.data().payload().size();
+    mp[ih] = ct+1;
```
e-hndrks, line 104: Why store the instance [handle, count] pair in a map? Given that this is a keyless topic, there is going to be only one instance and therefore only one instance handle. Why not just have a global variable that holds the count and compare that the count() field of the current sample? Now the code seems pretty complex for no apparent reason.

examples/throughput/subscriber.cpp
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    if (reader.subscription_matched_status().current_count() == 0)
+      done = true;
+
+    for (const auto &c:conditions)
+    {
+      if (c == sc)
+      {
+        wait_again = false;
```
e-hndrks, line 156: Am I correct in stating that if you blocked on the WaitSet, and it triggers, then you leave the loop? In other words, you never block more than once for incoming data? Why is that?

e-hndrks (changes requested):
Review comments can be found in the code sections to which they apply. Overall the code looks nice, but I would suggest some changes to simplify the code where possible.

examples/throughput/publisher.cpp
```diff
@@ -0,0 +1,223 @@
...
+    burstInterval = std::chrono::milliseconds(atoi (argv[2])); /* The time interval between each burst in ms */
+
+  if (argc > 3)
+    burstSize = static_cast<uint32_t>(atoi (argv[3])); /* The number of samples to send each burst */
+
+  if (argc > 4)
+    timeOut = std::chrono::seconds(atoi (argv[4])); /* The number of seconds the publisher should run for (0 = infinite) */
+
+  if (argc > 5)
+    partitionName = argv[5]; /* The name of the partition */
```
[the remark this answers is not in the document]
poetinger: I used getopt for the omg interopt testing. But that is hidden away, and there are many more arguments to parse (of different types). I think for such a small example (a few numbers), this style is okay. I don't see much added value to having a full fledge CLI here.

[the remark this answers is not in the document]
reicheratwork: I did not use an option list compared to fixed parameter ordering because my approach was to have the C++ implementation be completely exchangeable with the C implementation. So that means no option list as command line parameters.

examples/throughput/publisher.cpp
```diff
@@ -0,0 +1,223 @@
...
+        reportstart = n;
+      }
+
+    if (writer.publication_matched_status().current_count() == 0)
+      timedOut = true;
+    }
+
+    std::cout << "\n" << pubprefix << (done ? "Terminated" : "Timed out") << ", " << sample.count() << " samples written.\n" << std::flush;
+  }
+  /*wr.write_flush();  currently the C++ binding does not support batched writing*/
```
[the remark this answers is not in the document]
reicheratwork: I don't completely understand what you mean here, as on both occasions write_flush would be called, it is commented out (publisher.cpp lines 128,156,175).

As for the reason for that comment, I was planning to have that there as a reminder to enable this when the batched writing interface is implemented on C++.

examples/throughput/publisher.cpp
```diff
@@ -0,0 +1,223 @@
...
+{
+  ThroughputModule::DataType sample;
+
+  if (parse_args(argc, argv) == EXIT_FAILURE) {
+    return EXIT_FAILURE;
+  }
+
+  dds::domain::DomainParticipant participant(domain::default_id());
+
+  dds::topic::Topic<ThroughputModule::DataType> topic(participant, "Throughput");
```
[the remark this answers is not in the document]
reicheratwork: I chose this because this completely mirrors the QoS settings done in the C throughput examples, where the topic has the default QoS set, while the writer has the explicit QoS settings.
These choices for QoS etc. make the C and C++ examples interchangeable (e.g. having a C++ throughput publisher and a C subscriber).

examples/throughput/subscriber.cpp
reicheratwork: Good suggestion.

examples/throughput/subscriber.cpp
reicheratwork: The choices in the publisher and subscriber regarding the QoS settings and the command line parameters were both done to make the C and C++ examples as interchangeable as possible.

examples/throughput/subscriber.cpp
reicheratwork: Maybe abbreviated the variable name a little too much here :)
I will look at all variable names and make them a little more explicit.

examples/throughput/subscriber.cpp
reicheratwork: It is not the handle of the handle of the instance being stored, but the handle to the writer, to be able to distinguish between multiple writers.

examples/roundtrip/roundtrip_common.hpp
```diff
@@ -0,0 +1,98 @@
...
+    waitset.attach_condition(rsc);
+    waitset.wait(timeout);
+  } catch (const dds::core::TimeoutError &) {
+    return false;
+  }
+  return true;
+}
+
+template<typename T>
+class RoundTripListener: public dds::sub::DataReaderListener<T>
```
e-hndrks: Again, better to extend from NoOpDataReaderListener here,

e-hndrks (changes requested):
Generally looks good to me. Two small suggestions.

examples/throughput/publisher.cpp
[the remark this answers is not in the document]
e-hndrks: Seems like a reason to modify the C example as well in that case. If our examples are to show best-practices, then we shouldn't start by not setting the Qos on the topics ourselves.

examples/throughput/publisher.cpp
```diff
@@ -0,0 +1,223 @@
...
+  wqos.policy(
+    dds::core::policy::Reliability(
+      dds::core::policy::ReliabilityKind::Type::RELIABLE,
+      dds::core::Duration::from_secs(10)));
+  wqos.policy(
+    dds::core::policy::History(
+      dds::core::policy::HistoryKind::Type::KEEP_ALL,
+      0));
+  wqos.policy(
+    dds::core::policy::ResourceLimits(MAX_SAMPLES));
```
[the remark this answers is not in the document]
e-hndrks: Still, you are using the constructors of the policies, which requires a lot of code. The static utility functions are a much shorter (and more often used) way to do that. (See the example I used on the Topic).

examples/throughput/subscriber.cpp
e-hndrks: Ah, you are right: it is the publisher handle, not the instance handle. Missed that one. In that case in makes totally sense what you are doing.....

examples/roundtrip/roundtrip_common.hpp
reicheratwork: Yep, good suggestion.

examples/roundtrip/roundtrip_common.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+template<typename T>
+bool match_readers_and_writers(dds::sub::DataReader<T> &rd, dds::pub::DataWriter<T> &wr, dds::core::Duration &timeout)
+{
+  dds::core::cond::WaitSet waitset;
+
+  dds::core::cond::StatusCondition wsc(wr), rsc(rd);
+  wsc.enabled_statuses(dds::core::status::StatusMask::publication_matched());
+  rsc.enabled_statuses(dds::core::status::StatusMask::subscription_matched());
+
```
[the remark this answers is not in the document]
reicheratwork: Let me look a little bit deeper into this.

examples/throughput/publisher.cpp
```diff
@@ -0,0 +1,223 @@
...
+      {
+        timedOut = true;
+      } else if (reportstart + std::chrono::seconds(1) < n) {
+        std::cout << "." << std::flush;
+        reportstart = n;
+      }
+
+    if (writer.publication_matched_status().current_count() == 0)
+      timedOut = true;
+    }
```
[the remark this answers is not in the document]
reicheratwork: Let me look deeper into this.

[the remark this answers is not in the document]
reicheratwork: I will make this give more descriptive feedback to the end user.

examples/throughput/subscriber.cpp
reicheratwork: You are correct, the trigger value on the status conditions being checked is missing here.
I will check this.

examples/throughput/publisher.cpp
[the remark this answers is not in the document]
reicheratwork: Let me rewrite this in with the streaming operators, as that approach is more compact (and legible)

reicheratwork:
After some discussion with @e-hndrks it was decided to do the restructuring of the QoS assignments.
There will also be another PR mirroring these changes in the CycloneDDS examples.

e-hndrks (approved):
This looks good to me. I have no further comments.

e-hndrks (commented):
Making the default policies of the Reader/WriterQos derive from the overlapping TopicQos is something that already happens in the C API of Cyclone, but is not something that is currently in the DDS specification for the C++ API. It is something I would like to see in the DDS spec though, and we might want to submit a proposal for that purpose. Question is: do we want to make examples that already anticipate on such a change, and have this be an example that will not run on other implementations of the C++ API, or do we want to restrict ourselves to the C_++ API to that is in the current spec?  If it is the first, then I see no other issues in this commit. If it is the latter, then we might need to explicitly derive the Reader/WriterQos from the TopicQos by having a statement like:
dds::sub::qos::DataReaderQos rqos = tqos;
and pass rqos to the constructor of the DataReader. 
(And likewise for the DataWriter of course).

reicheratwork:
> Making the default policies of the Reader/WriterQos derive from the overlapping TopicQos is something that already happens in the C API of Cyclone, but is not something that is currently in the DDS specification for the C++ API. It is something I would like to see in the DDS spec though

You are right in this, though we should probably discuss this with @eboasson as well before coming to a conclusion.

examples/roundtrip/ping.cpp
```diff
@@ -169,10 +171,14 @@ static bool parse_args(int argc, char *argv[])
 {
   if (argc >= 2)
   {
-    if (0 == strcmp(argv[1], "-l"))
+    if (0 == strcmp(argv[1], "-l")) {
       use_listener = true;
-    else if (0 == strcmp(argv[1], "-h"))
+    } else if (0 == strcmp(argv[1], "-h")) {
```
e-hndrks, line 176: I guess the -h option is supposed to print help, right? Then why does specifying this option result in the application exiting with EXIT_FAILURE? I can imagine it does that when specifying the wrong parameters, but asking for help does not seem a reason to report a FAILURE to me....

examples/roundtrip/ping.cpp
```diff
@@ -250,18 +256,23 @@ int main (int argc, char *argv[])
       use_listener ? dds::core::status::StatusMask::data_available() : dds::core::status::StatusMask::none());
 
   dds::core::Duration waittime = dds::core::Duration::from_secs(static_cast<double>(timeOut ? timeOut : 5));
-  if (!match_readers_and_writers(reader, writer, waittime))
+  if (sendTerm) {
+    std::cout << "sending termination command to pong program\n" << std::flush;
+    writer.dispose_instance(roundtrip_msg);
     return EXIT_FAILURE;
```
e-hndrks, line 257: Likewise, why is sending a TERM sample reason to terminate the application with EXIT_FAILURE?

e-hndrks (changes requested):

examples/roundtrip/roundtrip_common.hpp
```diff
@@ -69,6 +69,12 @@ class RoundTripListener: public dds::sub::NoOpDataReaderListener<RoundTripModule
     (void)_f(rd, _wr);
   }
 
+  void on_liveliness_changed(dds::sub::DataReader<RoundTripModule::DataType>&,
+        const dds::core::status::LivelinessChangedStatus& status) {
+    if (status.not_alive_count_change() > 0)
+      done = true;
+  }
+
```
e-hndrks, line 77: Why does the C++ API use a different condition to terminate then the C API? It seems that the C API terminates when the data gets disposed, where the C++ API terminates when the Writer disappears. Also in C the WriterDataLifcycle has its autodispose field set to FALSE, whereas in C++ it is set to TRUE. This means that in C you can instruct a Writer to write 10 samples of 10 bytes, gracefully exiting when done , and then restart it to write 10 additional samples of let's say 100 bytes, and continue doing this for ever increasing sizes. When done, you instruct ping to send a TERM message, in which case it disposes the data, after which pong will terminate itself gracefully. Up to that time, the pong application will keep on working to bounce back the samples produced by ping. In contrast, the C++, pong will terminate as soon as ping destructs its Writer, which means for the same scenario I not only need to restart ping, but restart pong as well. 
I would therefore like to propose the following:
- Set the autodispose field of WriterDataLifecycle to FALSE in C++.
- Remove the on_liveliness_changed handler in C++.
- When consuming samples, check its lifecycle and exit when it becomes DISPOSED.

reicheratwork: I will fix this

e-hndrks (approved):
This looks good to me: I have no further comments.

eboasson:
There is strange build failure on one platform because of `gcc` getting killed. Surely that's not a bug on our side, and I suspect it is simply running out of memory. However, merging code that breaks the builds is not such a great plan.

One option could be to just avoid the parallel build. I hate to suggest it, but it is either that or trying to understand why this code causes it to blow up.

Or maybe the problem goes away by switching to gcc 12 ...

reicheratwork:
The issue with the gcc build has been occurring on and off for the last few weeks...
I think if we try this with no parallel builds and see what happens then, we could give this a green light?

examples/throughput/publisher.cpp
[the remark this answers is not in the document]
poetinger: I couldn't find any issues related to the implementation of batched writing. Maybe we should create one and tie the comment to it (e.g. `TODO #00000 Uncomment when complete`).

examples/throughput/publisher.cpp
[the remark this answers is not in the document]
poetinger: @e-hndrks I'll take a look at the C examples and update them if needed so they match this style.

examples/throughput/publisher.cpp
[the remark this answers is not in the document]
reicheratwork: I will make this issue.

PR #330, opened 2022-10-10
Fix bug in copying of GenerationCount and Rank in SampleInfo

reicheratwork:
Setting the GenerationCount and Rank fields of the SampleInfo through the delegate did not copy the contents correctly, causing these fields to stay in the default constructed values (0)
Changed copying by calling the generation_count and rank setter functions
Added unittests to check for correct behaviour


eboasson (approved):
I'm glad you caught that one 🙂 Thanks!

Issue #329, opened 2022-10-06
xtypes is not working in our environment


javiersorianoruiz:
Hello everybody.

We (@Bit-Seq, @jordialexalar and I) have been trying examples with **XTypes and cyclonedds-cxx** and they don't work.

It works under Python (with cyclonedds-python).

But with cyclonedds-cxx it doesn't work for us. We don't know if we are doing something wrong.

We work with this IDL (I have a publisher and subscriber sending and receiving this message. We are using the helloword example provided by the same cyclonedds-cxx as base of our testing):

```
module HelloWorldData
{
  @mutable
  @autoid(HASH)
  struct Msg
  {
    @key long userID;
    string message;
  };
};
```

We have created a new publisher and subscriber with the previous message but modified (new variable `long extra_value;`)

```
module HelloWorldData
{
  @mutable
  @autoid(HASH)
  struct Msg
  {
    @key long userID;
    long extra_value;
    string message;
  };
};

```

They don't speak between them. 

That's to say, subscriber1 (original message) only receives messages from publisher1, and it doesn't receive any message from the publisher2 (modified message).

On the other hand, subscriber2 only receives messages from publisher2, and it doesn't receive any message from the publisher1.

We have checked with `cmake -LAH` both parameters that are indicated in the link [What syntax and datatypes are supported in IDL](https://cyclonedds.io/content/blog/xtypes-features.html):

```
ENABLE_TOPIC_DISCOVERY:BOOL=ON
ENABLE_TYPE_DISCOVERY:BOOL=ON
```

So, we think that everything is properly configured.

Are we doing something wrong in cyclonedds-cxx with XTypes or is there an actual problem with XTypes in C++?, because with cyclonedds-python a similar example worked well.

All the binaries, (publisher1,subscriber1, publisher2 and subscriber2) are launched on 4 running containers over the same docker image. So, configurations and settings are exactly the same for all the binaries.

The Dockerfile to reproduce the problem is this:

```
FROM        ubuntu:20.04

# update and install dependencies
RUN         apt-get update \
                && apt-get install -y \
                    software-properties-common \
                    wget \
                && add-apt-repository -y ppa:ubuntu-toolchain-r/test \
                && apt-get update \
                && apt-get install -y \
                    make \
                    git \
                    curl \
                    vim \
                && apt-get install -y cmake \
                && apt-get install -y gcc-11 \
                && apt-get install -y g++-11 \
[...]

PR #331, opened 2022-10-13
Fix issue #329

reicheratwork:
Issue #329 was caused by default constructed and unset QoS Policies being indistinguishable for CDDS-CXX, which caused issues when CDDS-C would fill in certain fields based on other information if the Policy wasn't set. Such DataRepresentation in issue #329.
The solution was to make the C++ QoS implementations mirror the C sparse QoSes.
Policies that are unset in a C++ QoS are now NOT exported to the C QoS representation, wherease they previously always were.
Also, the contents of default constructors reader/writer/topic/publisher/subscriber QoS are now taken from CDDS-C, in stead of being explicitly defined, removing a double definition. C++'s default QoSes will now mirror C's default QoSes.

@javiersorianoruiz If would be so kind as to help with the review process?

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
@@ -20,8 +20,6 @@
 
 #include <dds/core/policy/CorePolicy.hpp>
 
-#include <dds/ddsi/ddsi_xqos.h>
-
 struct _DDS_NamedDomainParticipantQos;
 
 namespace org
@@ -38,6 +36,8 @@ namespace qos
 class OMG_DDS_API DomainParticipantQosDelegate
 {
 public:
+    DomainParticipantQosDelegate();
+
     void policy(const dds::core::policy::UserData& ud);
     void policy(const dds::core::policy::EntityFactory& efp);
 
@@ -73,13 +73,10 @@ DomainParticipantQosDelegate::policy<dds::core::policy::UserData> () const
 {
     return user_data_;
 }
+
 template<>
-inline dds::core::policy::UserData&
-DomainParticipantQosDelegate::policy<dds::core::policy::UserData> ()
-{
-    present_ |= QP_USER_DATA;
-    return user_data_;
-}
+dds::core::policy::UserData&
+DomainParticipantQosDelegate::policy<dds::core::policy::UserData> ();
 
 
 template<>
@@ -88,13 +85,10 @@ DomainParticipantQosDelegate::policy<dds::core::policy::EntityFactory> () const
 {
     return entity_factory_;
 }
+
 template<>
-inline dds::core::policy::EntityFactory&
-DomainParticipantQosDelegate::policy<dds::core::policy::EntityFactory> ()
-{
-    present_ |= QP_ADLINK_ENTITY_FACTORY;
-    return entity_factory_;
-}
+dds::core::policy::EntityFactory&
+DomainParticipantQosDelegate::policy<dds::core::policy::EntityFactory> ();
 
 }
 }
--- a/src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
@@ -20,6 +20,8 @@
 
 #include <dds/core/policy/CorePolicy.hpp>
 
+#include <dds/ddsi/ddsi_xqos.h>
+
 struct _DDS_NamedDomainParticipantQos;
 
 namespace org
@@ -36,11 +38,6 @@ namespace qos
 class OMG_DDS_API DomainParticipantQosDelegate
 {
 public:
-    DomainParticipantQosDelegate();
-    DomainParticipantQosDelegate(const DomainParticipantQosDelegate& other);
-
-    ~DomainParticipantQosDelegate();
-
     void policy(const dds::core::policy::UserData& ud);
     void policy(const dds::core::policy::EntityFactory& efp);
 
@@ -56,11 +53,13 @@ public:
     void check() const;
 
     bool operator ==(const DomainParticipantQosDelegate& other) const;
-    DomainParticipantQosDelegate& operator =(const DomainParticipantQosDelegate& other);
 
+    const uint64_t &present() const {return present_;}
+    uint64_t &present() {return present_;}
 private:
-    dds::core::policy::UserData user_data_;
-    dds::core::policy::EntityFactory entity_factory_;
+    uint64_t                          present_ = 0;
+    dds::core::policy::UserData       user_data_;
+    dds::core::policy::EntityFactory  entity_factory_;
 };
 
 
@@ -78,6 +77,7 @@ template<>
 inline dds::core::policy::UserData&
 DomainParticipantQosDelegate::policy<dds::core::policy::UserData> ()
 {
+    present_ |= QP_USER_DATA;
     return user_data_;
 }
 
@@ -92,6 +92,7 @@ template<>
 inline dds::core::policy::EntityFactory&
 DomainParticipantQosDelegate::policy<dds::core::policy::EntityFactory> ()
 {
+    present_ |= QP_ADLINK_ENTITY_FACTORY;
     return entity_factory_;
 }
 
--- a/src/ddscxx/src/org/eclipse/cyclonedds/pub/qos/DataWriterQosDelegate.cpp
+++ b/src/ddscxx/src/org/eclipse/cyclonedds/pub/qos/DataWriterQosDelegate.cpp
@@ -20,6 +20,8 @@
 
 #include <cassert>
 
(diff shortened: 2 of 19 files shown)


src/ddscxx/src/org/eclipse/cyclonedds/pub/qos/DataWriterQosDelegate.cpp
```diff
@@ -33,98 +35,111 @@ namespace qos
 
 DataWriterQosDelegate::DataWriterQosDelegate()
 {
-    reliability_.kind(dds::core::policy::ReliabilityKind::RELIABLE);
+    ddsc_qos(&ddsi_default_qos_writer);
+    present() &= ~QP_DATA_REPRESENTATION;
+    check();
+    policy(dds::core::policy::Reliability(dds::core::policy::ReliabilityKind::RELIABLE));
```
eboasson, line 41: Isn't this covered by `ddsc_qos(&ddsi_default_qos_writer)`?

src/ddscxx/src/org/eclipse/cyclonedds/sub/qos/DataReaderQosDelegate.cpp
```diff
@@ -249,18 +301,29 @@ DataReaderQosDelegate::operator==(const DataReaderQosDelegate& other) const
...
-    durability_  = tqos.policy<dds::core::policy::Durability>();
-    deadline_    = tqos.policy<dds::core::policy::Deadline>();
-    budget_      = tqos.policy<dds::core::policy::LatencyBudget>();
-    liveliness_  = tqos.policy<dds::core::policy::Liveliness>();
-    reliability_ = tqos.policy<dds::core::policy::Reliability>();
-    order_       = tqos.policy<dds::core::policy::DestinationOrder>();
-    history_     = tqos.policy<dds::core::policy::History>();
-    resources_   = tqos.policy<dds::core::policy::ResourceLimits>();
-    ownership_   = tqos.policy<dds::core::policy::Ownership>();
+    if (tqos.present() & QP_DURABILITY)
```
eboasson, line 304: This should be looking at the `DURABILITY` flag for all settings

src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
```diff
@@ -20,6 +20,8 @@
 
 #include <dds/core/policy/CorePolicy.hpp>
 
+#include <dds/ddsi/ddsi_xqos.h>
```
eboasson, line 23: I suspect this header file gets included by the application code as well, and if so, then including this header file pollutes the name space rather badly. It isn't really part of the public API, but language bindings are a bit special, and this does save quite a few lines of boilerplate compared to calling `dds_qget_` functions all the time and checking the return value.

The only thing you actually use from it (unless I overlooked something) are the `QP_` flags, so a the name space pollution could be addressed by reorganising the C header files a bit, but I think I'd be happier if the C++ binding used its own set of (namespaced) constants, with a single `.cpp` file including the C header file using static assertions to verify that the bits are actually the same.

Best would of course be if there's a way of decoupling it completely without adding a lot of boilerplate.

src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
```diff
@@ -36,11 +38,6 @@ namespace qos
 class OMG_DDS_API DomainParticipantQosDelegate
 {
 public:
-    DomainParticipantQosDelegate();
```
eboasson, line 39: I surprises me you can just throw these away, are you sure that can't cause any problems?

src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
```diff
@@ -56,11 +53,13 @@ class OMG_DDS_API DomainParticipantQosDelegate
...
-    DomainParticipantQosDelegate& operator =(const DomainParticipantQosDelegate& other);
 
+    const uint64_t &present() const {return present_;}
+    uint64_t &present() {return present_;}
 private:
-    dds::core::policy::UserData user_data_;
-    dds::core::policy::EntityFactory entity_factory_;
+    uint64_t                          present_ = 0;
+    dds::core::policy::UserData       user_data_;
+    dds::core::policy::EntityFactory  entity_factory_;
```
eboasson, line 62: I tried to find out whether the default constructor for the various policies will initialise the value properly, but I am not quite sure that it really does: for example, `EntityFactory` seems to end up in `PolicyDelegate.hpp:222` with the default-default constructor, and IINM, that means the boolean in there won't be initialized. (Or perhaps that was for POD and classes are different?)

I don't think there have been any changes to this aspect, but I think it is worth double-checking.

src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
```diff
@@ -92,6 +92,7 @@ template<>
 inline dds::core::policy::EntityFactory&
 DomainParticipantQosDelegate::policy<dds::core::policy::EntityFactory> ()
 {
+    present_ |= QP_ADLINK_ENTITY_FACTORY;
```
eboasson, line 95: This I don't really understand: this means that reading the value marks it as being set, which in turn means it gets passed on to the C API, right? That may or may not be wise (I can see arguments either way), but having two getters and only one of them setting the flag seems like a recipe for confusion. Given that the `const` variant can't set the flag, I think that means the non-const version also mustn't set the flag.

reicheratwork: The idea here was that we are passing a reference to the EntityFactory QoS, so the person receiving the reference can modifies members of the EntityFactory QoS and hereby implicitly set the presence flag for the EntityFactory QoS.

src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
reicheratwork: Well, the copy constructor and destructor are in my eyes the definition of boilerplate, and can be completely handed off to the compiler.
By setting the default values of the members, which do not have default constructors (like the present_ flag on line 60), these values will then be passed on during construction when no other values are supplied.
This allows the user to skip the definition of the default constructor and just fall back on the default values supplied in the definition of the class.

src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
reicheratwork: I will double check this.

src/ddscxx/src/org/eclipse/cyclonedds/pub/qos/DataWriterQosDelegate.cpp
reicheratwork: Let me double check.

src/ddscxx/src/org/eclipse/cyclonedds/sub/qos/DataReaderQosDelegate.cpp
reicheratwork: OUCH!!!
I will fix this immediately!!!

src/ddscxx/src/org/eclipse/cyclonedds/pub/qos/DataWriterQosDelegate.cpp
reicheratwork: As we just use ddsc_qos, and not ddsi_xqos_mergein_missing, we do need to "unset" the DataRepresentation policy.
Setting of the Reliability policy is not necessary, as that set through ddsc_qos, i will remove this.

src/ddscxx/include/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.hpp
reicheratwork: The old iteration with the explicit default constructor also takes the default constructor for the EntityFactory QoSPolicy.
Maybe fix it at the default constructor for the QoSPolicy?

reicheratwork:
Updated the PR, but now it cannot be merged until [CycloneDDS-C PR 1503](https://github.com/eclipse-cyclonedds/cyclonedds/pull/1503) is merged, as it exposes the ``ddsi_default_qos_participant`` which this PR requires.

eboasson (approved):
LGTM! And thank you for updating it several times while I kept merging stuff into the core library breaking this several times ☺️

javiersorianoruiz:
I have downloaded and installed current master versions of cyclonedds, cyclonedds-python and cyclonedds-cxx and my XTypes test for Cpp and Python are working now properly.

More info about these tests:

https://github.com/javiersorianoruiz/cyclonedds-xtypes-testing

Issue #326, opened 2022-10-02
IDL generated get_type_props can lead to heap use after free


sumanth-nirmal:
## Required information

**Operating system:**
Ubuntu 20.04 LTS

**Cyclone Version**

- cyclonedds - master branch, https://github.com/eclipse-cyclonedds/cyclonedds/commit/51aa411347bb56d7e5c90f5bb62b9cb926e0feea
- cyclonedds-cxx - master branch, 3c5cdd31d760106a84c26afd927824c942d23897

**Compiler**

- gcc `gcc version 8.4.0 (Ubuntu 8.4.0-3ubuntu2) `

**Observed result or behavior:**

For a simple type as follows

```hpp
module rmw_dds_common {
module msg {
module dds_ {

struct Gid_ {
octet data_[24];

};


};  // module dds_
};  // module msg
};
```
the generated code for the `get_type_props` is as follows:

```hpp
template<>
propvec &get_type_props<::rmw_dds_common::msg::dds_::Gid_>() {
  static thread_local std::mutex mtx;
  static thread_local propvec props;
  static thread_local entity_properties_t *props_end = nullptr;
  static thread_local std::atomic_bool initialized {false};
  key_endpoint keylist;
  if (initialized.load(std::memory_order_relaxed)) {
    auto ptr = props.data();
    while (ptr < props_end)
      (ptr++)->is_present = false;
    return props;
  }
  std::lock_guard<std::mutex> lock(mtx);
  if (initialized.load(std::memory_order_relaxed)) {
    auto ptr = props.data();
    while (ptr < props_end)
      (ptr++)->is_present = false;
    return props;
  }
  props.clear();

  props.push_back(entity_properties_t(0, 0, false, bb_unset, extensibility::ext_final));  //root
  props.push_back(entity_properties_t(1, 0, false, get_bit_bound<uint8_t>(), extensibility::ext_final, false));  //::data_

  entity_properties_t::finish(props, keylist);
  props_end = props.data() + props.size();
  initialized.store(true, std::memory_order_release);
  return props;
}
```

[...]

PR #332, opened 2022-10-14
Fix for issue #326

reicheratwork:
The type-specific entity_properties are now completely static, no longer thread_local
The getters for the entity_properties are also all const now, as the presence of struct fields is logged inside the streaming functions, and no modification of the member fields is necessary

@sumanth-nirmal, I can count on you to do a review of this PR?

### DIFF
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -1129,7 +1129,8 @@ print_switchbox_open(struct streams *streams)
 
 static idl_retcode_t
 print_constructed_type_close(
-  struct streams *streams)
+  struct streams *streams,
+  const void* node)
 {
   const char *fmt =
     "  return streamer.finish_struct(*props, member_ids);\n"
@@ -1139,9 +1140,25 @@ print_constructed_type_close(
     "  initialized.store(true, std::memory_order_release);\n"
     "  return props;\n"
     "}\n\n";
+  static const char *pfmt2 =
+    "static const propvec &properties_%1$s = get_type_props<%2$s>();\n\n";
+
+
+  char *fullname = NULL, *newname = NULL;
+  if (IDL_PRINTA(&fullname, get_cpp11_fully_scoped_name, node, streams->generator) < 0 ||
+      IDL_PRINTA(&newname, get_cpp11_fully_scoped_name, node, streams->generator) < 0)
+    return IDL_RETCODE_NO_MEMORY;
+
+  char *ptr = newname;
+  while (*ptr) {
+    if (*ptr == ':')
+      *ptr = '_';
+    ptr++;
+  }
 
   if (multi_putf(streams, ALL, fmt)
-   || putf(&streams->props, pfmt))
+   || putf(&streams->props, pfmt)
+   || idl_fprintf(streams->generator->header.handle, pfmt2, newname, fullname) < 0)
     return IDL_RETCODE_NO_MEMORY;
 
   return IDL_RETCODE_OK;
@@ -1239,7 +1256,7 @@ process_struct(
 
   if (revisit) {
     if (print_switchbox_close(user_data)
-     || print_constructed_type_close(user_data)
+     || print_constructed_type_close(user_data, node)
      || (!is_nested(node) && print_entry_point_functions(streams, fullname)))
       return IDL_RETCODE_NO_MEMORY;
 
@@ -1309,7 +1326,7 @@ process_union(
 
   if (revisit) {
     if (multi_putf(streams, MAX, pfmt)
-     || print_constructed_type_close(user_data)
+     || print_constructed_type_close(user_data, node)
      || (!is_nested(node) && print_entry_point_functions(streams, fullname))) /*only add entry point functions for non-nested (topic) types*/
       return IDL_RETCODE_NO_MEMORY;
 
--- a/src/idlcxx/src/streamers.c
+++ b/src/idlcxx/src/streamers.c
@@ -819,11 +819,11 @@ add_member_finish(
     if (IDL_PRINTA(&accessor, get_instance_accessor, decl, &loc) < 0
      || multi_putf(streams, (WRITE|MOVE), "      }\n")
      || multi_putf(streams, ALL,
-          "      if (!streamer.finish_member(*prop, %1$s.has_value()))\n"
+          "      if (!streamer.finish_member(*prop, member_ids, %1$s.has_value()))\n"
           "        return false;\n", accessor))
       return IDL_RETCODE_NO_MEMORY;
   } else {
-    if (multi_putf(streams, ALL, "      if (!streamer.finish_member(*prop))\n"
+    if (multi_putf(streams, ALL, "      if (!streamer.finish_member(*prop, member_ids))\n"
                                  "        return false;\n"))
       return IDL_RETCODE_NO_MEMORY;
   }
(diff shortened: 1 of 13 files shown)


eboasson:
Friendly ping @sumanth-nirmal: my ability to review this is limited, so if you have any observations on it, that'd be great.

sumanth-nirmal (commented):
@reicheratwork Thanks for fixing this, and looks good except for the following

I think there is still one odd scenario where this can still lead to a "use after free", the scenario is as follows, If we have a static object that is created first which accesses a writer through a singleton (created later) in the destruction, then the first time a sample is written the `propvec` will be allocated and initialized. \
Once the cleanup starts, first the static `propvc` is destroyed (cleanup in reverse order) and then the static object in main is destroyed during which if it tries to access this writer then it access a freed `propvec` leading to "use-after-free". Sample code as below

```cpp
class C2
{
public:
  ~C2()
  {
    writer->close();
  }

  static C2 & get()
  {
    dds::domain::DomainParticipant participant(org::eclipse::cyclonedds::domain::default_id());
    dds::pub::Publisher publisher(participant);
    dds::topic::Topic<DDSType> topic(participant, "test_topic");
    static C2 c2(publisher, topic);
    return c2;
  }

  void publish()
  {
    DDSType msg(1, "Hello World");
    writer.write(msg);
  }

private:
  dds::pub::DataWriter<DDSType> writer{dds::core::null};

  C2(dds::pub::Publisher & pub, dds::topic::Topic<DDSType> & topic)
      : writer(dds::pub::DataWriter<DDSType>(pub, topic))
  {}
};

class C1
{
public:
  C1()
  {}

  ~C1()
  {
    C2::get().publish();
  }

  void foo()
  {
    C2::get().publish();
  }
};


int32_t main(const int32_t argc, char ** const argv)
{
  static C1 c1;
  c1.foo();
}
```
(orchestrating static order destruction is tricky!)

This can be listed as a limitation something along lines of "application entities accessing the cyclone entities shouldn't outlive the cyclone entities themselves".
Maybe we can also provide some hooks that can be called in the application cleanup phase that properly clean up the cyclone entities, however, this will need some careful thinking, and redesigning and is out of the scope of this issue

src/idlcxx/src/streamers.c
```diff
@@ -1057,31 +1052,23 @@ print_constructed_type_open(struct streams *streams, const idl_node_t *node)
...
-    "propvec &get_type_props<%s>()%s";
+    "const propvec &get_type_props<%s>()%s";
   static const char *pfmt2 =
     " {\n"
-    "  static thread_local std::mutex mtx;\n"
-    "  static thread_local propvec props;\n"
-    "  static thread_local entity_properties_t *props_end = nullptr;\n"
-    "  static thread_local std::atomic_bool initialized {false};\n"
+    "  static std::mutex mtx;\n"
+    "  static propvec props;\n"
```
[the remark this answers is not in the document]
reicheratwork: Maybe we can use an assign to a static propvec calling the creation function during the creation of the statics, even before the main function is called?

I will have look into this as a possible solution.

reicheratwork:
@sumanth-nirmal Added a small fix to initialize the entity properties before everything else.
This still does not fix all the issues uncovered by your example though, as there are some lifetimes that do not behave as nicely as they should.
But as these problems are not caused by the entity properties vector's lifetime (such as the sertype_ops and serdata_ops in datatopic.hpp), this issue can be considered as solved if you also see no further issues @sumanth-nirmal ?
I will look into opening issues for these other issues, though they are quite a niche issue (static readers/writers etc.).

src/idlcxx/src/streamers.c
```diff
@@ -1129,20 +1116,36 @@ print_switchbox_open(struct streams *streams)
...
+    "  return streamer.finish_struct(*props, member_ids);\n"
     "}\n\n";
   static const char *pfmt =
     "\n  entity_properties_t::finish(props, keylist);\n"
-    "  props_end = props.data() + props.size();\n"
     "  initialized.store(true, std::memory_order_release);\n"
     "  return props;\n"
     "}\n\n";
+  static const char *pfmt2 =
+    "static const propvec &properties_%1$s = get_type_props<%2$s>();\n\n";
```
sumanth-nirmal, line 1131: If this `static const` variable is not used anywhere, the compiler can optimize this and could lead to the same lifetime issues, correct?

sumanth-nirmal (commented):
Running the current version with my reproducer causes `'dds::core::AlreadyClosedError'` when trying to publish in the destructor. We probably need to solve these lifetime issues in a followup or agree on what can be listed as limitations

src/idlcxx/src/streamers.c
reicheratwork: > If this `static const` variable is not used anywhere, the compiler can optimize this and could lead to the same lifetime issues, correct?

Pulling the following simple example program through godbolt seems to call the "initialization function" `f()` during the global initialization step for the major compilers (GCC x86-64, CLang x86-64, MSVC x64) even at the highest optimization level (-O3):
```
const int fac(int i);

const int fac(int i)
{
if (i)
    return i*fac(i-1);
else
    return 1;
}

const int& f()
{
static int m_i = fac(4);
return m_i;
}

const int &ref_i = f();
 
int main()
{
    return f();
}
```

So I think your fears regarding the function call being optimized away are unfounded.
I could be wrong, as this is very much outside my zone of experience (and comfort) regarding C++ :)

reicheratwork:
> @reicheratwork Running the current version with my reproducer causes `'dds::core::AlreadyClosedError'` when trying to publish in the destructor. We probably need to solve these lifetime issues in a followup or agree on what can be listed as limitations

We may have to open another ticket regarding this issue, as there does seem to be something going on with global DDS entity administration, lifetimes, cleanups and the like, but that is not part of this PR, I think.

src/idlcxx/src/streamers.c
sumanth-nirmal: @reicheratwork Shouldn't be `static const int &ref_i = f()`, to be exact for our scenario?

I have slightly simplified your example and check this godbolt link https://godbolt.org/z/nbhc5h5xM if we don't have the attribute `__attribute__((used))` or `volatile`, the unused static const is actually optimized out when compiled, is my understanding incorrect here?

reicheratwork: @sumanth-nirmal  I think you are correct in this regard, I will add the additional change to the PR.

reicheratwork: Looks like this does not work for Windows, let me look further.

sumanth-nirmal: @reicheratwork Windows is definitely out of my league, but GNU `_attribute__` equivalent  in MSVC is `__declspec` but not sure if it supports an attribute similar to what we are need here

reicheratwork: > @reicheratwork Windows is definitely out of my league, but GNU `_attribute__` equivalent in MSVC is `__declspec` but not sure if it supports an attribute similar to what we need here

Exactly, it looks like MSVC does not have a "used" attribute or something equivalent.

sumanth-nirmal: @reicheratwork Looks good to me.

We just a need a follow-up issues for fixing or agreeing on the limitation with the lifetime of the entities

eboasson (approved):
LGTM, thanks @reicheratwork and @sumanth-nirmal!

PR #338, opened 2022-11-01
Fix for coverity issue 398731

reicheratwork:
Suppressing false positive reported by coverity


eboasson (commented):
This is an odd one ... Looking at the code, I also think it is a false positive. Usually I can find at least something that could plausibly lead to a f.p., but here it is as if the recent-ish Coverity update has an incorrect model of `weak_ptr` and that seems weird.

Anyone around who can double-check? @e-hndrks perhaps?

reicheratwork:
> This is an odd one ... Looking at the code, I also think it is a false positive. Usually I can find at least something that could plausibly lead to a f.p., but here it is as if the recent-ish Coverity update has an incorrect model of `weak_ptr` and that seems weird.
> 
> Anyone around who can double-check? @e-hndrks perhaps?

I think a recent Coverity update is the problem, as this previously caused no issues.
Also, it seems that Coverity does not dig deep enough to get the correct conclusion, as the weak pointer being locked is an entry in a static std::map object, so this cannot go out of scope.

eboasson (approved):

Issue #343, opened 2022-11-16
issue in idlcxx generated code for the sequence of non-primitive type in Discriminated Union


vivekpandey02:
Hi Erik & Martijn,
I got some issue into the idlcxx generated code for the sequence of non-primitive type in Discriminated Union.
Basically, I am using RnR.idl, union is below:

union Kind switch(CommandKind) {
        case ADD_RECORD_COMMAND:                /* Record command */
            AddRecordCommand addRecord;
        case REMOVE_RECORD_COMMAND:
            RemoveRecordCommand removeRecord;
        case ADD_REPLAY_COMMAND:                /* Replay command */
            AddReplayCommand addReplay;
        case REMOVE_REPLAY_COMMAND:
            RemoveReplayCommand removeReplay;
        case CONFIG_COMMAND:                    /* Config command */
            **sequence<KeyValue> config;**
        case START_SCENARIO_COMMAND:            /* Scenario-control commands */
        case STOP_SCENARIO_COMMAND:
        case SUSPEND_SCENARIO_COMMAND:
            string name;
        case SETREPLAYSPEED_COMMAND:            /* Storage replay-speed command */
            SetReplaySpeedCommand setreplayspeed;
        case TRUNCATE_COMMAND:                  /* Storage truncate command */
            string storage;
        case GENERIC_COMMAND:                   /* For future extensibility */
            **sequence<KeyValue> extCommands;**
    };

Here is the error with the sequence<KeyValue>:

/mnt/d/Workspace/RnR.hpp: In function ‘bool org::eclipse::cyclonedds::core::cdr::write(T&, const RnR::Kind&, org::eclipse::cyclonedds::core::cdr::entity_properties_t*)’:
/mnt/d/Workspace/RnR.hpp:9918:52: error: ‘prop’ was not declared in this scope; did you mean ‘props’?
 9918 |       if (!write(streamer, instance.config()[i_1], prop))
      |                                                    ^~~~
      |                                                    props
/mnt/d/Workspace/RnR.hpp:9956:57: error: ‘prop’ was not declared in this scope; did you mean ‘props’?
 9956 |       if (!write(streamer, instance.extCommands()[i_1], prop))
      |                                                         ^~~~
      |                                                         props

reicheratwork:
Looks like the streaming functions are thinking that they are in a struct with regards to the member they are streaming, so they think that they have the properties for member `prop` to give to the streamer.

[...]

PR #344, opened 2022-11-17
Fix issue 343

reicheratwork:
This PR will fix issue #343
- Reproducer of the issue: idl union with a case containing a sequence of structs 
- Fix of the issue: add correct recognition of this case and fix streamer code generation to generate the correct entity properties

@vivekpandey02 : as the reporter of this issue, could you please do a review of this PR?


reicheratwork:
Had to rebase onto the new master, due to changes on the interface with CycloneDDS-C

PR #349, opened 2022-11-28
Added missing export of QoSPolicy non-const accessor functions

reicheratwork:
These functions were previously inline, and therefore did not need to be exported, but with commit 7b1f434 these are no longer inline and need to be exported for Windows compiles
Added exports for these functions now


eboasson (approved):
Thanks!!

Issue #336, opened 2022-10-25
appendable & encoding_version == xcdr_v1


rockcdr:
https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/ce6d5e60f3277031d6c6da1f60a1d35b78858a22/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L227

A little confused on 'getExtensibility()' & 'allowableEncodings()' combinations in read_header().
Should the 'PL_CDR' be changed to 'PLAIN_CDR' for 'ext_appendable' & encoding_version==xcdr_v1, as writer_header()?

And I am not sure that 'ext_appendable' with 'PLAIN_CDR' should be 'basic_cdr' optionally, same as the case in 'ext_final'.

rockcdr:
@reicheratwork Thanks for your response.
For the second point, I am wondering whether it should be
```
    case extensibility::ext_appendable:
      switch (field) {
        case PLAIN_CDR:
          if (TopicTraits<T>::allowableEncodings() & DDS_DATA_REPRESENTATION_FLAG_XCDR1)
            ver = encoding_version::basic_cdr;
          else
            ver = encoding_version::xcdr_v1;
          break;
```
that is, 'ext_appendable' and 'ext_final' are similar.

reicheratwork:
>  ...
> that is, 'ext_appendable' and 'ext_final' are similar.

@rockcdr : ah, now I get what you were indicating.

The reason why there is a hard choice for `xcdr_v1` streaming is because basic cdr streaming (so without xtypes functionality) does not support appendable types, so the sequence of decisions is the following in this case:
- the datatype is appendable, this is known when we create the Topic
- basic cdr does not support appendable types
- therefore receiving the `PLAIN_CDR` encoding flag in this case means that we are receiving a legacy xtypes stream (`xcdr_v1`) as any other choice is internally inconsistent.

I hope that answers your question.

Also, there is a PR out for your issue: #350 , if you could review it, that would be very much appreciated.

rockcdr:
Cool! Thanks for clarification

reicheratwork:
@rockcdr : As the PR for this issue is merged, could you close this issue?

rockcdr:
Thanks

PR #350, opened 2022-11-29
Fix for issue #336

reicheratwork:
Incorrect encoding flag for xcdr_v1 appendable types was being required The read_header function was expecting the flag PL_CDR, but this should be PLAIN_CDR


eboasson (approved):
LGTM!

PR #354, opened 2022-12-14
Fix for coverity issue 73406

reicheratwork:
This is a simple copy/paste error, and I forgot to change the second comparison parameter away from the first.


eboasson (approved):
Thanks @reicheratwork !

Issue #367, opened 2023-02-14
CycloneDDS crashes when using typedef sequences


trittsv:
Hi, we are using CycloneDDS-CXX 0.10.2 with the following IDL:

At runtime `ReportRecordingStatus1` crashes and `ReportRecordingStatus2` works fine.
```cpp
module HelloWorldData
{
    typedef sequence<string> StringListType;

    struct ReportRecordingStatus1
    {
        StringListType InProgressRecordingsList;
    };

    struct ReportRecordingStatus2
    {
        sequence<string> InProgressRecordingsList;
    };
};
```

```
=== [Publisher] Create writer.
1676386490.708493 [0] HWD_Status: determined enp0s3 (udp/10.0.2.31) as highest quality interface, selected for automatic interface.
on_publication_matched, current_count: 1
=== [Publisher] Write sample.
=== [Publisher] Write sample with contents.
terminate called after throwing an instance of 'dds::core::InvalidArgumentError'
  what():  Error Bad Parameter - write failed.
===============================================================================
Context     : void org::eclipse::cyclonedds::pub::AnyDataWriterDelegate::write
Node        : UnknownNode

fish: './HWD_Status1_publisher' terminated by signal SIGABRT (Abort)
```
Here is the logfile from cyclone and the sources:
[cyclone_error.zip](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/10733763/cyclone_error.zip)

@eboasson is this a known bug?

eboasson:
@reicheratwork for me it still fails on current master, and it looks like the generated code for the two types is nearly the same, but if one works and the other fails then surely there must be something significant in the difference. Would you be willing to take a look?

trittsv:
We found out that the crash is new in cyclone version 0.10.2 with version 0.9.1 it works fine.

reicheratwork:
Have made a PR for the fix [here](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/369)

trittsv:
Hey @reicheratwork @eboasson we just tested the changes (#369 ) in our software. It works fine 😃 From our perspective it can be merged. Thank you very much 🥳

PR #369, opened 2023-02-21
Fix for issue #367

reicheratwork:
Fix for issue #367 @trittsv, could you give this a look and check whether it fixes the issue on your side?


trittsv:
We just tested the changes in our software, it works fine 😃 Thank you very much 🥇

eboasson (approved):
Thanks @reicheratwork and @trittsv 🙂

PR #372, opened 2023-02-28
Add fix for Conan update

eboasson:
The default Conan version was updated to 2.0, so now we force fix the version to 1.59.0 as the current CUnit version (2.3.1) is currently not compatible with Conan v2.0


NoxPardalis (approved):
Looks good to me.

PR #373, opened 2023-03-03
Fix for misaligned access for 8 byte types

reicheratwork:
The maximum alignment for XCDR2 type streams is 4 bytes, which can cause misaligned accesses when attempting to simply assign primitive types of 8 bytes size, this commit fixes that by replacing simple assigns with memcpy


eboasson (commented):
Thanks @reicheratwork. I think this applies to all types now and not just the 64-bit integers and floats, right? That's definitely correct, but I am worried a bit about there now being library calls to `memcpy` for every field.

It it now ends up calling `memcpy` it may be wiser to specialise it for the 64-bit types (because the ≤32-bit types don't suffer from misaligned access) and to load 2 32-bit integers instead and combine them into a single thing. With `gcc`'s semantics that would be pretty straightforward, if I'm not mistaken something like:

```C++
const uint32_t *from_u32 = reinterpret_cast<const uint32_t *>(str.get_cursor());
union { uint32_t u32[2]; T val; } tmp;
tmp.u32[0] = from_u32[0];
tmp.u32[1] = from_u32[1];
toread = tmp.val;
```

would be correct (for 64-bit int & double).

But if all the compilers we deal with turn the call to `memcpy` into something like the above, then I would rather leave it like this.

Did you get a chance to check with, e.g., the compiler explorer?

eboasson (approved):
Thanks @reicheratwork for addressing my worries. I see one problem with the PR as it stands: on a big-endian machine the *test* will fail. Without a big-endian machine in the CI, it gets difficult to test that case anyway.

I believe it is best to merge it as-is and move forward rather than scramble to fix the test and then have a problem verifying the fix 😂

PR #380, opened 2023-03-10
Fix build problem when type/topic discovery disabled

### LINKED ISSUE #370: compile error
When compiling cyclonedds-cxx on ubuntu as per cyclonedds documentation I get the following error:

/home/<user>/Projects/Tools/cyclonedds-cxx/src/ddscxx/include/dds/topic/detail/TTopicImpl.hpp:312:77: error: ‘getTypeInfo’ is not a member of ‘org::eclipse::cyclonedds::topic::TopicTraits<dds::topic::TParticipantBuiltinTopicData<org::eclipse::cyclonedds::topic::ParticipantBuiltinTopicDataDelegate> >’

This error appears 4 times.
Using the latest release of cyclonedds-cxx

### LINKED ISSUE #374: DDS  type object && communication frag
**Problem background:**
In Extensible and Dynamic Topic Types for DDS Version 1.3， Minimal TypeObject has a smaller amount of data in the discovery process. In cyclone dds source code , these parameters can be set,  relevant enumeration values in the below:
typedef enum ddsi_ typeid_ kind {
DDSI_ TYPEID_ KIND_ MINIMAL,
DDSI_ TYPEID_ KIND_ COMPLETE,
DDSI_ TYPEID_ KIND_ PLAIN_ COLLECTION_ MINIMAL,
DDSI_ TYPEID_ KIND_ PLAIN_ COLLECTION_ COMPLETE,
DDSI_ TYPEID_ KIND_ FULLY_ DESCRIPTIVE
} ddsi_ typeid_ kind_ t;

**Specific problem description:**
Now I have a few questions to ask:
1. If DDSI_ TYPEID_ KIND_ If MINIMAL is used, will it reduce the amount of data exchanged by both sides in the process of discovery and data manipulation at both ends (the default setting in the cyclone dds code is DDSI_TYPEID_KIND_COMPLETE).
2. If I want to change the program to Minimal TypeObject mode (so that the program communication will be lighter), how do I do it? (If IDL is complex, but I don't want to use DDS frag, what should you do? I try to change the DDSI_TYPEID_KIND_COMPLETE in the code to DDSI_TYPEID_KIND_MINIMAL, but it doesn't work)

**Thanks**

eboasson:
This makes it possible again to build without type/topic discovery support, and also adds a CI configuration to reduce the likelihood of this breaking again.

I've also bumped the boost version used on the CI, simply because I noticed that it was building the old version of boost from scratch and figured that the new version would be more likely to have binaries available for the updated set of platforms. (So far, so good.)

@reicheratwork my fix is not the most elegant because I completely disabled the "find topic" operation, but I think there are some cases that can be supported even when these parts of discovery are disabled. In the grand scheme of things, I think it is better to disable it altogether in this case than for it not to work at all, so I am in favour of merging this quickly and possibly improving the fix later. But I'd be interested in your opinion on this.

Fixes #370 #374

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -115,6 +115,8 @@ steps:
             -DCMAKE_PREFIX_PATH="${BUILD_SOURCESDIRECTORY}/iceoryx/build/install" \
             -DSANITIZER=${SANITIZER:-none} \
             -DENABLE_SHM=${ICEORYX:-off} \
+            -DENABLE_TYPE_DISCOVERY=${TYPE_DISCOVERY:-on} \
+            -DENABLE_TOPIC_DISCOVERY=${TOPIC_DISCOVERY:-on} \
             ${GENERATOR:+-G} "${GENERATOR}" -A "${PLATFORM}" -T "${TOOLSET}" ..
       cmake --build . --config ${BUILD_TYPE} --target install -- ${BUILD_TOOL_OPTIONS}
     name: install_cyclonedds
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -77,6 +77,13 @@ strategy:
       sanitizer: address
       cc: clang-12
       cxx: clang++-12
+    'Ubuntu 22.04 LTS with GCC 12 (Debug, x86_64, no type discovery)':
+      image: ubuntu-22.04
+      sanitizer: address
+      cc: gcc-12
+      cxx: g++-12
+      type_discovery: off
+      topic_discovery: off
     'macOS 11 with Clang 12 (Debug, x86_64)':
       image: macOS-11
       sanitizer: address
--- a/conanfile.txt
+++ b/conanfile.txt
@@ -1,6 +1,6 @@
 [requires]
 gtest/1.12.1
-boost/1.78.0
+boost/1.81.0
 
 [generators]
 cmake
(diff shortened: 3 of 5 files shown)


eboasson:
Good point @reicheratwork, thanks!

eboasson:
@phbasler I only included fixes in 0.10.4 that I knew to be causing problems in practice — less work, but also less risk of accidents. That's all.

How big of a problem is the absence of this fix in 0.10.4? I am not against a 0.10.5, but I would prefer to tag master instead 🙂

phbasler:
@eboasson It's a minor inconvience. We have to split up a testing endpoint from the main code to be tested because we want to disable this option in the main code.

> but I would prefer to tag master instead
So you are not using a fixed commit of the master branch as the release tag? How do you decide when to base a release on the current master branch or some other branches?

eboasson:
> So you are not using a fixed commit of the master branch as the release tag? How do you decide when to base a release on the current master branch or some other branches?

There is a branch for every release (releases/0.9.x, releases/0.10.x, etc), the only updates that get pushed on such a branch are fixes warranting a patch release. The release tags point to commits on these branches.

What I meant by "tagging master" is really making a new release branch based on current master, getting the release-specific details (like version numbers) sorted and tagging it.

Master is the current development version and so in theory there is a QA cycle in the process as well, but in practice I don't really need it: it is almost always good enough for a release because I have seen what it takes to regain control over quality once it has been lost and I don't want to have to go through that with Cyclone.

There were a few things I wanted in, but they are in now. Before there can be an official release (a non-pre-release) there has to be a review from the Eclipse Foundation and for that I have to do paperwork, but for a pre-release there is no such need.

PR #381, opened 2023-03-10
Lifetime fix

reicheratwork:
Add some fixes for lifetimes of entities not matching up with callback functions onto the same entities.
This causes issues like null dereferences and deadlocks.

This fix includes:
- reproducer for the issue
- ensurances that the lifetimes of entities are at least as long as the their callback functions
- listeners are present when callbacks are executed

This will not fix the entire issue, which lies at some choices made in the CycloneDDS-C layer.
If similar issues are still encountered, we advise to set the listener on an entity to `nullptr` before destroying it.


eboasson (approved):
Thanks @reicheratwork !

Issue #383, opened 2023-03-13
idlcxx does not handle "-o" output directory flag


Pro:
If I call idlc with cxx and `-o` flag, the generated files are still generated in the current directory, and not in the directory given as output dir.

```
 cd $HOME/cwd/hello_world && idlc -llibcycloneddsidlcxx.so -x appendable -o$HOME/outdir/idl HelloWorldData.idl
```

Puts the files in `$HOME/cwd/hello_world` and not `$HOME/outdir/idl`

reicheratwork:
@Pro , please take a look at PR #418 , this should implement the functionality you desire (in addition to BASE_DIR).

PR #384, opened 2023-03-13
feat: add OUTPUT_DIR to IDLCXX_GENERATE

Pro:
Depends on https://github.com/eclipse-cyclonedds/cyclonedds/pull/1603

Allows to set OUTPUT_DIR for IDLCXX_GENERATE

Does not fully work yet, due to #383

### DIFF
--- a/src/idlcxx/Generate.cmake
+++ b/src/idlcxx/Generate.cmake
@@ -11,7 +11,7 @@
 #
 
 function(IDLCXX_GENERATE)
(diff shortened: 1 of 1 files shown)


eboasson (approved):
LGTM, thanks @Pro. As you note, this of course is of no use without addressing #383 but I don't see why I shouldn't merge this PR already.

Regarding the CI failures, it looks to me like those are because of the addition of `OUTPUT_DIR`, so I restarted them now that I have merged the PR that adds the `OUTPUT_DIR` option in the core repository.

PR #385, opened 2023-03-13
Update SharedMemory test to include partition name

eboasson:
This updates the tests that interact directly with Iceoryx using the service names used for DDS topics to include the partition name, in accordance with the fix in the core repository for preventing cross-talk between DDS partitions when Iceoryx is used.


eboasson:
Just a fixup of a test, so merging.

PR #387, opened 2023-03-15
Bump version number to 0.10.3

eboasson:
... plus backport of
* fix for crashes in some cases where destruction is delayed because of refcounting
* fix for listeners not being installed correctly if set in constructor

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -76,7 +76,7 @@ steps:
     name: cache_conan
   - bash: |
       set -e -x
-      pip install conan --user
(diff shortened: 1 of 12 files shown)


eboasson:
> @eboasson would it be possible to include this bug fix [686c06d](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/686c06da3b8ce3072caf22631ef9989b7b159fc6) too in the release?

It is a tiny change with a clean cherry-pick, so yes, good catch! I've pushed it onto this PR.

eboasson:
On consideration, #373 should be in as well, I think.

Issue #389, opened 2023-03-17
DurabilityService is not supported by DataWriterQos


tez1szh:
DurabilityService is applicable for both TopicQos and DataWriterQos according to DDS v1.4
However, in CycloneDDS C++ binding, DurabilityService can only be set to TopicQos, not DataWriterQos.

As a comparision, DurabilityService is support by DataWriterQos in CycloneDDS - C binding. See https://github.com/eclipse-cyclonedds/cyclonedds/issues/1616

In case I want to have 2 respective datawriters for 2 instances with different qos settings, this might cause some trouble.
Just wanna check if DurabilityService should be supported by DataWriterQos or there's any special consideration not doing so.
Draft patch is available, and I can create a PR to add DurabilityService in DataWriterQos after some tests if necessary.

PR #390, opened 2023-03-20
Fix for issue #389

tez1szh:
This commit enables setting DurabilityService policy of DataWriterQos as described in OMG DDS v1.4 spec

image


eboasson (approved):
Thank you so much @tez1szh !

PR #400, opened 2023-04-26
Fix setting of DurabilityService policy in DataWriter QoS

reicheratwork:
The DurabilityService QoSPolicy was not being copied correctly when initializing a DataWriter QoS from a Topic QoS
This issue was missed as the DurabilityService policy of the DataWriter QoS was not being checked for its values in the unittests, this was also fixed


eboasson (approved):
Thanks @reicheratwork !

PR #401, opened 2023-04-26
Add profiles to conan install

reicheratwork:
Conan requires profiles in the invoke of the install command starting with version 2.0, with it being an optional parameter at this time
Added the default profile to conan install to head this future issue off

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -109,7 +109,7 @@ steps:
                 cyclonedds
       mkdir cyclonedds/build
       cd cyclonedds/build
-      conan install -b missing -s arch=${ARCH} -s build_type=${BUILD_TYPE} ..
+      conan install -b missing -s arch=${ARCH} -s build_type=${BUILD_TYPE} .. -pr:b=default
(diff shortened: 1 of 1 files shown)


eboasson (commented):
Once we add the build profile, can we switch to Conan 2.0?

I also think we should do a PR for the core repository, having different ways of using Conan in the C and C++ repositories is asking for trouble. (Maybe it was a bad idea to have separate repositories, or maybe it would be a good idea to have the common build bits in yet another one that we import, or ... But right now the repositories are what they are.)

reicheratwork:
> Once we add the build profile, can we switch to Conan 2.0?
> 
> I also think we should do a PR for the core repository, having different ways of using Conan in the C and C++ repositories is asking for trouble. (Maybe it was a bad idea to have separate repositories, or maybe it would be a good idea to have the common build bits in yet another one that we import, or ... But right now the repositories are what they are.)

As far as I know, this should make us ready for `conan 2.0`, yes.

I will also have a look at the `CycloneDDS-C` repo and see what's going on with `conan` there.

P.S.: this was fixed by using information from [this](https://github.com/conan-io/conan/issues/12271) issue on the `conan` github.

PR #402, opened 2023-05-01
Add GitHub workflow for checking MISRA rule regressions

NoxPardalis:
This PR adds a GitHub workflow for checking MISRA rule violations. It currently includes a check for MISRA rule 3.1 but more will be added as the code-base becomes more compliant.

See eclipse-cyclonedds/cyclonedds#1673 for the same PR included in the core Cyclone repository. The only major difference in is that the build for the code-base is manually specified in this workflow instead of being identified by auto-build ([misra-regressions-check.yml:64-71](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/compare/master...NoxPardalis:cyclonedds-cxx:master?expand=1#diff-ff15e3d24fedce3faac6a63b42d98272ad5bb80ec6fbf00fddc6cd5ff1567e68R67-R71))

See #403 for the PR that solves the rules checked for by the workflow introduced within this PR.

### DIFF
--- a/.github/codeql/resolved-misra-rules.yml
+++ b/.github/codeql/resolved-misra-rules.yml
@@ -1,0 +1,5 @@
+disable-default-queries: true
+
+# List of MISRA rules that have been resolved to check for regressions against.
+queries:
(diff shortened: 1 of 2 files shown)


eboasson (approved):
LGTM, Thanks @NoxPardalis !

PR #403, opened 2023-05-01
Address MISRA rule 3.1 violations

NoxPardalis:
This resolves the issue of // within /*...*/ comments. These typically feature within the URLs present in the copyright blurbs and are allowed within comments starting with a // but not the /*...*/ style comments. This is applied across all .c, .h, .cpp and .hpp files. See #402 for the PR that includes the GitHub code scanning for the same rule.


eboasson (approved):
LGTM, Thanks @NoxPardalis !

PR #404, opened 2023-05-03
Throughput example fixes

reicheratwork:
Made the C++ throughput example program to be more in line with the expected output from the same example in CycloneDDS-C.

### DIFF
--- a/examples/throughput/publisher.cpp
+++ b/examples/throughput/publisher.cpp
@@ -146,8 +146,10 @@ void start_writing(
         sample.count()++;
       }
 
-      writer->write_flush();
-      std::this_thread::sleep_until(burstStart + burstInterval);
+      if (burstInterval > std::chrono::seconds(0)) {
+        writer->write_flush();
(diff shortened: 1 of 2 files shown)


eboasson (commented):
The transfer rate seems off ...
```
=== [Subscriber] 1050151917.000 Payload size: 8192 | Total received: 533591 samples, 4371177472 bytes | Out of order: 0 samples Transfer rate: 0.00 samples/s, 0.00 Mbit/s
```
It has to do with computing `deltaTime` (it is off by a factor of 10⁹) but I don't know the elegant solution in C++.

Otherwise I am fine with it. I can't in all honesty say "looks good to me" but I just find these "throughput" and "roundtrip" examples horrible on two counts and neither is your doing:
* They are in a sense abusing DDS and not showing anything interesting, and one runs the risk of people using it to actually do benchmarking ...
* The code has become so convoluted over the years with different options that is also not a straightforward example anymore

(Rant over 🙂)

I don't see lost samples and the rate is inline with expectations, so if you could have another look at the rate calculation, that'd be great.

PR #406, opened 2023-05-04
Writer batching qos

reicheratwork:
Implementing a C++ equivalent of the CycloneDDS-C writer batching QoS, allowing users to make CycloneDDS to queue writes until flushed.


src/ddscxx/include/dds/core/policy/CorePolicy.hpp
```diff
@@ -152,6 +152,9 @@ UserData;
 typedef dds::core::policy::detail::WriterDataLifecycle
 WriterDataLifecycle;
 
+typedef dds::core::policy::detail::WriterBatching
+WriterBatching;
```
eboasson, line 156: After all the effort spent by the OMG in defining the API to allow vendor-specific extensions accessible only via the `->` operator and presumably with the intent to avoid the need to modify the OMG-provided header files, I wonder if it is necessary (and if not, if it is best) to add the new QoS here rather than in an implementation-specific file.

It looks to me like OpenSplice managed to add a "ReaderLifespan" QoS setting [in the delegate](https://github.com/ADLINK-IST/opensplice/blob/a4818af5f5931834501055cd7ce2339de4f0af13/src/api/dcps/isocpp2/include/org/opensplice/core/policy/PolicyDelegate.hpp#L663) without modifying the OMG header files for it, other than a comment and (shockingly) a need to maintain a [copy](https://github.com/ADLINK-IST/opensplice/blob/a4818af5f5931834501055cd7ce2339de4f0af13/src/api/dcps/isocpp2/include/dds/core/detail/ProprietaryApi.hpp#L969) of proprietary stuff just for [generating documentation](https://github.com/ADLINK-IST/opensplice/blob/a4818af5f5931834501055cd7ce2339de4f0af13/src/api/dcps/isocpp2/include/dds/core/detail/ProprietaryApi.hpp#L53) ...

To me, that solution is worse than simply adding it to the OMG header files. But perhaps you could do me a favour and have a look whether my conclusion is correct and whether you agree with my sentiment that your choice is likely the lesser of the evils?

(I have to admit to being torn between trying to follow what the OMG did, and throwing out most of it and just having a simpler structure with fewer levels of indirection in the templates.)

eboasson:
One more thing @reicheratwork: it looks fine to me in all other respects, thank you!

src/ddscxx/tests/Qos.cpp
```diff
@@ -363,6 +366,9 @@ TEST(Qos, DataWriter)
     DataWriterQos dwQosTAssigned2;
     dwQosWAssigned2 = dwQosShifted;
     dwQosTAssigned2 = tQosShifted;
+    dwQosTConstructed << nonDefaultWriterBatching;  /* !!!inelegant!!! */
```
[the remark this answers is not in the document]
reicheratwork: This is exactly the case.
As the TopicQoS does not support the batching policy, it needs to be added to the DataWriterQoS at some point, as it does not get it from the TopicQoS.
And keeping all the tests the same allows the test to not need a whole jungle of cases/compares/ifs/ands/buts.

The inelegant tag was not really meant for the final commit, it was more as a means of exactly triggering this type of discussion :)
Do we also add the WriterBatching QoSPolicy to the TopicQoS? Probably not, as it is not part of the TopicQoS set of policies.
So maybe just making that note there as you suggested is the best approach for the moment.

Though I do think the CycloneDDS-C approach to QoS sets is the more elegant solution compared to the one chosen for CycloneDDS-CXX.

src/ddscxx/include/dds/core/policy/CorePolicy.hpp
eboasson: It appears that the two of us are in agreement. Surely that means we are right 🤣 Merging it before anyone complains!

PR #408, opened 2023-05-11
fix: avoid memleak when setting user data policy

Pro:
`org::eclipse::cyclonedds::core::convertByteSeq` allocates a new memory block, which is then copied via `dds_qset_userdata` but not freed afterwards.


src/ddscxx/src/org/eclipse/cyclonedds/core/policy/PolicyDelegate.cpp
```diff
@@ -1573,6 +1573,7 @@ void UserDataDelegate::set_c_policy(dds_qos_t* qos) const
         void* data = NULL;
         org::eclipse::cyclonedds::core::convertByteSeq(value_, data, static_cast<int32_t>(value_.size()));
         dds_qset_userdata(qos, data, value_.size());
+        dds_free(value);
```
Pro, line 1576:
```suggestion
        dds_free(data);
```

Issue #407, opened 2023-05-09
unmatched serialization and deserialization function invocations when using SHM


SeanYu81:
Hello Guys,
    
    Thanks for your support in advance!
    Recently, I am testing cyclonedds SHM communication, so I refer to official document to build/setup SHM environment (https://cyclonedds.io/docs/cyclonedds/latest/shared_memory/shared_memory.html).
    When I run my test case, SHM does not work as I expected. The test case always generates core dump due to invalid memory access. Related error message comes from cyclonedds-cxx function.
    -------------------------------------------
    include/ddscxx/org/eclipse/cyclonedds/topic/datatopic.hpp:69: bool to_key(const T&, ddsi_keyhash_t&) [with T = AgvStartStop; ddsi_keyhash_t = ddsi_keyhash]: Assertion `false' failed.
    Aborted (core dumped)
    -------------------------------------------
    
    In order to debug this problem and try to find the root cause, I use gdb debugger to track test case execution procedure step by step. Finally, I find the problem maybe come from unmatched serialization and deserialization function invocations. The following lists a brief callflow.
    -------------------------------------------
    dds_write_impl (dds_writer *wr, const void * data, dds_time_t tstamp, dds_write_action action) in dds_write.c
    --> dds_write_impl_iox (wr, ddsi_wr, writekey, data, tstamp, action)
          --> get_iox_chunk (wr, data, &iox_chunk)
                --> create_and_fill_chunk (wr, data, iox_chunk)
                      --> shm_create_chunk (wr->m_iox_pub, required_size)
                      --> fill_iox_chunk (wr, data, *iox_chunk, required_size)
                            --> ddsi_sertype_serialize_into(wr->m_wr->type, sample, iox_chunk, size)
                                   --> tp->ops->serialize_into(tp, sample, dst_buffer, dst_size)
                                          (**Note: in serialize_into function, buffer is filled with 3 calls(write_header, write, finish_header). So the buffer is not equal to sample!!!** )
          --> ddsi_serdata_from_loaned_sample (ddsi_wr->type, writekey ? SDK_KEY : SDK_DATA, iox_chunk)
                --> type->serdata_ops->from_iox_buffer (type, kind, NULL, (void *) sample)
                      --> serdata_from_iox_buffer(const struct ddsi_sertype * typecmn, enum ddsi_serdata_kind kind, void * sub, void * iox_buffer) in datatopic.hpp
[...]

PR #411, opened 2023-05-16
fix shm deserialize data error for those data with no fixed size

SeanYu81:
link to the issue: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/issues/407

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
@@ -647,7 +647,7 @@ ddsi_serdata * serdata_from_iox_buffer(
     }
 
     // key handling
-    if (typecmn->fixed_size == 1) {
+    if (typecmn->fixed_size) {
       const auto& msg = *static_cast<const T*>(d->iox_chunk);
       d->key_md5_hashed() = to_key(msg, d->key());
       d->populate_hash();
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
@@ -647,16 +647,22 @@ ddsi_serdata * serdata_from_iox_buffer(
     }
 
     // key handling
-    T msg;
-    iceoryx_header_t *iox_header = iceoryx_header_from_chunk(d->iox_chunk);
-    if (!deserialize_sample_from_buffer(d->iox_chunk, iox_header->data_size, msg, kind)) {
-      delete d;
-      d = nullptr;
-    } else {
+    if (typecmn->fixed_size == 1) {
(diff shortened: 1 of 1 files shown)


eboasson:
Thank you very much @SeanYu81!

After looking at the proposed change and at the CI failures, it looks like you fixed the case where the data gets transferred through shared memory in serialized form (for example because it contains a `std::string`, and we can't copy that from one process into the next) but broke the case where the data does not need to be serialized (for example a type with only some integers).

I think both the old and the new code need to be present, choosing the one based on what is actually in memory. If you're happy to take a look at that, then that would be great 😊 

The other obvious thing we are missing is a test case for the case that failed ... but that's not a blocker for merging a fix, just something that really must be done soon.

SeanYu81:
Hello @eboasson,

  Thanks for your comments.
  You are right, and I make a mistake to ignore logic for fixed size data. I will make changes again and please help to review.

Regards,
Yu, Zhi Qiang

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -647,9 +647,21 @@ ddsi_serdata * serdata_from_iox_buffer(
     }
 
     // key handling
-    const auto& msg = *static_cast<const T*>(d->iox_chunk);
-    d->key_md5_hashed() = to_key(msg, d->key());
-    d->populate_hash();
+    if (typecmn->fixed_size == 1) {
```
eboasson, line 650: Nitpick:
```suggestion
    if (typecmn->fixed_size) {
```
(or, alternatively, `typecmn->fixed_size != 0`)

What you wrote is perfectly correct given that `fixed_size` is a single bit, but it seems a possible risk if anything gets refactored.

eboasson (commented):
Thanks @SeanYu81 this looks good to me. I'd have merged it if I didn't have a nitpick-y comment. Would you please have a look at that?

eboasson:
There's one more thing, @SeanYu81: the Eclipse Foundation requires that the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) is signed (the "eclipsefdn/eca" check). Would you be willing to sign that?

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
SeanYu81: hi @eboasson Your change is great and it will make codes more robust. Thanks~

SeanYu81:
> There's one more thing, @SeanYu81: the Eclipse Foundation requires that the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) is signed (the "eclipsefdn/eca" check). Would you be willing to sign that?

hi @eboasson No problem, and I have signed the ECA.

PR #412, opened 2023-05-17
Removed superfluous template parameters

reicheratwork:
This due to an error in the spec which is not due to be fixed until C++23
As CLang compiles with pedantic by default, the template signature warnings are now also included by default, these warnings are now suppressed, but only for CLang version>=12

This should also solve PR #187 without having code inside #ifdefs


eboasson (approved):
Thanks @reicheratwork !

Issue #413, opened 2023-05-19
not using the output directory when using idlc to generat cxx header and source files


vecenyang:
I tried to use idlc to generate header file and source files to assigned output dir. And I found that **-o** argument can only work for c files generating and it doesn't work for cxx files generating. e.g.:
```
/// demo.idl
module demo
{
    enum e1
    {
        item1,
        item2
    };
};
```
Please try the following command:
```
idlc -o ./src demo.idl
idlc -l cxx -o ./src demo.idl
```
The generated cxx files could not be generated to assigned output directory **./src**.
It seems that in the generate() (line 971, cyclonedds-cxx/src/idlcxx/src/generator.c) function does not consider when output dir config->output_dir is not null, it only generate the files under the same path where the IDL file is.
I tried to fix this issue with the following code:
```
  if (idl_isabsolute(path) || !sep)
  {
    if (config->output_dir)
    {
      if (idl_isabsolute(config->output_dir))
        dir = idl_strndup(config->output_dir, strlen(config->output_dir));
      else
      {
        char* basedir = idl_strndup(path, (size_t)(sep-path));
        idl_asprintf(&dir, "%s/%s", basedir, config->output_dir);
        free(basedir);
      }
      if(idl_mkpath(dir) < 0)
        goto err_basename;
    }
    else
      dir = empty;
  }
```
Please someone have a look.

PR #418, opened 2023-06-01
Added output directory to idlcxx

reicheratwork:
The output directory parameter was not being used in when passed from idlc compiler to idlcxx library
Fixed so the output directory is now used in idlcxx
This should fix issue #413

@vecenyang : thanks for the suggested solution, can I rely on you to do a review of this PR?

### DIFF
--- a/src/idlcxx/Generate.cmake
+++ b/src/idlcxx/Generate.cmake
@@ -11,7 +11,7 @@
 #
 
 function(IDLCXX_GENERATE)
-  set(one_value_keywords TARGET DEFAULT_EXTENSIBILITY OUTPUT_DIR)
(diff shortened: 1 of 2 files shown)


vecenyang (approved):
It works on my side. I approved.

reicheratwork:
Added some additional improvements and simplifications, so now it also supports the BASE_DIR functionality.

eboasson (approved):
Thanks @reicheratwork, it looks good to me and I think it'll make various people happy!

eboasson (approved):
Thanks @reicheratwork !

herrfrei:
For me version 0.10.4 does not work. The PR #418 seems not be merged into this release.

Issue #410, opened 2023-05-15
on_subscription_matched / on_publication_matched listeners not reliable / only called once


Pro:
I am still facing the issue described in #378 with version 0.10.3
(ping @trittsv)

I.e., whichever process starts second, it does not properly call the corresponding `on_subscription_matched` or `on_publication_matched` callbacks in the Listeners.

It is very random, i.e., 1 out of 10 times it is called. Note that I am explicitly creating the entities in disabled state. So independent of the specification interpretation (as mentioned in #378), the listener callbacks must always be called for all received messages.

But this is not the case at all. Note that the same code works perfectly fine with RTI Connext DDS.

I used the same code as in #378 for reproduction steps: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/10918029/listener-not-reliable.zip

With a small modification, to re-create the publisher and writer in a loop (see code below) and keep the subscriber running forever.

The interesting part now is, that the on_publication_matched callback is only called once (in the best case).
For subsequent creations of the publisher & writer, the on_publication_matched is not called at all.

I.e., if you start the subscriber.cpp, and then the publisher.cpp, you will see the following output:

```
=== [Publisher] Create writer.
on_publication_matched, current_count: 1
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
=== [Publisher] Waiting for sample to be accepted.
=== [Publisher] Create writer.
=== [Publisher] Write sample.
[...]

PR #420, opened 2023-06-05
Fixed missing callbacks for pending events

e-hndrks:
When creating an Entity with a Listener, any instantaneous events that occurred during creation would miss their callback. 
This was caused by the C++ API not being able to wrap the underlying C API with a C++ object prior to C already trying to invoking the callback. 
That is now resolved by not setting the Listener at creation time, but by setting it after successfully putting a C++ wrapper around the underlying C Entity.
This should fix issue #410
A prerequisite for applying this fix is to make sure pull request #https://github.com/eclipse-cyclonedds/cyclonedds/pull/1717 is applied to the cyclonedds repository.

### DIFF
--- a/src/ddscxx/include/dds/pub/detail/DataWriter.hpp
+++ b/src/ddscxx/include/dds/pub/detail/DataWriter.hpp
@@ -60,9 +60,7 @@ public:
 
     DataWriter(const dds::pub::Publisher& pub,
                const ::dds::topic::Topic<T>& topic,
-               const dds::pub::qos::DataWriterQos& qos,
-               dds::pub::DataWriterListener<T>* listener,
-               const dds::core::status::StatusMask& mask);
+               const dds::pub::qos::DataWriterQos& qos);
 
     virtual ~DataWriter();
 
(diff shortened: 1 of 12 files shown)


eboasson (approved):
Looks good to me, thanks @e-hndrks !

(I'll wait with merging it until the C part has been merged.)

eboasson:
I started a CI run for this PR after merging https://github.com/eclipse-cyclonedds/cyclonedds/pull/1717 and it looks like you missed something in Iceoryx-specific code: https://dev.azure.com/eclipse-cyclonedds/cyclonedds-cxx/_build/results?buildId=4762&view=logs&j=e4f8c82b-de6b-5575-3950-14744af47318&t=aacbc600-1f24-51b1-8fd9-0679ad3e2601

Perhaps you could have another look?

e-hndrks:
Indeed, I didn't realize that the C++ repository had specialized Shared Memory tests, and therefore didn't run an IceOryx build. However, those remaining issues should now also have been fixed.

eboasson (approved):
Thanks @e-hndrks !

PR #434, opened 2023-07-24
Clean up the C++ code to read/take samples from a DataReader

e-hndrks:
Clean up the C++ code to read/take samples from a DataReader by delegating them to the new dds_read/take_with_collector calls added to the C-API.

The old code base was using some dirty tricks to peek into the contents of the reader cache in order to allocate matching C++ resources, prior to the invocation of the regular dds_read/take calls of the underlying C-API. In order for the contents of the reader cache not to change in between the peek and the actual consumption of the data, the readercache lock would not be released by the peek operation and not be claimed by the dds_read/take operation (upon passing it a special value for the max_samples parameter). However, such a dirty trick goes against many coding conventions, and for that purpose a more flexible set of data accessor operations have been added as part of https://github.com/eclipse-cyclonedds/cyclonedds/pull/1744.

### DIFF
--- a/src/ddscxx/CMakeLists.txt
+++ b/src/ddscxx/CMakeLists.txt
@@ -59,6 +59,7 @@ set(sources
     src/org/eclipse/cyclonedds/sub/SubscriberDelegate.cpp
     src/org/eclipse/cyclonedds/sub/BuiltinSubscriberDelegate.cpp
     src/org/eclipse/cyclonedds/sub/QueryDelegate.cpp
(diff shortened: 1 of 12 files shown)


eboasson (approved):
Thanks @e-hndrks ! LGTM — with the usual caveat that I don't know the C++ binding all that well ...

CI is all green if built using https://github.com/eclipse-cyclonedds/cyclonedds/pull/1744 instead of Cyclone DDS's master branch: https://dev.azure.com/erikboasson/Cyclone%20DDS%20CXX/_build/results?buildId=885&view=results

PR #436, opened 2023-07-27
Keyhash field sequencing

reicheratwork:
This fixes the issue with the members being in definition order when serializing for keyhashes

### DIFF
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
@@ -48,10 +48,11 @@ using org::eclipse::cyclonedds::core::cdr::xcdr_v1_stream;
 using org::eclipse::cyclonedds::core::cdr::xcdr_v2_stream;
 using org::eclipse::cyclonedds::core::cdr::extensibility;
 using org::eclipse::cyclonedds::core::cdr::encoding_version;
+using org::eclipse::cyclonedds::core::cdr::key_mode;
 using org::eclipse::cyclonedds::topic::TopicTraits;
 
-template<typename T, class S>
-bool get_serialized_size(const T& sample, bool as_key, size_t &sz);
+template<typename T, class S, key_mode K>
+bool get_serialized_size(const T& sample, size_t &sz);
 
 template<typename T>
 bool to_key(const T& tokey, ddsi_keyhash_t& hash)
@@ -64,7 +65,7 @@ bool to_key(const T& tokey, ddsi_keyhash_t& hash)
   {
     basic_cdr_stream str(endianness::big_endian);
     size_t sz = 0;
-    if (!get_serialized_size<T, basic_cdr_stream>(tokey, true, sz)) {
+    if (!get_serialized_size<T, basic_cdr_stream, key_mode::sorted>(tokey, sz)) {
       assert(false);
       return false;
     }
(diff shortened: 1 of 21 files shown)


src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
```diff
@@ -316,7 +319,7 @@ bool serialize_into(void *buffer,
   S str;
   str.set_buffer(calc_offset(buffer, CDR_HEADER_SIZE), buf_sz-CDR_HEADER_SIZE);
   return (write_header<T,S>(buffer)
-        && write(str, sample, as_key)
+        && write(str, sample, as_key ? key_mode::unsorted : key_mode::not_key)
```
eboasson, line 322: I think that it would make sense to use `key_mode` as an argument here, too, rather than `as_key`: here you have a conditional expression, and at the call sites you either have the inverse conditional or a constant.

src/ddscxx/src/org/eclipse/cyclonedds/core/cdr/cdr_stream.cpp
```diff
@@ -144,6 +133,45 @@ bool cdr_stream::finish_member(const entity_properties_t &props, member_id_set &
...
 
+void cdr_stream::set_mode(stream_mode mode, key_mode key)
+{
+  assert(key != key_mode::unset);
+  m_mode = mode;
+  m_key = key;
+  reset();
+}
+
+size_t cdr_stream::incr_position(size_t incr_by)
```
eboasson, line 144: I image this function (and perhaps some of the others that were moved here) gets called very often and it is still very simple. I would expect inlining this one to have performance benefit.

(Also: `align` which definitely gets called often. That one is fairly complex, but specializing it for read vs write and doing the complex bits in a separate function might be interesting.)

src/ddscxx/src/org/eclipse/cyclonedds/core/cdr/extended_cdr_v2_ser.cpp
```diff
@@ -217,7 +217,9 @@ const entity_properties_t* xcdr_v2_stream::first_entity(const entity_properties_
 bool xcdr_v2_stream::read_em_header(entity_properties_t &props)
 {
   uint32_t emheader = 0;
-  if (!read(*this,emheader))
+  size_t align_bytes = (4-(position()%4))%4;
```
eboasson, line 220: I was surprised by this, because it would seem likely that `read` takes care of aligning the stream and of checking whether sufficient data remains available. If I am not mistaken, it does. So something interesting that I'm not seeing must be going on here.

eboasson (commented):
LGTM with a few minor comments. I'd be interested what your take is on these points.

src/ddscxx/src/org/eclipse/cyclonedds/core/cdr/extended_cdr_v2_ser.cpp
reicheratwork: The issue was that this needed to be done in two steps, first a "peek" to check whether there is enough space at the current stream level (as indicated by d-header) and ONLY if there is, then "take" the data from the stream into the emheader.

The old approach immediately tries to "take" the emheader from the stream, which might raise an error if there is not enough bytes left, which causes the streamer to enter the error state, and reject further deserialization.

So everything was working as intended, just a little bit too well.

src/ddscxx/src/org/eclipse/cyclonedds/core/cdr/cdr_stream.cpp
reicheratwork: Inlining some of these functions will probably improve performance, though to what degree, I am not certain.

Compilers are very smart these days, and I suspect that especially for small functions, the inlining is being done implicitly for any significant optimization.

src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp
reicheratwork: I will improve on this.

reicheratwork:
Suggestions implemented in 3ff3721

To keep this PR within a well defined scope the optimizations like inlining will be for a future PR

PR #439, opened 2023-08-01
Replace type_discovery build option with typelib

dpotman:
This replaces the `enable_type_discovery` build option with `enable_typelib`, which reflects a similar change in Cyclone C. Additionally a missing include directive for `ddsi_typelib.h` is added.


dpotman:
The force-push to [87cb93f](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/commit/87cb93f571fd232df4f79048153747ebcaf26fb5) is because an `#ifdef` was missing for including `ddsi_typelib.h` in `TopicTraits.hpp`

reicheratwork:
This very much agrees with what my own local solution to the desynchonization between CycloneDDS-C and CXX came out to be.

Therefore this PR looks good to me, and can be merged.

eboasson (approved):

PR #445, opened 2023-08-21
Append IDL MD5 hash to create unique header guard

wadehunk:
This commit utilizes the MD5 hash for IDL source files and utilizes the result to generate a unique header guard macro when combined with the header file path.

Fixes: eclipse-cyclonedds/cyclonedds#1312
Requires: eclipse-cyclonedds/cyclonedds#1806


eboasson (approved):

eboasson:
Thanks @wadehunk !

PR #458, opened 2023-11-07
Fix clang to newer version in Ubuntu 22 pipelines

reicheratwork:
Following this declaration:
https://github.com/actions/runner-images/issues/8263 updated the Ubuntu 22.04 clang pipelines to use clang(++) version 13 as Github only wants to support the latest 3 versions of clang, which are 13,14,15 starting 25.09.2023


eboasson (approved):
Thanks @reicheratwork !

Issue #457, opened 2023-11-06
Equality operator returns false for an equal empty union 


jordanLatta155:
I have an IDL union with an enum discriminator with one of the enum cases being empty. The generated C++ code's == operator will return false when comparing two "empty" unions , even if their discriminator is the same.

example:
IDL:
```idl
enum AllocationState {
    Free,
    Allocated,
    Offline
};

union AllocationStateUnion switch (AllocationState) {
    case Allocated:
    case Offline:
        string reason;
};
```

C++:
```C++
bool operator==(const AllocationStateUnion& _other) const
{
  if(_d() != _other._d()) return false;
    switch(_d()) {
      case AllocationState::Allocated:
      case AllocationState::Offline:
        return reason() == _other.reason();
  }
  return false;
}
```

The last `return false` in the generated C++ is the branched reached when both `_d()` are equal to `AllocationState::Free`.

eboasson:
Interesting ... makes me wonder if it shouldn't simply end in `return true;` instead. @reicheratwork am I barking up the right tree?

reicheratwork:
Let me have a look at this

reicheratwork:
Also found some other issues that were exposed by creating a unittest that used unions with default cases, fixing those as well :)

@jordanLatta155 , can I count on you to verify my fix and review my PR when I create it?

PR #459, opened 2023-11-07
Fix issue #457

reicheratwork:
This fixes the issue of comparison operators not returning true for unions with default cases not associated with a type for said default case
Also added unittests for union comparison operators

This PR will fail the Ubuntu + Clang builds until this #458 PR has been merged.

@jordanLatta155 , could you give this one a look please?

### DIFF
--- a/src/ddscxx/tests/Regression.cpp
+++ b/src/ddscxx/tests/Regression.cpp
@@ -404,4 +404,63 @@ TEST_F(Regression, unaligned_access)
   ASSERT_EQ(s.ll(), int64_t(0x08090A0B0C0D0E0F));  //size 4 reads should be done at 4 byte offsets in stream
 }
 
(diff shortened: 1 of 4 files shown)


eboasson (approved):
LGTM, thanks @reicheratwork !

@jordanLatta155 it'd be great if you could confirm it works for you as well 🙂

jordanLatta155:
@reicheratwork, @eboasson, confirmed that this worked for me.

Additionally, adding the default case to the switch statements fixed another issue I had where warnings were triggered from not handling all possible cases.

Thanks for the quick turnaround!

eboasson:
Thanks @jordanLatta155 ☺️

Successful CI run with the updated CI build settings: https://dev.azure.com/eclipse-cyclonedds/cyclonedds-cxx/_build/results?buildId=5351&view=results

PR #471, opened 2024-02-21
Fix hash population.  Closes #470

### LINKED ISSUE #470: datatopic.hpp: populate_hash() does not populate the hash
CycloneDDS-CXX e41e5c87192423e3443b68ebed064f4debdcbea6 06 Nov 23

The following logic results in a failure to generate an instance's hash value, leading to an OOM in the underlying hopscotch hash table due to hash table key collisions.

Related:  https://github.com/eclipse-cyclonedds/cyclonedds/issues/1932

[populate_hash](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/master/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L880)

    template <typename T>
    void ddscxx_serdata<T>::populate_hash()
    {
      if (hash_populated)
        populate_hash(*getT());
    }

Solution:  if the hash is *not* populated, populate the hash.

clunietp:


eboasson (approved):
Ouch, that was a nasty one. Thank you very much @clunietp !

PR #472, opened 2024-02-26
Feature/generate ostream operators

### LINKED ISSUE #467: Generate streaming operators for std::ostream
It is a common task to somehow log the received samples.
For this `std::ostream` streaming operators are needed which are currently (v0.10.4) not generated from the `.idl` files.
Other Vendors do also generate streaming operators, for cyclone we have to manually create them currently.

So every datatype shall have a streaming operator.

Examples
```cpp
// Stream all class members
std::ostream& operator<<(std::ostream& s, T_TextMarkupSupportedType const& rhs)
{
    return s << "["
             << "A_boldSupported: " << (rhs.A_boldSupported())
             << ", A_italicSupported: " << (rhs.A_italicSupported())
             << ", A_strikethroughSupported: " << (rhs.A_strikethroughSupported())
             << ", A_highlightSupported: " << (rhs.A_highlightSupported())
             << ", A_boldAndItalicSupported: " << (rhs.A_boldAndItalicSupported())
             << ", A_specialCharactersSupported: " << (rhs.A_specialCharactersSupported())
             << "]";
}

// Stream enums
std::ostream& operator<<(std::ostream& s, T_HorizontalAlignmentType const& rhs)
{
    switch (rhs)
    {
        case P_Display_Extension_PSM::T_HorizontalAlignmentType::L_HorizontalAlignment__Invalid:
            s << "T_HorizontalAlignmentType::L_HorizontalAlignment__Invalid"; break;
        case P_Display_Extension_PSM::T_HorizontalAlignmentType::L_HorizontalAlignment__Default:
            s << "T_HorizontalAlignmentType::L_HorizontalAlignment__Default"; break;
        case P_Display_Extension_PSM::T_HorizontalAlignmentType::L_HorizontalAlignment__Left:
            s << "T_HorizontalAlignmentType::L_HorizontalAlignment__Left"; break;
        case P_Display_Extension_PSM::T_HorizontalAlignmentType::L_HorizontalAlignment__Centre:
            s << "T_HorizontalAlignmentType::L_HorizontalAlignment__Centre"; break;
        case P_Display_Extension_PSM::T_HorizontalAlignmentType::L_HorizontalAlignment__Right:
            s << "T_HorizontalAlignmentType::L_HorizontalAlignment__Right"; break;
        default: break;
    }
    return s;
}
```

trittsv:
fixes #467

This pull requests generates `ostream` operators for each type.

### DIFF
--- a/examples/helloworld/publisher.cpp
+++ b/examples/helloworld/publisher.cpp
@@ -54,7 +54,7 @@ int main() {
         HelloWorldData::Msg msg(1, "Hello World");
 
         /* Write the message. */
-        std::cout << "=== [Publisher] Write sample." << std::endl;
+        std::cout << "=== [Publisher] Write sample: " << msg << std::endl;
         writer.write(msg);
 
         /* With a normal configuration (see dds::pub::qos::DataWriterQos
--- a/examples/helloworld/subscriber.cpp
+++ b/examples/helloworld/subscriber.cpp
@@ -72,9 +72,7 @@ int main() {
                      * is set. The other data parts are not.
                      * Check if this sample has valid data. */
                     if (info.valid()) {
-                        std::cout << "=== [Subscriber] Message received:" << std::endl;
-                        std::cout << "    userID  : " << msg.userID() << std::endl;
-                        std::cout << "    Message : \"" << msg.message() << "\"" << std::endl;
+                        std::cout << "=== [Subscriber] Message received: " << msg << std::endl;
 
                         /* Only 1 message is expected in this example. */
                         poll = false;
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -943,7 +943,28 @@ generate_includes(const idl_pstate_t *pstate, struct generator *generator)
   if (fputs(fmt, generator->header.handle) < 0)
     return IDL_RETCODE_NO_MEMORY;
 
-  if (generator->uses_sequence || generator->uses_bounded_sequence || generator->uses_array)
+  if (generator->uses_array)
+  {
+    fmt = "template <typename T, size_t N>\n"
+          "std::ostream& operator<<(std::ostream& os, std::array<T, N> const& rhs)\n"
+          "{\n"
+          "  os << \"[\";\n"
+          "  for (std::size_t i = 0; i < N; i++)\n"
+          "  {\n"
+          "    os << rhs[i];\n"
+          "    if (i < N - 1)\n"
+          "    {\n"
+          "      os << \", \";\n"
+          "    }\n"
+          "  }\n"
+          "  os << \"]\";\n"
+          "return os;\n"
+          "}\n\n";
+    if (fputs(fmt, generator->header.handle) < 0)
+      return IDL_RETCODE_NO_MEMORY;
+  }
+
+  if (generator->uses_sequence || generator->uses_bounded_sequence)
   {
     fmt = "template<typename T>\n"
         "std::ostream& operator<<(std::ostream& os, std::vector<T> const& rhs)\n{\n"
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -819,6 +819,41 @@ register_bitmask(
   return IDL_RETCODE_OK;
 }
 
+const char *seq_tmpl = "std::vector<{TYPE}>";
+const char *seq_inc = "<vector>";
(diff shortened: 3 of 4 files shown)


eboasson:
I actually started looking at it yesterday afternoon ☺️ The first impression was fine, but there are a few things I want to check/make sure I understand:

* Why the bump from C++14 to C++17 is needed. (I am fine with it, I think ROS2 did that, too, but I'd like to understand it. I also want to be sure can still build C++11 code with it in a pinch.)
* Whether it plays nice with the possibility of using some other type than `std::vector<T>` and similar cases (so `-f sequence-template`, `-f string-template`, etc.)
* Whether it can easily be used to (finally) implement [`serdata_print`](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/2784662ecabddd27f635ea2609e6816aa294d604/src/ddscxx/include/org/eclipse/cyclonedds/topic/datatopic.hpp#L621)
* What the generated code looks like 🙂

You can probably answer some of these straightaway, but as I like to really check any code that goes in you can only save me so much time by providing them 😂

trittsv:
ah great @eboasson, here are my answers:

- i had to bump some to c++17 where `(generator->uses_array || generator->uses_sequence || generator->uses_bounded_sequence || generator->uses_optional)` because of the usage of `std::vector`. Otherwise i would have to split the file [rc/ddscxx/include/org/eclipse/cyclonedds/util/ostream_operators.hpp](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/472/files#diff-30d488fe78033d490c9cdb312a1a33de255c32b2519e1e29f4bf899d257307e0) into multiple files where each containing only vector, array, etc. None of the examples used optional so it was not necessary so far. But i can split the file if you like to remove the c++17 from the examples?
- `-f sequence-template, -f string-template` should work fine. The only thing that could happen is that the custom type may need to provide a ostream operator.
- What is the usecase for `serdata_print`? 
Where can i call the function to test it?
```cpp
template <typename T>
size_t serdata_print(
  const ddsi_sertype* tpcmn, const ddsi_serdata* dcmn, char* buf, size_t bufsize)
{
  (void)tpcmn;
  (void)dcmn;
  //implementation to follow!!!

  std::stringstream ss;
  ss << static_cast<T*>(dcmn->loan->sample_ptr);
  std::string sample_as_string = ss.str();

  std::cout << "sample printed" << sample_as_string << std::endl;

  if (bufsize > 0)
    buf[0] = 0x0;
  return 0;
}
```
- Here are examples of the generated code (the idls from the ddscxx tests)
[generated-code.zip](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/14574585/generated-code.zip)

eboasson (approved):

eboasson:
Good that I asked: another look at the changes with your answers in mind sufficed 😀

> * i had to bump some to c++17 where `(generator->uses_array || generator->uses_sequence || generator->uses_bounded_sequence || generator->uses_optional)` because of the usage of `std::optional`. Otherwise i would have to split the file [rc/ddscxx/include/org/eclipse/cyclonedds/util/ostream_operators.hpp](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/pull/472/files#diff-30d488fe78033d490c9cdb312a1a33de255c32b2519e1e29f4bf899d257307e0) into multiple files where each containing only vector, array, optional etc. None of the examples used optional so it was not necessary so far. But i can split the file if you like to remove the c++17 from the examples?

I think especially for the examples there is no point in avoiding C++17. Just about everyone has access to a C++17 compiler these days, so it is only the older embedded and/or regulated and/or corporate stuff that might not be able to use it for production code. Supporting those cases is nice, but to put effort into examples ... nah.

> * `-f sequence-template, -f string-template` should work fine. The only thing that could happen is that the custom type may need to provide a ostream operator.

I now see what you mean. I think that's a perfectly sensible expectation.

> * What is the usecase for `serdata_print`?

The trace files can include the content of samples, that's when it uses `serdata_print`. In other situations (involving OpenSplice) I've found it to be a *very* useful feature when trying to understand what exactly is going on in some complicated scenario involving a lot of different things, and so I retained it in Cyclone.

If you do `Tracing/Category` = `trace` and look for lines with `write_sample` or `application` you can easily find cases where it dumps the contents. If you take the C "hello world", you have the contents, if you use the C++ one, you see no contents.

With your suggested implementation, that should change.

I do consider that separate from this PR: to me, this PR simply makes that possible, and there's no point in delaying the merging of it because it enables this small improvement.

PR #480, opened 2024-04-03
Implement QosProvider for C++ API.

e-hndrks:
Add QosProvider functionality to C++ in accordance with the work done for https://github.com/eclipse-cyclonedds/cyclonedds/pull/1955.

### DIFF
--- a/.azure/templates/build-test.yml
+++ b/.azure/templates/build-test.yml
@@ -118,6 +118,7 @@ steps:
       cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
             -DCMAKE_INSTALL_PREFIX=install \
             -DCMAKE_PREFIX_PATH="${BUILD_SOURCESDIRECTORY}/cyclonedds/build/install;${BUILD_SOURCESDIRECTORY}/iceoryx/build/install;${BUILD_SOURCESDIRECTORY}/googletest/build/install" \
+            -DCYCLONE_SOURCE_DIR="${BUILD_SOURCESDIRECTORY}/cyclonedds" \
             -DANALYZER=${ANALYZER:-off} \
             -DSANITIZER=${SANITIZER:-none} \
             -DENABLE_ICEORYX=${ICEORYX:-off} \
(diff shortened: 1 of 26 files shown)


src/ddscxx/src/org/eclipse/cyclonedds/domain/qos/DomainParticipantQosDelegate.cpp
```diff
@@ -103,8 +103,7 @@ DomainParticipantQosDelegate::check() const
 bool
 DomainParticipantQosDelegate::operator ==(const DomainParticipantQosDelegate& other) const
 {
-    return other.present_             == present_ &&
```
eboasson, line 106: This change in the behaviour of `operator==` (and the same for the other QoS types) I think warrants its own commit with its own justification for no longer checking whether the `present` flags are the same.

I do suspect there is a difference before-and-after, because I suspect that:
```
 f(const DataReaderQos& qa, const DataReaderQos& qb) {
  Topic<T> t(..., Reliability::Reliable());
  DataReader<T> a(..., qa);
  DataReader<T> b(..., qb);
  return a.qos() == b.qos();
}
```
will show a change in behaviour if, e.g., `qa` has `Reliability` set to `BestEffort` (the default) and `qb` has not set reliability, because in that case `b` will inherit the reliability setting from the topic, which has it set to `Reliable` (not the default).

It may be that it is currently impossible to make a sparse QoS object in C++ (I just don't know the C++ binding well enough to be sure even if I do try to check). In that case, there is no observable change in behaviour given the available interfaces.

Otherwise, I think that instead of changing `operator==` there should be an operation to complete the QoS with the default values, so that you could do `qa.dense() == qb.dense()` vs `qa == qb`. (Not that I am certain that `dense` is the best name, it is just the first thing that came to mind because I think of the QoS objects as "sparse" things.)

Then, whatever prompted this change (I suspect it is test code) could take the detour via this `dense()` function.

.azure/templates/build-test.yml
```diff
@@ -118,6 +118,7 @@ steps:
       cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
             -DCMAKE_INSTALL_PREFIX=install \
             -DCMAKE_PREFIX_PATH="${BUILD_SOURCESDIRECTORY}/cyclonedds/build/install;${BUILD_SOURCESDIRECTORY}/iceoryx/build/install;${BUILD_SOURCESDIRECTORY}/googletest/build/install" \
+            -DCYCLONE_SOURCE_DIR="${BUILD_SOURCESDIRECTORY}/cyclonedds" \
```
eboasson, line 121: I suspect this is needed for including `dds__sysdef_model.h` in the test code, but it is really inelegant to require the C sources to be present to build the C++ binding — we never had that tight a relationship, and I don't think the QoS provider should change that.

Now, I haven't made an effort yet to find out which bits exactly are needed but it appears likely that only a fraction of what is in that file is needed. Before merging it like this, I'd like to know whether the bits that are needed can reasonably be installed by the core library. That would eliminate the need for this hack.

(Apologies for not spotting this earlier.)

eboasson (commented):
Two more thingummies, I'm afraid ...

Issue #484, opened 2024-04-14
ostream headers 


mav5499:
## To Reproduce
```
module Measurements
{
    typedef octet NumericGUID[16];
};
```
```
```

## Expected Behavior

mav5499:
Attaching the IDLs used and generated cpp types:
[generate-test.zip](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/files/14971541/generate-test.zip)

PR #485, opened 2024-04-14
Moves ostream_operators.hpp include to all implementation files

michelle-vu19:
Resolves issue #484

### DIFF
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -538,7 +538,7 @@ static bool sc_union(const idl_union_t *_union)
 
   const idl_case_t *_case = NULL;
   IDL_FOREACH(_case, _union->cases) {
-    if (!is_selfcontained(_case->type_spec))
+    if (is_external(_case) || !is_selfcontained(_case->type_spec))
       return false;
   }
 
@@ -549,7 +549,7 @@ static bool sc_struct(const idl_struct_t *str)
 {
   const idl_member_t *mem = NULL;
   IDL_FOREACH(mem, str->members) {
-    if (!is_selfcontained(mem->type_spec))
+    if (is_external(mem) || !is_selfcontained(mem->type_spec))
       return false;
   }
 
@@ -563,7 +563,7 @@ bool is_selfcontained(const void *node)
 {
   if (idl_is_sequence(node)
    || idl_is_string(node)
-   || idl_is_optional(node)) {
+   || is_optional(node)) {
     return false;
   } else if (idl_is_typedef(node)) {
     return is_selfcontained(((const idl_typedef_t*)node)->type_spec);
@@ -575,9 +575,8 @@ bool is_selfcontained(const void *node)
     const idl_node_t *parent = ((const idl_node_t*)node)->parent;
     assert (idl_is_typedef(parent));
     return is_selfcontained(parent);
-  } else {
-    return !is_external(node);
   }
+  return true;
 }
 
 idl_extensibility_t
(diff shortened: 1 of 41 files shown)


michelle-vu19:
@trittsv Good morning, can you take a look at this?

trittsv:
Hey @michelle-vu19, seems that i missed that case described in #484.

I put the include of the `ostream_operators.hpp` explicitly into the `cpp` file.
Because this file modifies the users std namespace and that is dangerous.
So i chose to encapsulate it in the cpp files.

I strongly recommend to keep it in the cpp files.

My suggestion is:
Iterate all the included idl files and [check if they are using std types](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/a5898dbafdd0da1ec9b19cd55ae31ea4c493b0fb/src/idlcxx/src/generator.c#L902) and include the `ostream_operators.hpp` in the cpp files there too.

michelle-vu19:
@trittsv Thank you for the suggestion!

jkav77:
I'm running into this problem as well. Manually adding `#include <org/eclipse/cyclonedds/util/ostream_operators.hpp>` to the files that need it seem to help. I have almost 700 IDL files in my project so this is suboptimal. Is there a fix for this?

src/idlcxx/src/generator.c
```diff
@@ -901,11 +901,11 @@ generate_includes(const idl_pstate_t *pstate, struct generator *generator)
...
   {
-    // ostream cpp
+    // ostream hpp
     // streaming operators for the used std types
     const char *fmt;
     fmt = "#include <org/eclipse/cyclonedds/util/ostream_operators.hpp>\n\n";
-    if (fputs(fmt, generator->impl.handle) < 0)
+    if (fputs(fmt, generator->header.handle) < 0)
       return IDL_RETCODE_NO_MEMORY;
   }
```
reicheratwork, line 911: > Hey @michelle-vu19, seems that i missed that case described in #484.
> 
> I put the include of the `ostream_operators.hpp` explicitly into the `cpp` file. Because this file modifies the users std namespace and that is dangerous. So i chose to encapsulate it in the cpp files.
> 
> I strongly recommend to keep it in the cpp files.
> 
> My suggestion is: Iterate all the included idl files and [check if they are using std types](https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/a5898dbafdd0da1ec9b19cd55ae31ea4c493b0fb/src/idlcxx/src/generator.c#L902) and include the `ostream_operators.hpp` in the cpp files there too.

I agree with @trittsv on this point. Moving the `ostream_operators.hpp` include from the implementation to the header file might give some unintended side effects, especially since the insertion operators are inside `std` namespace.

The solution I suggest might be a little crude, but it works: just always include the `ostream_operators.hpp` in the implementation file. Since the functions included are template functions, if they're not used, no binary code is generated and since they're in the implementation file, they won't cause any contamination of the `std` namespace.

```suggestion
  // ostream cpp
  // streaming operators for the used std types
  static const char *fmt = "#include <org/eclipse/cyclonedds/util/ostream_operators.hpp>\n\n";
  if (fputs(fmt, generator->impl.handle) < 0)
    return IDL_RETCODE_NO_MEMORY;

  return IDL_RETCODE_OK;
```

michelle-vu19:
@trittsv @reicheratwork @eboasson Sorry for taking so long to apply this suggestion :) Can I get a review on this? I am currently in the process of compiling CPP types just to verify that this works.

trittsv:
hey @michelle-vu19, looks good, you still need to sign the eclipsefdn/eca to be able to merge

michelle-vu19:
@trittsv Done!

michelle-vu19:
Update: my code generated successfully with the new changes :)

reicheratwork (approved):
Looks good to me!

yaniswav (approved):
This change works fine for me! Resolving ostream_operators issues

michelle-vu19:
When can this be merged? :)

eboasson (approved):
I am so sorry to have kept you waiting for so long @michelle-vu19

Issue #461, opened 2023-11-15
DataWriter fails to write message


plament:
I am using `0.10.4` and have the following `message.idl`

```cpp
struct Message
{
    @optional int32 a0;
    @optional int32 a1;
};
```

Then I generate the data types using: `idlc -b . -l <path to libcycloneddsidlcxx> -x final message.idl`

Then I create 2 messages and try to write them using `DataWriter`. However sending the second message fails with `write failed` exception.

Here is my code:

```cpp
#include <iostream>
#include "dds/dds.hpp"
#include "gen/message.hpp"

using namespace org::eclipse::cyclonedds;

int main() {

    std::cout << "=== [Publisher] Create writer." << std::endl;

    dds::domain::DomainParticipant participant(0);
    dds::topic::Topic<Message> topic(participant, "TOPIC");
    dds::pub::Publisher publisher(participant);
    dds::pub::DataWriter<Message> writer(publisher, topic);

    Message m1;
    m1.a1() = 10;

    Message m2;
    m2.a0() = 10;

    try {
        std::cout << "=== Will write m1" << std::endl;
        writer.write(m1);

        std::cout << "=== Will write m2" << std::endl;
        writer.write(m2);

        std::cout << "=== Write done" << std::endl;
    } catch (const std::exception& ex) {
        std::cerr << "=== [Publisher] Exception: " << ex.what() << std::endl;
        return EXIT_FAILURE;
    }

    return EXIT_SUCCESS;
}
```

The output is:

```
=== [Publisher] Create writer.
=== Will write m1
=== Will write m2
=== [Publisher] Exception: Error Bad Parameter - write failed.
===============================================================================
Context     : void org::eclipse::cyclonedds::pub::AnyDataWriterDelegate::write
Node        : UnknownNode
```

So writing `m2` fails. However if I first write `m2` and then `m1` then everything works fine.

The actual method which fails is:

```
bool cdr_stream::bytes_available(size_t N, bool peek)
```

For some reason there is no space left in the `cdr_stream` buffer to serialize the second message. 

@eboasson

plament:
Seems the issue is that the serialized size is not calculated properly.

```
template<typename T, class S>
bool get_serialized_size(const T& sample, bool as_key, size_t &sz)
{
  if (TopicTraits<T>::isSelfContained()) {
    if ((as_key && !get_serialized_fixed_size<T,S,true>(sample, sz)) ||
[...]

PR #486, opened 2024-04-25
repair issue #461

ruoruoniao:


eboasson (commented):
Thank you @ruoruoniao for looking into #461 (I seem to have forgotten it ...). I am surprised that this change would fix it, though because the difference between `idl_is_optional` and `is_optional` seems to be that:
* `idl_is_optional` checks whether *this* "node" is marked as optional,
* `is_optional` walks the typedef chain to see if any of those nodes are marked as optional,
and but `is_selfcontained` itself also walks that typedef chain, and so I would expect the two to behave the same.

Could you please help me understand?

Two other small things:
* It is much easier for me (or anyone trying to find a correlation between a problem and a fix) if the commit subject and the PR title briefly explain what it is about, rather than just refer to a number. If the subject is something mentioning optionals and self-contained it'll be much easier.
* The Eclipse Foundation has a rule that commit authors must have signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) before it can be merged.

ruoruoniao:
> Thank you @ruoruoniao for looking into #461 (I seem to have forgotten it ...). I am surprised that this change would fix it, though because the difference between `idl_is_optional` and `is_optional` seems to be that:
> 
> * `idl_is_optional` checks whether _this_ "node" is marked as optional,
> * `is_optional` walks the typedef chain to see if any of those nodes are marked as optional,
>   and but `is_selfcontained` itself also walks that typedef chain, and so I would expect the two to behave the same.
> 
> Could you please help me understand?
> 
> Two other small things:
> 
> * It is much easier for me (or anyone trying to find a correlation between a problem and a fix) if the commit subject and the PR title briefly explain what it is about, rather than just refer to a number. If the subject is something mentioning optionals and self-contained it'll be much easier.
> * The Eclipse Foundation has a rule that commit authors must have signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php) before it can be merged.

Well, I have an idl like this:
```
@final struct OptionalTest
{
    uint64 id;
    uint64 eventID;
    @optional uint64 errorID;
};
This structure member will convert to this:
```
{
    //node(OptionalTest)
    members: [
        { type_spec: node(id), ... },
        { type_spec: node(eventID), ... },
        { type_spec: node(errorID), ... }
    ],
    ......
}
{
    //node(errorID)
    mask: 1551,
    idl_is_optional(this)     : false,
    idl_is_member(this)     : false,
    idl_is_declarator(this)  : false,
    idl_is_type_spec(this)  : true,
    parent: node(optional), 
    ....
}
{
    //node(optional)
    idl_is_optional(this) : true,
    idl_is_member(this) : true,
    ....
}
```
So if call `idl_is_optional`, `node(errorID)` is not a member or a declarator, it cannot route to node(errorID)->parent. Also, add a if-else branch like:
```
...
else if (idl_is_type_spec(node)) {
    const idl_node_t *parent = ((const idl_node_t*)node)->parent;
    return is_selfcontained(parent);
}
...
```

eboasson:
Thanks for the explanation! I completely forgot about the typedef-vs-type-spec thing. Yes, then it is obvious that it makes a difference.

I prefer your second version, where you keep the call to `idl_is_optional` and add
> ```
> else if (idl_is_type_spec(node)) {
>     const idl_node_t *parent = ((const idl_node_t*)node)->parent;
>     return is_selfcontained(parent);
> }
> ```
If you change it this way I'll be happy to merge it.

The reason is simply that the current version looks "off" and I expect it will cause people (me?) to go down the same rabbit hole again in the future. It won't solve the underlying problem that such mistakes are just too easy to make with the way types are represented internally in `idlc`, but fixing that is quite a project ... so for the time being the best we can do is fix the problems that crept in.

PR #487, opened 2024-04-30
repair issue #461

ruoruoniao:
except branch idl_is_type_spec(node)

As you say, change a fix way~ @eboasson

### DIFF
--- a/src/idlcxx/src/generator.c
+++ b/src/idlcxx/src/generator.c
@@ -23,18 +23,16 @@
 
 #include "generator.h"
 
-static size_t istok(const char *str, size_t len, const char **toks)
-{
+static size_t istok(const char *str, size_t len, const char **toks) {
   for (size_t num = 0; toks && toks[num]; num++) {
     if (strncmp(toks[num], str, len) == 0)
-      return num+1;
+      return num + 1;
   }
   return 0;
 }
 
 static int makefmtp(
-  char **fmtp, const char *str, const char **toks, const char **flags)
-{
+  char **fmtp, const char *str, const char **toks, const char **flags) {
   char buf[64], *fmt;
   size_t len = 0;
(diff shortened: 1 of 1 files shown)


eboasson:
Every time I think I have wrapped my head around `idlc`'s inside well enough to review PRs and do so maintenance on it, it seems to catch me in yet another bit of weirdness. Who would have thought that this runs into an infinite recursion between alternating between `y = x->type_spec` and `x = y->parent` ... I did not anyway.

That means what seemed like the less surprising solution doesn't work and your original proposal is infinitely better, thanks to the absence of infinite recursion. Can you please revive it? I am sorry to have caused you this much trouble.

ruoruoniao:
> Every time I think I have wrapped my head around `idlc`'s inside well enough to review PRs and do so maintenance on it, it seems to catch me in yet another bit of weirdness. Who would have thought that this runs into an infinite recursion between alternating between `y = x->type_spec` and `x = y->parent` ... I did not anyway.
> 
> That means what seemed like the less surprising solution doesn't work and your original proposal is infinitely better, thanks to the absence of infinite recursion. Can you please revive it? I am sorry to have caused you this much trouble.

Sure, I am quite surprised when I found the pipeline return error...I don't have a good solution either, unless we can record the path that the function runs on, and terminate it when it encounters the same node, but that's too strange...

eboasson (commented):
I think an over-active auto-formatter decided to interfere with this PR — there is the intended change (`idl_is_optional` → `is_optional`) and there is a whole lot of whitespace changes. Can you please update the PR to not have the whitespace changes? Thanks!

ruoruoniao:
Of course (￣▽￣)"

eboasson (approved):
Thank you @ruoruoniao ☺️

PR #495, opened 2024-06-27
Fix UAF on listener arg on event during enable()

### LINKED ISSUE #493: Random crash when create DataReader/DataWriter...
## Problem descriptor

I found a random crash when create a DataWriter or a DataReader, it crush in:
```cpp
/* src/ddscxx/src/org/eclipse/cyclonedds/core/ListenerDispatcher.cpp : 191 */
    if (la->reset_on_invoke && la->cpp_ref->obtain_callback_lock()){  // crash because la->cpp_ref is deleted.
        ...
    }
    ...
``` 

## Minimal Reproducible Example:
```cpp
//Publisher
#include "dds/dds.hpp"
#include "HelloWorldData.hpp"

using namespace org::eclipse::cyclonedds;

int main() {
    dds::domain::DomainParticipant participant(13);
    dds::topic::Topic<HelloWorldData::Msg> topic(participant, "topic");
    auto p_qos = participant.default_publisher_qos();
    dds::pub::Publisher publisher(participant, p_qos);
    auto qos = publisher.default_datawriter_qos();
    class Listener : public dds::pub::NoOpDataWriterListener<HelloWorldData::Msg> {
    public:
        void on_publication_matched(dds::pub::DataWriter<HelloWorldData::Msg> & dataWriter,
            const dds::core::status::PublicationMatchedStatus &) override {
            std::cout << dataWriter->get_ddsc_entity() << " matched." << std::endl;
        }
    };
    auto listener = new Listener();
    while(true) {
        dds::pub::DataWriter<HelloWorldData::Msg> writer(publisher, topic, qos, listener, dds::core::status::StatusMask::all());
    }
}
```
```cpp
// Subscriber
#include "dds/dds.hpp"
#include "HelloWorldData.hpp"

using namespace org::eclipse::cyclonedds;

int main() {
    dds::domain::DomainParticipant participant(13);
    dds::topic::Topic<HelloWorldData::Msg> topic(participant, "topic");
    auto s_qos = participant.default_subscriber_qos();
    dds::sub::Subscriber subscriber(participant, s_qos);
    auto qos = subscriber.default_datareader_qos();
    class Listener : public dds::sub::NoOpDataReaderListener<HelloWorldData::Msg> {
    public:
        void on_subscription_matched(dds::sub::DataReader<HelloWorldData::Msg> &dataReader,
            const dds::core::status::SubscriptionMatchedStatus &) override {
            std::cout << dataReader->get_ddsc_entity() << " matched." << std::endl;
        }
    };
    auto listener = new Listener();
    while(true) {
        dds::sub::DataReader<HelloWorldData::Msg> reader(subscriber, topic, qos, listener, dds::core::status::StatusMask::all());
    }
}
```

## Problem location

I found the problem occurs here(eg. with DataWriter, also in DataReader):
```cpp
 /* src/ddscxx/include/dds/pub/DataWriter.hpp : 170 */
DataWriter(const dds::pub::Publisher& pub,
                  const ::dds::topic::Topic<T>& topic,
                  const dds::pub::qos::DataWriterQos& qos,
                  dds::pub::DataWriterListener<T>* listener = NULL,
                  const dds::core::status::StatusMask& mask = ::dds::core::status::StatusMask::none()); //This index with (1)
```

```cpp
 /* src/ddscxx/include/dds/pub/detail/DataWriter.hpp : 62 */
DataWriter(const dds::pub::Publisher& pub,
               const ::dds::topic::Topic<T>& topic,
               const dds::pub::qos::DataWriterQos& qos); (2)
```

```cpp
    /* src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp : 392 */
    this->listener_set(nullptr, dds::core::status::StatusMask::all(), false);
    dds_entity_t ddsc_writer = dds_create_writer (ddsc_pub, ddsc_topic, ddsc_qos, this->listener_callbacks); //This index with (3)
    dds_delete_qos(ddsc_qos);
    ...
```

```cpp
    /* src/ddscxx/src/org/eclipse/cyclonedds/core/EntityDelegate.cpp : 226 */
org::eclipse::cyclonedds::core::EntityDelegate::listener_set(void *_listener, 
                                                                                            const dds::core::status::StatusMask& mask, 
                                                                                            bool reset_on_invoke) { // This index with (4)
    ....
    if (this->enabled_) // This index with (5)
    {
        dds_return_t ret;
        ret = dds_set_listener(this->ddsc_entity, callbacks);
        ISOCPP_DDSC_RESULT_CHECK_AND_THROW(ret, "Setting listener failed.");
    }
    // Delete previous ddsc listener callbacks object
    if (this->listener_callbacks != NULL)
    {
        void *prev_arg;
        dds_lget_data_available_arg(this->listener_callbacks, nullptr, &prev_arg, nullptr);
        dds_delete_listener(this->listener_callbacks);
        delete reinterpret_cast<org::eclipse::cyclonedds::core::ListenerArg *>(prev_arg); //This index with (6)
    }
    ...
```

I. When call function `(1)`, it will call `(2)` before `(1)`.
II. When call `(1)`, it will call `(3)` to set a default `dds_listener_t` to `dds_entity_t`.
III. After this, the real listener will be generate in (4).
IV. But at (5), `this->enabled_` is false, so it will not set real listener to `dds_entity_t`.
V. Then, current `dds_listener_t` which binding on `dds_entity_t` will be delete by (6).
VI. After all of above, (1) will call `enable()` to set real listener to `dds_entity_t`.

So when an event comes between V and VI, it will crashed reason of (6), arg is released.

## Solution

I. Make `this->listener_callbacks` to nullptr.
II. After (3), set `this->enable_ = true`

I don't know why there is a `this->listener_callbacks`, so I dont know how to repair it will make a lowest effect.

eboasson:
The C++ code first installs a no-op listener in the entity, then completes construction of the C++ object and only then sets the application listener.  The dds_set_listener operation then invokes the listeners for any events that happened before the application listener was installed.

This is different from not installing a no-op listener: if there is no listener, an event is propagates up the entity hierarchy, checking whether any of those has a listener installed.  If there is one, it is invoked and the event is considered handled, and the status flag cleared (leaving no trace of the event).  If the application is trying to install a listener for the event, then this results in invoking the wrong listener.

Cyclone offers a mode in which the status flag is not cleared.  This means one can define a set of listener functions that do absolutely nothing and without clearing the status flags.

The code re-used its normal callback functions.  Those inspect the listener argument, and during construction of the entity there is a window where the argument for the no-op listener is freed but the callback functions remain installed.  This fixes that by using separate functions that do nothing so that we don't have to touch the listener callbacks/argument data maintained in the C++ binding.

The normal functions used to check the "reset on invoke" flag that is used to enable/disable clearing the status flags for deciding whether or not to actually do anything.  The standard API doesn't allow this mode of not clearing the status flag on listener invocation, so it could be abused in this manner.  With the introduction of a separate set of no-op functions, the need for this abuse disappeared.  That, too, has been fixed.

Details of the UAF:

1. When creating a data writer, it will first use EntityDelegate::listener_set to set the listener_callbacks to the special "no-op" mode of the normal listener C++ functions.

2. It then updates the listener_callbacks to those of the "real" listener (again using EntityDelegate::listener_set).  The entity is not yet "enabled" and the call to the C API dds_set_listener is skipped.

3. It then frees the previous listener_callbacks with its argument.

4. If an event causing a listener invocation happens now, the callback function is called with the freed argument pointer because the C API still references it.

Thanks to @ruoruoniao for the discovery and root-cause analysis.

Fixes #493

### DIFF
--- a/src/ddscxx/CMakeLists.txt
+++ b/src/ddscxx/CMakeLists.txt
@@ -31,6 +31,7 @@ set(sources
     src/org/eclipse/cyclonedds/core/EntityDelegate.cpp
     src/org/eclipse/cyclonedds/core/ReportUtils.cpp
     src/org/eclipse/cyclonedds/core/ListenerDispatcher.cpp
+    src/org/eclipse/cyclonedds/core/NoopListener.cpp
     src/org/eclipse/cyclonedds/core/InstanceHandleDelegate.cpp
     src/org/eclipse/cyclonedds/core/EntitySet.cpp
     src/org/eclipse/cyclonedds/core/MiscUtils.cpp
--- a/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp
+++ b/src/ddscxx/include/dds/pub/detail/DataWriterImpl.hpp
@@ -29,6 +29,7 @@
 #include <dds/pub/PublisherListener.hpp>
 #include <dds/domain/DomainParticipantListener.hpp>
 #include <org/eclipse/cyclonedds/core/ListenerDispatcher.hpp>
+#include <org/eclipse/cyclonedds/core/NoopListener.hpp>
 
 namespace dds
 {
@@ -385,8 +386,7 @@ dds::pub::detail::DataWriter<T>::DataWriter(
 
     std::string name = topic.name() + "_datawriter";
 
-    this->listener_set(nullptr, dds::core::status::StatusMask::all(), false);
-    dds_entity_t ddsc_writer = dds_create_writer (ddsc_pub, ddsc_topic, ddsc_qos, this->listener_callbacks);
+    dds_entity_t ddsc_writer = dds_create_writer (ddsc_pub, ddsc_topic, ddsc_qos, org::eclipse::cyclonedds::core::make_noop_listener().get());
     dds_delete_qos(ddsc_qos);
     ISOCPP_DDSC_RESULT_CHECK_AND_THROW(ddsc_writer, "Could not create DataWriter.");
     topic_.delegate()->incrNrDependents();
--- a/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
+++ b/src/ddscxx/include/dds/sub/detail/TDataReaderImpl.hpp
@@ -23,6 +23,7 @@
 #include <dds/sub/detail/SamplesHolder.hpp>
 #include <dds/domain/DomainParticipantListener.hpp>
 #include "dds/core/macros.hpp"
+#include <org/eclipse/cyclonedds/core/NoopListener.hpp>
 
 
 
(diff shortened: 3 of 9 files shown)


ruoruoniao:
I'm very glad to see that fix works well~Thx~O(∩_∩)O

PR #497, opened 2024-07-17
Set RUNPATH in installed targets and fix roundtrip example install

PatrickM-ZS:
This sets RUNPATH (same as in [eclipse-cyclonedds/cyclonedds](https://github.com/eclipse-cyclonedds/cyclonedds) so the runtime linker can use it to find dependent libraries (like `libddsc`) when they are installed to the same directory as the C++ library.
Also found and fixed some typo's which caused the roundtrip example to not be correctly installed.


eboasson (approved):
Thank you!

PR #503, opened 2024-08-14
Do not throw an exception on times before 19700101

eboasson:
This stops throwing the C++ wrapper from throwing an exception in the dds:core::Time class on trying to construct a time stamp prior to the 00:00:00 UTC on 1 January 1970. There is no reason to do that and it makes it vulnerable to receiving samples with earlier time stamps.

It is tricky to test: Cyclone won't allow writing these. I've tested it in a rather convoluted manner by hand ...

### DIFF
--- a/src/ddscxx/src/dds/core/Time.cpp
+++ b/src/ddscxx/src/dds/core/Time.cpp
@@ -50,13 +50,7 @@ int64_t dds::core::Time::sec() const
 
 void dds::core::Time::sec(int64_t s)
 {
-    if(s < 0 && s != -1) {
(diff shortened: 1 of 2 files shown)


noxpardalis (approved):
This seems fine to me in principal and since Cyclone DDS (Core) prevents writing with such invalid timestamps and can handle receiving them it makes sense to propagate the handle of this to the Core. Being liberal in what we consume and conservative in what we produce and all that.

However, from an internal discussion I think some more work might be needed in the back and forth mapping between the Core and C++ bindings' notions of invalid timestamps (i.e. `DDS_TIME_INVALID` and `DDSI_TIME_INVALID` for the Core).

Approving this as it fixes an issue now and another PR/investigation can be done to fixup/confirm the mapping of timestamps.

PR #505, opened 2024-08-21
delete associated dds_entity in DDScObjectDelegate.close

MarcelJordense:
The cyclone dds_entity associated with an CXX Entity must be deleted when the close on that entity is called


eboasson (approved):
Thanks @MarcelJordense !

PR #508, opened 2024-09-30
Fix compilation error on RedHat when compiling with -Werror=sign-conversion

e-hndrks:
The following compilation error was generated on RedHat platforms with older gcc compilers (e.g. gcc8.4) when compiling with the  -Werror=sign-conversion:
"conversion to 'unsigned int' from 'int' may change the sign of the result"
This was caused by an implicit promotion to int32_t from the result of a logical AND operator on an unsigned char with a uint32_t as the mask. A subsequent logical OR on this result with an unsigned char shifted by four bits would cause the signed/unsigned conflict. By making sure all parameters are promoted to type int, we avoid this situation.


eboasson (approved):
LGTM

PR #511, opened 2024-10-05
Correctly names CMake version configuration file 

### LINKED ISSUE #510: <Package>ConfigVersion.cmake in not installed
For `find_package(CycloneDDS-CXX ${CYCLONEDDS_CXX_VERSION_REQUIRED} REQUIRED)` to work in a cmake project that uses CycloneDDS-CXX, we need `CycloneDDS-CXXConfigVersion.cmake` to be installed. However, what's installed is `CycloneDDS-CXXVersion.cmake`. Therefore find_package fails. This is fixed by just renaming the version file to the expected name.

I am on version 0.10.2.

Fix is to modify the following line: https://github.com/eclipse-cyclonedds/cyclonedds-cxx/blob/master/CMakeLists.txt#L215, so that we have the following

```
# Generate <Package>Version.cmake
write_basic_package_version_file(
  "${PROJECT_NAME}ConfigVersion.cmake"
  VERSION ${PROJECT_VERSION}
  COMPATIBILITY SameMajorVersion)
```

(This is set correctly in the [C project CMakeLists.txt](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/CMakeLists.txt)

cvilas:
Fixes #510

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,13 +212,13 @@ configure_package_config_file(
 
 # Generate <Package>Version.cmake
 write_basic_package_version_file(
-  "${PROJECT_NAME}Version.cmake"
+  "${PROJECT_NAME}ConfigVersion.cmake"
   VERSION ${PROJECT_VERSION}
   COMPATIBILITY SameMajorVersion)
 
 install(
   FILES "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
-        "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Version.cmake"
+        "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
   DESTINATION "${CMAKE_INSTALL_CMAKEDIR}"
   COMPONENT dev)
 


eboasson (commented):
Thank you! I suppose no-one ever tried requiring a specific version then ... 🤔 This change looks correct to me, and completely in line with the CMake documentation.

Perhaps it would be possible to test it, e.g., the core repository does a trial run for building an example using the installed version and adding the version number requirement would verify that this gets installed correctly.
https://github.com/eclipse-cyclonedds/cyclonedds/blob/2d705c2319342fa10e3916dbf72a93c70165f0f7/.azure/templates/build-test.yml#L180-L196

It turns out such a test is not part of the C++ binding CI at all. Hmm. Anyway, if you're inclined to have a look at that, that'd be great.

Before I can merge this PR (or an extended version of it) there is the matter of the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/eca/). Would you be willing to sign it? Please let me know in a comment on the ticket if you did it or if you don't want to do it. 🙏

cvilas:
@eboasson Thanks for the review. I can confirm that I just signed the ECA.

eboasson (approved):
Thanks @cvilas !

PR #514, opened 2024-10-23
Remove absolute path in implementation header include

AlexisPotvinARA:
The header include statement in the source was using an absolute path.
Update idlcxx file generator.c to matches the reference implementation in [libidlc__generator.c](https://github.com/eclipse-cyclonedds/cyclonedds/blob/447912e79fd08e2a0dcb439a0022116a96267c90/src/tools/idlc/src/libidlc/libidlc__generator.c#L150
)
Now the header import in the source only contains the header file.


eboasson (approved):
LGTM, thanks @AlexisPotvinARA !

PR #515, opened 2024-10-31
Support all debian architectue in cpack 

AlexisPotvinARA:
Sister PR: https://github.com/eclipse-cyclonedds/cyclonedds/pull/2121
Cpack enforces architecture debian architecture i386 or amd64 even when compiling for arm64.
Now the architecture is detected by dpkg as per the standard debian process:
https://github.com/Kitware/CMake/blob/f3f9ad9499526963a46c10733ab9e8b3c0ebfcb1/Modules/Internal/CPack/CPackDeb.cmake#L541


eboasson (approved):
Thank you! ☺️

Issue #513, opened 2024-10-22
Unions fail to compile if they contain multiple strings with different bounds


banburybill:
I have been blessed with some problematic IDL. Here's a test case:

```
    union Strings switch(long)
    {
    case 0: string<10> str0;
    case 1: string str1;
    };
```

Not unreasonably, Cyclone represents both union members as `std::string`. The generated union code maps to a `std::variant` containing two items of `std::string`, which the standard explicitly permits. The getter and setter code references union members by type, e.g. `std::get<std::string>()`, which fails as this only works if all types in the variant are unique.

I'm guessing this is a variant of #127, fixed in #223. This looks to suggest that buggy `std::variant` implementations exist which can't handle multiple members of the same type, which is a pity since my first thought was to fix this by switching to access via discriminator rather than type. :-(

banburybill:
Got it. Yes, that would work. And I can see a type implementing a proper bounded string type would be nice.

That being said, how do you feel about PR #516 which extends variant type de-duplication to check the generated cpp11 type as well as the IDL type? As you say, it's not a given that `string` and `string<n>` get mapped to the same type, so it seems to me there is no alternative but to check the generated cpp11 type.

banburybill:
PR #519  implements option 2 above.

eboasson:
I kinda like your suggestion of using indices, at least in my limited knowledge of C++ vagaries it seems it would eliminate the potentialy thorny question of deciding if two types are equal. Only solving it properly for C++17 also makes sense to me. That would simply mean the C++11 support is somewhat restricted, and it may well be those restrictions never become a problem.

I've pinged the one contributed most of the C++ type mapping and IDLC backend, I trust his judgment much more than my own here 🙂

PR #516, opened 2024-11-04
Extend type deduplication in variants to check cpp11 type

banburybill:
This fixes issue #513 .


banburybill:
Abandoning in favour of using index access to variants.

PR #519, opened 2024-11-07
Switch union std::variant access to being by index

banburybill:
Accessing std::variant by index means the type list for the variant doesn't have to be de-duped, saving a chunk of logic and ensuring bounded and unbounded strings can be mixed in unions with no problems.

Preserve access by type for use with boost::variant, where it is the only option.

### DIFF
--- a/src/ddscxx/tests/Regression.cpp
+++ b/src/ddscxx/tests/Regression.cpp
@@ -218,6 +218,11 @@ TEST_F(Regression, union_duplicate_types)
   d_s.c_1({1,2,3,4,5,6});
 }
 
+TEST_F(Regression, union_duplicate_string_types)
+{
+  duplicate_string_types_union d_t;
+}
+
 TEST_F(Regression, delimiters_bitmask)
 {
   bytes s_bm1_bytes =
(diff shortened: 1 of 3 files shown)


eboasson (commented):
I don't know the relevant C++ bits and the IDL compiler backend for it well enough to quickly figure out if this is good or not. @reicheratwork knows what's what much better than I do, and with a bit of luck he can find some time to look at it.

One unrelated thing is that contributions to Cyclone can only be merged if the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/eca/) has been signed by all authors of commits. The check now says you haven't yet. Perhaps you'd be willing to sign it?

reicheratwork:
> I don't know the relevant C++ bits and the IDL compiler backend for it well enough to quickly figure out if this is good or not. @reicheratwork knows what's what much better than I do, and with a bit of luck he can find some time to look at it.
> 
> One unrelated thing is that contributions to Cyclone can only be merged if the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/eca/) has been signed by all authors of commits. The check now says you haven't yet. Perhaps you'd be willing to sign it?

Looking at the changes, the PR seems to do what is claims.
However, looking the way that boost::variant works, with there being no index based access, this approach would cause people using boost to still experience the same issue.
I propose to have a more integral solution, which is to have the `is_same_type` function correctly compare the types in the union, not the more simplistic implementation currently in use.

eboasson (approved):
I think it makes the most sense to merge this and solve the problem today at least for those that can use C++17. If/when some solution for C++11 / Boost comes along, I'll be happy to merge that too.

PR #520, opened 2024-11-21
Improve union types deduplication

### LINKED ISSUE #513: Unions fail to compile if they contain multiple strings with different bounds

reicheratwork:
Fixes issue #513
Rewrote `is_same_type` function to (more) correctly deduplicate union types, but only when `boost::variant` is used, as this does not have access by index
Added unittests to test more constructs in addition to those implemented by Jim Hague in PR #519 

@banburybill , could you give this PR a look to see whether this fixes #513 for you?

### DIFF
--- a/src/ddscxx/tests/Regression.cpp
+++ b/src/ddscxx/tests/Regression.cpp
@@ -223,6 +223,16 @@ TEST_F(Regression, union_duplicate_string_types)
   duplicate_string_types_union d_t;
 }
 
+TEST_F(Regression, union_duplicate_array_types)
+{
+  duplicate_array_types_union d_t;
+}
+
+TEST_F(Regression, union_duplicate_bitmask_types)
+{
+  duplicate_bitmask_types_union d_t;
+}
+
 TEST_F(Regression, delimiters_bitmask)
 {
   bytes s_bm1_bytes =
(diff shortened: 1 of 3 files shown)


banburybill:
My specific failure case was a union with bounded and unbounded strings. This looks good to me, and the refactoring is great.

I guess it's vulnerable to someone using the templating to replace `std::string` with a genuine bounded string class. But I think that is probably a very niche usage, and fixing the out-of-the-box problem is far more important.

reicheratwork:
> I guess it's vulnerable to someone using the templating to replace `std::string` with a genuine bounded string class. But I think that is probably a very niche usage, and fixing the out-of-the-box problem is far more important.

That is also being checked for in the following code part:
```
const bool bstr_unique = (strcmp(gen->bounded_string_format, gen->string_format) != 0),
                  bseq_unique = (strcmp(gen->bounded_sequence_format, gen->sequence_format) != 0);
```
So in the case when a "normal" string is `std::string` and a bounded string is something else, the boolean `bstr_unique` is true and the bounded versions of strings will have their bound values added to the array identifying the type being compared. And therefore will "look like" different type, when compared to the unbounded versions.

The same is checked for when encountering (un)bounded sequences.

banburybill:
Ah, I missed that bit. You're way ahead of me.

eboasson (approved):
LGTM 😀 Thank you @reicheratwork !

PR #521, opened 2025-01-21
Fix for cross compiling examples

ThijsSassen:
matched the behavior with cyclonedds


eboasson (approved):
LGTM, thanks @ThijsSassen

PR #523, opened 2025-01-28
feature/implement-matched publication data

trittsv:
This pr implements `matched_publication_data` function and ostream operator for keys.
It also fixes the datatype for keys.

For example to get the partitions of the writer inside a wildcard-partition reader (`*`) callback.
```cpp
auto handle = sample_iter->info().publication_handle();
auto pub_data = dds::sub::matched_publication_data(reader, handle);
for (auto& n : pub_data.partition()->name()) {
    std::cout << "Partition: " << n << std::endl;
}
std::cout << "key: " << pub_data.key() << std::endl;
std::cout << "topic_name: " << pub_data.topic_name() << std::endl;
std::cout << "type_name: " << pub_data.type_name() << std::endl;
```
Output:
```
Partition: Hello
key: 01104b56-17d6-1860-43c8-032200000202
topic_name: HelloWorldData_Msg
type_name: HelloWorldData::Msg
```

@eboasson could you have a look?

### DIFF
--- a/src/ddscxx/include/dds/topic/TBuiltinTopicKey.hpp
+++ b/src/ddscxx/include/dds/topic/TBuiltinTopicKey.hpp
@@ -43,14 +43,14 @@ public:
      *
      * @return the BuiltinTopicKey
      */
-    const uint8_t* value() const;
+    const std::array<uint8_t, 16>& value() const;
 
     /**
      * Sets the BuiltinTopicKey.
      *
-     * @param v the value to set
+     * @param key the value to set
      */
-    void value(uint8_t v[]);
+    void value(const std::array<uint8_t, 16>& key);
 };
 
 #endif /* OMG_DDS_TOPIC_TBUILTIN_TOPIC_KEY_HPP_ */
--- a/src/ddscxx/include/dds/topic/TBuiltinTopicKey.hpp
+++ b/src/ddscxx/include/dds/topic/TBuiltinTopicKey.hpp
@@ -43,14 +43,14 @@ public:
      *
      * @return the BuiltinTopicKey
      */
-    const int32_t* value() const;
+    const uint8_t* value() const;
 
     /**
      * Sets the BuiltinTopicKey.
      *
      * @param v the value to set
      */
-    void value(int32_t v[]);
+    void value(uint8_t v[]);
 };
 
 #endif /* OMG_DDS_TOPIC_TBUILTIN_TOPIC_KEY_HPP_ */
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
@@ -16,7 +16,9 @@
 #define CYCLONEDDS_TOPIC_BUILTIN_TOPIC_KEY_DELEGATE_HPP_
 
 #include <array>
+#include <algorithm>
 #include <iomanip>
+
 
 namespace org
 {
--- a/src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
+++ b/src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
@@ -15,6 +15,8 @@
 #ifndef CYCLONEDDS_TOPIC_BUILTIN_TOPIC_KEY_DELEGATE_HPP_
 #define CYCLONEDDS_TOPIC_BUILTIN_TOPIC_KEY_DELEGATE_HPP_
 
+#include <iomanip>
+
 namespace org
 {
 namespace eclipse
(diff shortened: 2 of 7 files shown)


src/ddscxx/include/dds/topic/TBuiltinTopicKey.hpp
```diff
@@ -43,14 +43,14 @@ class dds::topic::TBuiltinTopicKey : public ::dds::core::Value<D>
      *
      * @return the BuiltinTopicKey
      */
-    const int32_t* value() const;
```
reicheratwork, line 46: Isn't this part of the specification, and therefore should this not be changed?

reicheratwork (commented):
I am just a little bit confused about the necessity of replacing the key value representation from `int32_t` to `int8_t`.
I am afraid this might change the interface away from the specification or even break it.

For the rest, this PR implements what it claims to.

src/ddscxx/include/dds/topic/TBuiltinTopicKey.hpp
reicheratwork: Never mind, after further discussion with @eboasson , it seems that changing this to int8_t will not cause large issues.

src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
```diff
@@ -27,37 +29,44 @@ namespace topic
...
-                 && other.key_[1] == key_[1]
-                 && other.key_[2] == key_[2];
+        return other.key_ == key_;
+    }
+
+    friend std::ostream& operator<<(std::ostream& os, const BuiltinTopicKeyDelegate& key)
+    {
+        for (size_t i = 0; i < key.key_.size(); ++i)
+        {
+            if (i == 4 || i == 6 || i == 8 || i == 10)
```
eboasson, line 59: Actually, I would print it differently. The 4/6/8/10 I think comes from ye standard GUID formatting, but the GUIDv1..v6 (or however many versions of GUIDs have been defined so far) have a completely different structure from the GUIDs used in DDSI (and in practice, DDS, except for OpenSplice which actually implements the old DDS spec). In fact, I would calling them "GUID"s in DDS turns out to be an unfortunate choice because of this.

The GUIDs of DDS are structured as: a 12-byte identifier for the DomainParticipant, followed by 4 byte identifier for the DataReader/DataWriter/... within the participant. The latter 4 bytes have a structure as well: the last byte contains an entity kind (reader/writer/participant/topic/...) and a "source" (application, vendor, built-in).

Then, since DDSI 2.2, the first two bytes of the GUID are supposed to be the vendor code. This was done to solve the problem that the spec doesn't define the way the 12 participant bytes had to be chosen. I think it is a shame that 2 bytes of entropy were thrown away there. Most implementations do that (OpenSplice is different, but that goes back to it following the old DDS spec).

What I usually do is simply print the bytes in hex, grouped in 4 byte units ...

But that's just me, you might very good reasons for using this format 🙂

src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
```diff
@@ -27,37 +29,44 @@ namespace topic
...
+    BuiltinTopicKeyDelegate(uint8_t v[16])
     {
-        key_[0] = v[0];
-        key_[1] = v[1];
-        key_[2] = v[2];
+        std::copy(v, v + 16, key_.begin());
     }
 public:
-    const int32_t* value() const
+    const uint8_t* value() const
```
eboasson, line 40: I can't figure it out if the application can ever get hold of this `uint8_t*` or not ... I think it shouldn't be able to, but it also appears as if this comes directly from `TBuiltinTopicKey` and that (therefore) the OMG-provided API definition includes the ability to do this.

A `std::array<uint8_t,16>&` would be much nicer, but that would be more of a portability issue than this, so I guess we'll just have to live with it ...

eboasson (commented):
Thank you! It looks good to me, but before I merge it, I would appreciate it if you could tell me how well these `uint8_t *`s are hidden from the application (I commented on one of them).

I also wonder if it makes the most sense to stringify the GUIDs like this. I am sure you have a reason, so just let me know.

src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
trittsv: @eboasson I used this format because the same representation is used in cyclones python binding. So I thought that's the correct one.
But I can implement any other representation if you like, just tell me🙂

src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
trittsv: @eboasson theoretical a app can call this function. But from my experience i only ever saw someone using the BuiltinTopicKey object with its comparison operator and its stringified function.

I personally would also prefer to return a reference of the std::array. A raw pointer to a struct is not really cpp-style.

I saw a other vendor returning a vector<uint32> and a even other one returning a custom array-Like-type.

trittsv:
@eboasson and @reicheratwork thank you two for your reviews. I answered on the comments. Let me know what you think about them🙂

src/ddscxx/include/org/eclipse/cyclonedds/topic/BuiltinTopicKeyDelegate.hpp
eboasson: It looks like the OMG spec is limited to:
```C++
+template <typename D>
+class dds::topic::TBuiltinTopicKey : public ::dds::core::Value<D> {
+public:
+  typedef typename D::VALUE_T VALUE_T;
+public:
+  VALUE_T value() const;
+};
```
In ancient times, OpenSplice mapped this to `VALUE_T` to `uint32_t *` and that seems to have made its way into Cyclone despite not matching what would actually fit with Cyclone. So if my understanding is correct and I'm guessing my way through C++ correctly, `std::array<uint8_t,16>&` would be a perfectly acceptable choice.

As we have to break the broken thing anyway, we might as well try to use the opportunity to make it a bit more friendly than `uint8_t*`. I don't have a good sense for how much work that takes, but perhaps you'd be willing to take a look?

trittsv: @eboasson i changed it from `uint8_t*` to `std::array<uint8_t,16>&`. I think its a good change, now its more cpp-style and consistent with cyclone.

eboasson (approved):
LGTM! Thank you ☺️

PR #526, opened 2025-02-04
fix potential invalid instance handle for matched_publications

trittsv:
In the last PR i fixed it for `matched_subscriptions` but forgot `matched_publications` .. sorry, so here it is :)


eboasson:
> In the last PR i fixed it for `matched_subscriptions` but forgot `matched_publications` .. sorry, so here it is :)

When I made the comment about that, I figured there was no need to point out that there were two places where it applied. That was a correct assumption (this PR is the evidence for that), but it was rather stupid of me to not think of it when reviewed the update just a little bit later ...

Anyway, I am happy you realised it was overlooked and that it gets fixed now.

PR #527, opened 2025-02-11
remove ambiguous operator

trittsv:
This PR removes a ambiguous operator.

It solves the following situation.
```cpp
dds::core::InstanceHandle handle = domainParticipant.instance_handle();
std::ostringstream s;
s << handle;
```
Error:
```
error C2593: 'operator <<' is ambiguous 
D:\conan\short\f72057\1\include\ddscxx\dds/core/detail/TInstanceHandleImpl.hpp(105,22): message : could be 'std::ostream &operator <<( 
std::ostream &,const dds::core::TInstanceHandle<org::eclipse::cyclonedds::core::InstanceHandleDelegate> &)' [found using argument-depe 
ndent lookup]
D:\conan\short\f72057\1\include\ddscxx\dds/core/detail/TInstanceHandleImpl.hpp(112,22): message : or       'std::ostream &operator <<( 
std::ostream &,const dds::core::TInstanceHandle<org::eclipse::cyclonedds::core::InstanceHandleDelegate>)' [found using argument-depend 
ent lookup]
 message : while trying to match the argument list '(std::ostringstream, dds::core::InstanceHandle)'
```

### DIFF
--- a/src/ddscxx/include/dds/core/detail/TInstanceHandleImpl.hpp
+++ b/src/ddscxx/include/dds/core/detail/TInstanceHandleImpl.hpp
@@ -108,13 +108,6 @@ inline std::ostream& operator << (std::ostream& os, const dds::core::TInstanceHa
     return os;
 }
 
-//Overloaded operator << (workaround issues 252 and 162)
-inline std::ostream& operator << (std::ostream& os, const dds::core::TInstanceHandle<org::eclipse::cyclonedds::core::InstanceHandleDelegate> h)
-{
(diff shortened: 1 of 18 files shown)


eboasson:
@trittsv that failure seems to have something to do with an interaction between gtest and old gcc. I have no idea what/why/how/..., but it builds fine with this commit on top of yours: https://github.com/eboasson/cyclonedds-cxx/commit/ee6d57b9f207209d5149e7054c1eae6757716f81 

What do you think?

PR #537, opened 2025-06-03
Fixes an issue with non-final fields not being finished correctly

reicheratwork:
Fields which have a length header in XCDR were not skipping the cursor of the stream to the position indicated by the length field in the preceding header, this is fixed
Added tests to check whether this works as intended


eboasson (approved):
Thanks @reicheratwork !

kasars:
I can confirm that the change fixes the problem! Thanks!

My template coding fu is not the strongest, so it is possible I have missed something, but I cannot see anything that would strike me as wrong. :)

(I would probably not have used rand() to populate the structures, as that could lead to random failures, but the probability of that in this scenario is minimal)

Big thanks again!

kasars:
Looks good to me! :)

eboasson (approved):
LGTM! Thank you @reicheratwork and @kasars ☺️

PR #562, opened 2025-11-25
Disable security/ssl in Cyclone build on CI

dpotman:
Avoids test failures in C++ Windows CI builds, caused by unresolved OpenSSL library.


eboasson (approved):
Thanks @dpotman !

PR #563, opened 2025-12-08
Remove the idlcxx package and merged it into dev

AlexisPotvinARA:
Remove the component idlcxx and merge it with the dev component.
Remove the packaging of the idlcxx library.


AlexisPotvinARA:
@eboasson what do you think about this? 
Would it be better to add this component to the -dev package?

eboasson:
> @eboasson what do you think about this? Would it be better to add this component to the -dev package?

The name you propose is definitely a lot nicer, but I think your proposal to include it in the `-dev` package is better still. Would you be willing to make that change?

